PR #14, opened 2017-10-27
Add permissions field to module.json

izqui:
Added a new permissions field to the metadata of apps, so when creating an instance of an app that needs permissions over another app, it can be displayed to the user in the client.

We should probably extend this and document all other initialization values so the client can show options to initialize apps.

### DIFF
--- a/apps/finance/module.json
+++ b/apps/finance/module.json
@@ -7,5 +7,12 @@
         { "name": "Execute payments", "id": "EXECUTE_PAYMENTS_ROLE" },
         { "name": "Disable payments", "id": "DISABLE_PAYMENTS_ROLE" }
     ],
+    "permissions": [
+        {
+            "appName": "vault.aragonpm.test",
+            "role": "TRANSFER_ROLE",
+            "entity": { "type": "INIT_VALUE", "value": "_vault" }
+        }
+    ],
     "path": "contracts/Finance.sol"
 }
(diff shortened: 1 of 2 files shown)


onbjerg:
This makes sense, but can you elaborate a bit?

- What is the `entity` key, more specifically the `entity.type` key? Can there be differen types, if so, which?
- Do we want to keep it as `role` (singular) or do we want to change it to `roles` (an array) before merging?
- Do you have something in mind for the generated artifact in regards to the permissions? E.g. I assume the `appName` should become `appId` with the namehash calculated, same with `role` - should probably add the role bytes to it.
- Is this strictly for UI/UX purposes, or do we need to enforce some kind of behaviour regarding these permissions?
- Is dependencies a better name?

I hope the questions make sense 😄

onbjerg:
Also, this should be reflected in `aragon-dev-cli` as well.

onbjerg:
Proposed types:

```typescript
type ApplicationRole = {
  name: string,
  id: string
}

type ApplicationDependency = {
  appName: string,
  role: string,
  entity: {
    type: string,
    value: string
  }
}

type ApplicationIcon = {
  src: string,
  sizes: string
}
```

```typescript
// app.json
type ApplicationDescriptor = {
  appName: string,
  version: string,
  roles: Array<ApplicationRole>,
  dependencies: Array<ApplicationDependency>,
  path: string
}

// manifest.json
type ApplicationManifest = {
  name: string,
  description: ?string,
  icons: Array<ApplicationIcon>
}
```

PR #16, opened 2017-11-01
Group: client app (WIP)

bpierre:
The app is using create-react-app, Aragon UI and styled-components.

## Instructions

To test it, you need to use the master branch of Aragon UI (for now).

In the Aragon UI directory:

```sh
npm link
```

In the Aragon Apps directory:

```sh
git fetch && git checkout -b aragon-ui-group origin/aragon-ui-group
cd apps/group/app
npm install
npm link '@aragon/ui'
npm start
```


apps/group/app/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "Group",
```
onbjerg: This file is a duplicate of the `manifest.json` in the app root. They serve the same purpose. Also, `short_name` should be the same as `name`, as we only use `name`.

The icon at `icons[0]` will actually be used for the icon in the sidebar as well, just a small side note.

We also won't use `start_url`, `display`, `theme_color` or `background_color` - so these are just no-ops 😊

apps/group/app/src/components/EmptyStateCard/EmptyStateCard.js
```diff
@@ -0,0 +1,40 @@
...
+  h1 {
+    margin: 20px 0 5px;
+  }
+  button {
+    width: 150px;
+    margin-top: 20px;
+  }
+`
+
+// AragonApp provides everything needed to start an Aragon App.
```
onbjerg: "Dead" comment

apps/group/app/src/index.js
```diff
@@ -0,0 +1,7 @@
+import React from 'react'
+import ReactDOM from 'react-dom'
+import App from './App'
+import registerServiceWorker from './registerServiceWorker'
+
+ReactDOM.render(<App />, document.getElementById('root'))
+registerServiceWorker()
```
onbjerg: I don't think this is possible, since it runs in a sandboxed iframe.

apps/group/app/src/registerServiceWorker.js
```diff
@@ -0,0 +1,108 @@
+// In production, we register a service worker to serve assets from local cache.
```
onbjerg: Same as above comment.

apps/group/app/public/manifest.json
bpierre: Cool, I’ll remove it!

apps/group/app/src/components/EmptyStateCard/EmptyStateCard.js
bpierre: 👍

apps/group/app/src/index.js
bpierre: Yes, removing this (it’s generated by create-react-app).

bpierre:
@onbjerg thanks for the feedback! 💯

PR #27, opened 2017-11-12
Update aragon-core dependency to 1.0.2 and adapt tests to Byzantium

### LINKED ISSUE #21: Update aragon-core dependency to 1.0.2 and adapt tests to Byzantium
A similar PR was already merged for aragon-core which could be used for reference: https://github.com/aragon/aragon-core/pull/145

yippee-ki-yay:
I updated the aragon-core dependency to 1.0.2 and upgraded the test to check for assertRevert (the new opcode added in Byzantium) also had to upgrade ethereumjs-testrpc so it has the revert opcode.

Resolve #21

### DIFF
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -498,6 +498,12 @@ contract Finance is App, Initializable, ERC677Receiver {
             return MAX_UINT;
 
         uint256 spent = periods[currentPeriodId()].tokenStatement[_token].expenses;
+
+        // A budget decrease can cause the spent amount to be greater than period budget
+        // If so, return 0 to not allow more spending during period
+        if (spent >= settings.budgets[_token])
+            return 0;
+
         return settings.budgets[_token].sub(spent);
     }
 
--- a/apps/finance/package-lock.json
+++ b/apps/finance/package-lock.json
@@ -21,9 +21,9 @@
 			}
 		},
 		"acorn": {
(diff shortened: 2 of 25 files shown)


izqui:
The CI build seems to be failing because with solidity-coverage 0.3.5, the testrpc-sc binary isn't found. Tried locally and it was still the case. Try downgrading to solidity-coverage 0.3.0.

yippee-ki-yay:
Yeah sorry about that, even when switching to 0.3.0 I get the same problem. As far as I can tell the solidity-coverage version that has the updated version to work with Byzantium also removes the testrpc-sc.

Changing the testrpc.sh in /shared folder line 18 to: 'node_modules/ethereumjs-testrpc/build/cli.node.js' fixes that issue but then I get out of gas on a few tests when running coverage.

izqui:
Which tests are you getting out of gas errors? Is it just on coverage or also when doing normal tests?

yippee-ki-yay:
Token Manager, Voting, Finance and Fundrasing they all get out of gas on the "before Each" methods so it doesnt go any further there.
I get error only when running coverage, when I run just the tests it works fine.

izqui:
Finally ready to be merged. Thanks for the contribution @yippee-ki-yay

yippee-ki-yay:
Oh great, sorry I couldn't finish it on my own, the next one maybe.
Feel free to ping me if there are some other issue I can solve (try) 👍

PR #30, opened 2017-11-13
Multiple voting app small changes

### LINKED ISSUE #18: Voting app events
~**Done** We should add an event to log when the `changeMinAcceptQuorumPct` happens.~

Also we should the number of votes that got casted in the `CastVote` event

### LINKED ISSUE #19: Execution logic in Voting app
At the moment, the result of a vote in the voting app is automatically executed if casting one vote makes the voting decided.

The problem with this is that if the executing call fails, the entire transaction is reverted and the vote isn't counted.

Possible solutions (complimentary):
- Add a boolean parameter to the `castVote` function to specify if the vote should be executed if accepted.
- Allow execution to fail but don't revert the transaction, similar to how the Gnosis multisig does when the execution fails (marks tx confirmation, but marks as not executed). This would require a way to execute EVMCallScripts without reverting when one underlying call fails.

Right now there is no notion of an expired voting, if it was approved then it can be executed at any point in time (by anyone). This could result in 'toxic votes' that get forgotten because execution was failing (a removed permission), but that could be used down the line if the becomes executable.

izqui:
- Change all instances where 'voting' was being used instead of 'vote'. Closes https://github.com/aragon/aragon/issues/53 after wiki documentation update.
- Logs voter stake in cast vote event. Closes #18 
- Add boolean parameter when casting a vote on whether it should be automatically executed . Closes #19

### DIFF
--- a/apps/voting/contracts/Voting.sol
+++ b/apps/voting/contracts/Voting.sol
@@ -50,9 +50,9 @@ contract Voting is App, Initializable, EVMCallScriptRunner, EVMCallScriptDecoder
     /**
     * @notice Initializes Voting app (some parameters won't be modifiable after being set)
     * @param _token MiniMeToken address that will be used as governance token
-    * @param _supportRequiredPct Percentage of voters that must support a voting for it to succeed (expressed as a 10^18 percetage, (eg 10^16 = 1%, 10^18 = 100%)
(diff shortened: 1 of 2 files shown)


onbjerg:
Can you rename the app to vote instead of voting too?

Issue #17, opened 2017-11-07
Multisig app


izqui:
At the present moment the most widely used governance mechanism is through a multisig. 

Different alternatives for implementation:
- Use a widely used trusted multisig like Gnosis.
- Extend Group app to require `n` confirmations before performing action (currently Group app is a 1 out of `m` multisig)

I personally think it is a better idea to implement our own solution for this on top of the Group app. Another question if we go down this route is whether we have two separate apps (Group and Multisig) or the current Group functionality is just a Multisig with only one required signature.

onbjerg:
I think adding confirmations to the group app is an exciting prospect as it has different use cases. Also, a group app with multiple confirmations + the vault is pretty much a multisig.

PR #41, opened 2017-11-22
Multisig group

cpurta:
Allow groups to require multiple signatures from group members before allowing any action to be taken. This requires that at least M out of N confirmations are set before allowing any action to take place. 

Also added testing to check that this works for M out of N and N out of N (group consensus) confirmations.

Adresses #17


apps/group/contracts/Group.sol
```diff
@@ -11,10 +11,17 @@ import "@aragon/core/contracts/misc/Migrations.sol";
 
 contract Group is App, Initializable, IForwarder, EVMCallScriptRunner {
     string name;
-    mapping (address => bool) members;
+    address[] public members;
+    mapping (address => bool) isMembers;
```
izqui: `isMembers` -> `isMember`

apps/group/contracts/Group.sol
```diff
@@ -45,26 +65,78 @@ contract Group is App, Initializable, IForwarder, EVMCallScriptRunner {
...
+            changeRequirement(members.length);
+
         RemoveMember(_entity);
     }
 
+    function numMembers() public constant returns (uint) {
+        return members.length;
+    }
+
+    function getRequired() public constant returns (uint) {
```
izqui: If `required` is public there is no need to have another getter

apps/group/contracts/Group.sol
```diff
@@ -23,19 +30,32 @@ contract Group is App, Initializable, IForwarder, EVMCallScriptRunner {
...
 
+        required = _required;
         name = _name;
     }
 
+    /**
+    * @notice Changes the number of members required to confirm before the group can perform actions
+    * @param _required New number of members required to confirm action
+    */
+    function changeRequirement(uint _required) public {
```
izqui: Any group member shouldn't be allowed to change the requirement, it completely defeats the purpose as anyone would be able to set the requirement to 1 and act.

izqui: Change requirement should have its own role and be protected with the auth modifier

apps/group/contracts/Group.sol
```diff
@@ -45,26 +65,78 @@ contract Group is App, Initializable, IForwarder, EVMCallScriptRunner {
...
+        return members.length;
+    }
+
+    function getRequired() public constant returns (uint) {
+        return required;
+    }
+
+    function confirm() public {
+        require(isGroupMember(msg.sender));
+        confirmations[msg.sender] = true;
```
izqui: Confirmations should happen on an action per action basis, right now once enough confirmations have been given to the contract, any member can perform any actions at any moment (until confirmations are revoked).

For an example of how this can be achieved you can see [Gnosis' Multisig](https://github.com/gnosis/MultiSigWallet/blob/master/contracts/MultiSigWallet.sol#L30).

apps/group/contracts/Group.sol
```diff
@@ -45,26 +65,78 @@ contract Group is App, Initializable, IForwarder, EVMCallScriptRunner {
...
+        return count >= required;
+    }
+
     /**
     * @dev IForwarder interface conformance. Forwards actions to any group member.
     * @param _evmCallScript Script being forwarded
     */
     function forward(bytes _evmCallScript) external {
-        require(isGroupMember(msg.sender));
+        require(isGroupMember(msg.sender) && isConfirmed());
```
izqui: After adopting a confirmations per action model, forwarding should confirm the action and then if the action has enough confirmations, perform the action.

apps/group/test/group.js
```diff
@@ -8,16 +8,19 @@ contract('Group app', accounts => {
     let app = {}
 
     const member = accounts[1]
+    const member2 = accounts[2]
+    const member3 = accounts[3]
     const groupName = 'Test Group'
+    const required_confirmations = 1;
```
onbjerg: Please change this variable name to camel case to conform to our style guide 😊

cpurta:
Updated the code to include some the changes requested. Would love a review when you guys have some time.

apps/group/contracts/Group.sol
```diff
@@ -45,26 +69,113 @@ contract Group is App, Initializable, IForwarder, EVMCallScriptRunner {
...
+            if (members[i] == _entity) {
+                members[i] = members[members.length - 1];
+                delete members[members.length - 1];
+                members.length -= 1;
+                break;
+            }
+        }
+
+        if (required > members.length)
+            changeRequirement(members.length);
```
izqui: Because `changeRequirement` could be called if a requirement reset is needed, in order for `removeMember` to work, the sender must also have the `CHANGE_REQUIREMENTS_ROLE ` role. I would say, in case `removeMember` requires a change in requirement, always allow it to happen. 

A good way to implement this could be moving the logic for changeRequirement to an internal function that both `removeMember` and external `changeRequirement` call.

apps/group/contracts/Group.sol
```diff
@@ -45,26 +69,113 @@ contract Group is App, Initializable, IForwarder, EVMCallScriptRunner {
...
+    }
+
+    /**
+    * @notice Allows a group member to confirm that a specific action can be executed
+    * @param _actionId Index of an action
+    */
+    function confirm(uint _actionId) public {
+        require(isGroupMember(msg.sender));
+        confirmations[_actionId][msg.sender] = true;
+        Confirmation(_actionId, msg.sender);
```
izqui: If confirmation makes the action approved, execution should be triggered to minimize the number of required transactions.

apps/group/contracts/Group.sol
```diff
@@ -45,26 +69,113 @@ contract Group is App, Initializable, IForwarder, EVMCallScriptRunner {
...
+
+    /**
+    * @notice Adds an action for the group to take based on confimations
+    * @param _evmCallScript Script being forwarded
+    */
+    function addAction(bytes _evmCallScript) public constant returns (uint _actionId) {
+        _actionId = actionCount;
+        actions[_actionId] = _evmCallScript;
+        actionCount += 1;
+        ActionAdded(_actionId);
```
izqui: If the sender is a member, `addAction` could add the initial confirmation.

izqui: This action should only be callable by group members to avoid spam.

apps/group/contracts/Group.sol
```diff
@@ -45,26 +69,113 @@ contract Group is App, Initializable, IForwarder, EVMCallScriptRunner {
...
+    }
+
     /**
     * @dev IForwarder interface conformance. Forwards actions to any group member.
     * @param _evmCallScript Script being forwarded
     */
     function forward(bytes _evmCallScript) external {
```
izqui: The idea is that the `forward` function should be the one that initiates the action. The reason for that is when the client sees that the Group is the entity that can perform the action, then if the Group returns true to `canForward(sender, payload)`, it then performs a `forward` transaction. After that, the process of when or how the action gets finally performed varies. In the case of this multisig group, other member confirmations are needed for execution. In the case of the Voting app, forwarding starts a vote for token holders to decide whether the action is performed or not.

apps/group/contracts/Group.sol
```diff
@@ -45,26 +69,113 @@ contract Group is App, Initializable, IForwarder, EVMCallScriptRunner {
...
-        require(isGroupMember(msg.sender));
+        uint actionId = getActionId(_evmCallScript);
+        require(isGroupMember(msg.sender) && isConfirmed(actionId));
```
izqui: Keep track that the action has been executed so it cannot be executed multiple times.

cpurta:
Appreciate the review and comments, @izqui. Currently working on some of those changes and getting the tests to work with those.

apps/group/contracts/Group.sol
cpurta: Added some logic at the end of the function to execute the action if it has enough confirmations.

izqui:
Would love to get your input on https://github.com/aragon/aragon-apps/issues/54 @cpurta!

cpurta:
@izqui feel free to close this out, based on the discussion in #54

izqui:
Yeah, sorry for that @cpurta, this was making a lot of interesting progress... We are deleting an incredible amount of code as a result of #54, which in the long run I believe is good. 

Thanks for taking the time and looking forward to other contributions :)

PR #42, opened 2017-11-27
[WIP] Group: update for publicURL

sohkai:
Requires https://github.com/aragon/aragon-ui/pull/18.

Adds publicURL sourcing to Group app.

Todo:
* [ ] Actually copy over the assets; @bpierre as far as I can tell, this is going to require us to drop out of `react-scripts` and use our own webpack/dev-server configs?

### DIFF
--- a/apps/group/app/src/components/FakeShell/FakeShell.js
+++ b/apps/group/app/src/components/FakeShell/FakeShell.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
-import { AragonApp, Button, DropDown, Text, theme } from '@aragon/ui'
+import { AragonApp, Button, DropDown, Text, publicUrlInjector, theme } from '@aragon/ui'
 import { OverlayPanel } from '..'
 
 import chevron from './assets/chevron.svg'
@@ -60,15 +60,16 @@ const TopBarStart = styled.div`
 `
 const TopBarEnd = styled.div`
   padding-right: 30px;
-  ${Button} {
-    width: 150px;
-  }
 `
 
 const TopBarTitle = styled.h1`
   padding-right: 20px;
   margin-right: 20px;
   background: url(${chevron}) no-repeat 100% 50%;
+`
+
+const TopBarEndButton = styled(Button)`
+  width: 150px;
 `
 
 class FakeShell extends React.Component {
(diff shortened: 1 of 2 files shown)


apps/group/app/src/components/FakeShell/FakeShell.js
```diff
@@ -71,6 +68,10 @@ const TopBarTitle = styled.h1`
   background: url(${chevron}) no-repeat 100% 50%;
 `
 
+const TopBarEndButton = styled(Button)`
+  width: 150px;
+`
```
sohkai: See https://github.com/aragon/aragon-ui/pull/19; we can't use reverse selectors directly with `Button`s.

bpierre:
> Actually copy over the assets; @bpierre as far as I can tell, this is going to require us to drop out of react-scripts and use our own webpack/dev-server configs?

What about a task to sync the files? Something like:

```json
{
  "scripts": {
    "update-assets": "rsync -rtu node_modules/@aragon/ui/dist/ public/aragon-ui/",
    "start": "npm run update-assets && react-scripts start",
    "build": "npm run update-assets && react-scripts build"
  }
}
```

sohkai:
@bpierre +1. After some fiddling, this seems to work!

apps/group/app/scripts/sync-assets.js
```diff
@@ -0,0 +1,25 @@
...
+}
+
+// Sync each directory
+for ({ from, to, opts = {} } of Object.values(assets)) {
+  syncFiles(from, to, { ...defaultSyncOpts, ...opts }, (event, data) => {
+    if (event === 'error') {
+      console.error(data.message || data)
+    }
+  })
+}
```
sohkai: @onbjerg this might be useful for the [react boilerplate](https://github.com/aragon/aragon-react-boilerplate).

apps/group/app/package.json
```diff
@@ -11,12 +11,14 @@
     "styled-components": "^2.2.3"
   },
   "devDependencies": {
-    "react-scripts": "1.0.16"
+    "react-scripts": "1.0.16",
+    "sync-files": "^1.0.3"
```
sohkai: @bpierre `rsync` isn't platform agnostic, so I ended up using [`sync-files`](https://github.com/byteclubfr/node-sync-files/) instead.

sohkai:
I've avoided committing the `public/aragon-ui` folder for now, since everything there will be modified when the next version is released. We can add it once @aragon/aragon-ui is stable.

PR #58, opened 2018-01-07
finance: Add receiveApproval function

### LINKED ISSUE #56: Add receiveApproval function to Finance contract
As stated in the wiki, in order to deposit ERC20 tokens in Finance app, right now we need 2 calls:
```
token.approve(finance, amount)
finance.deposit(token, amount, string reference)
```
While for ERC677 tokens we just need one:
```
token.transferAndCall(finance, amount, string reference)
```
Adding the function `receiveApproval` would allow to deposit ERC20 tokens using just one call too.

bingen:
To be able to deposit ERC20 tokens with `approveAndCall`.

Closes #56.

### DIFF
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -151,7 +151,6 @@ contract Finance is App, Initializable, ERC677Receiver {
     )
         transitionsPeriod
         external
-        returns (bool success)
     {
         ERC20 token = ERC20(_token);
         _recordIncomingTransaction(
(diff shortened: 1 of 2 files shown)


apps/finance/contracts/Finance.sol
```diff
@@ -136,6 +136,34 @@ contract Finance is App, Initializable, ERC677Receiver {
...
+    {
+        ERC20 token = ERC20(_token);
+        _recordIncomingTransaction(
+            token,
+            _from,
+            _amount,
+            string(_data)
+        );
+        require(token.transferFrom(_from, address(vault), _amount));
+        return true;
```
izqui: Same as this https://github.com/aragon/aragon-apps/pull/59#discussion_r160099249. Lets make the same in both cases :)

PR #59, opened 2018-01-07
finance: Add function to send tokens to Vault

### LINKED ISSUE #57: Add function to Finance to send tokens to Vault
If accidentally somebody sends tokens to Finance contract address directly through `transfer`, tokens could be locked in that contract as Finance doesn't have a way to send transactions (it's using Vault for that).
My suggestion is to add a function that allows to send all tokens in Finance to Vault anytime.

bingen:
If accidentally somebody sends tokens to Finance contract address
directly through transfer, tokens could be locked. This function
allows to send them to Vault.

Closes #57.

### DIFF
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -272,6 +272,30 @@ contract Finance is App, Initializable, ERC677Receiver {
     }
 
     /**
+     * @dev Allows make a simple payment from this contract to Vault,
+            to avoid locked tokens in contract forever.
+            This contract should never receive tokens with a simple transfer call,
+            but in case it happens, this function allows to recover them.
+     * @notice Send tokens to Vault
+     * @param _token Token whose balance is going to be transferred.
+     * @return success boolean indicating whether transaction suceeded
+     */
+    function depositToVault(address _token) public returns (bool success) {
+        ERC20 token = ERC20(_token);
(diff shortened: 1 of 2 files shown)


apps/finance/contracts/Finance.sol
```diff
@@ -271,6 +271,30 @@ contract Finance is App, Initializable, ERC677Receiver {
...
+    function depositToVault(address _token) public returns (bool success) {
+        ERC20 token = ERC20(_token);
+        uint256 value = token.balanceOf(this);
+        require(value >= 0);
+
+        _recordIncomingTransaction(
+            token,
+            this,
+            value,
+            "Transfer from Finance"
```
izqui: I would make this 'Deposit to Vault' to make it more descriptive

apps/finance/contracts/Finance.sol
```diff
@@ -271,6 +271,30 @@ contract Finance is App, Initializable, ERC677Receiver {
...
+        );
+        require(token.transfer(address(vault), value));
+        return true;
```
izqui: Unsure why the return boolean is needed, as it always will either fail through a revert or return true. 

I propose to either remove the return value altogether or do `return token.transfer(address(vault), value)`, I think both would be fine, but I would side with removing the return value tbh.

PR #66, opened 2018-01-10
Update testing to use geth and parity clients

### LINKED ISSUE #64: Run tests in geth and parity
Same as https://github.com/aragon/aragon-core/issues/171

cpurta:
Allow for coverage tests to use either a running geth instance or parity instance. This allows for more robust testing against private development networks. 

Will resolve #64 when completed.

All apps are currently configured to use the development network. Based on the truffle-config.js in the aragon-core repo it will cause tests to use the ganache-cli testrpc. We can then later update that config to have the development network point to a specific instance of geth or parity preferably locally.


cpurta:
Just as a note seems like it may be worth adding in account creation and password file creation if geth/parity have no accounts set up. Default dev account passwords would just be empty (or default to "password") to allow for simple password file creation.

izqui:
Woah! This looks *really* promising! Thanks so much for taking the time :)

- Regarding only using the first account, I think there are some tests that require multiple accounts mostly to test access control. 
- I think it would be important to make this tests be run in the CI, for that we need to either install both geth and parity directly or use Docker images for them so we can properly pin the versions.

cpurta:
Yeah that would make sense. I agree that it would be really important to add this to CI. Was thinking along the lines of use docker since it would be much easier to configure at run time. Going to be looking at some published images or perhaps building some and pushing those to docker hub.

izqui:
I have used official parity’s images in the past and work great. I think geth has images too.

Regarding the CI, I don’t think we need to run coverage on parity and geth. We can do just a test with parity and geth and then coverage on the forked version of testrpc of solidity coverage.

future-apps/fundraising/package.json
```diff
@@ -3,7 +3,9 @@
   "version": "1.0.0",
   "main": "index.js",
   "scripts": {
-    "test": "truffle test",
+    "test": "truffle test --network development",
```
izqui: No need to run tests on Fundraising for now

shared/test-helpers/testrpc.sh
```diff
@@ -3,30 +3,120 @@
...
+    # check that docker exists before attempting to pull/run images
+    if ! [ -x "$(command -v docker)" ]; then
+        echo 'Error: docker is not installed' >&2
+    fi
+}
+
+docker_start_parity() {
+    check_docker
+    # pull the most stable release of parity
+    docker pull parity/parity:stable-release --chain dev
```
izqui: I think we should pin to specific geth and parity versions!

cpurta: I agree. Are you thinking about just using the most recent versions for right now?

izqui: Yes!

izqui: The purpose of this is to avoid random CI failures because a geth or parity update happened.

cpurta:
The current geth image that I am using only has a latest tag. Checking some other images for geth seems to have "outdated" tags `arachnid/geth` has `v1.4.20` as the only tagged version. Also based on the recent push it looks like it's being built along side the other tags. Most other images are a minor version back <=1.6.*. Are we ok with keeping the `latest` tag for geth?

Also is docker enabled on travis CI? If so I will update the CI file to use the dockerized tests and see if everything has worked.

izqui:
The `.travis.yml` file needs that to enable docker:

```
sudo: required
services:
  - docker
```

izqui:
I think we should really do this at some point @sohkai @bingen

Evalir:
Should we aim to do this or just deprioritize and close? @sohkai @bingen

sohkai:
@cpurta Sorry for dragging this on _forever_; we never could find time to prioritize this given other efforts.

We're consolidating our test tooling and etc. to be more robust against real/ganache chains, so hopefully adding tests against a real node in the future will be much, much easier than where we were in 2018. For now though, let's consider these contracts to be well tested given their reliability on mainnet over the last year and a half :).

PR #70, opened 2018-01-12
truffle & solc version update + view/pure fn

### LINKED ISSUE #60: Upgrade truffle and solc and use 'view' and 'pure'
Similar to https://github.com/aragon/aragon-core/pull/169

In the case of the Finance app there can be some `constant` functions that need refactoring to convert to view, given that they modify state.

renexdev:
Closes #60


renexdev:
@izqui a cleaner PR

PR #71, opened 2018-01-12
Gas reporter

### LINKED ISSUE #65: Add gas reporter for tests
https://github.com/cgewecke/eth-gas-reporter

It would be awesome to know how much gas our tests are taking.

cpurta:
Add `eth-gas-reporter` for app testing. Creating a PR in aragon-core so that the truffle-config file needed will be imported when that is merged. 

Resolves #65


cpurta:
Not sure if the package-lock.json files should have been included in this PR. Seems like I should remove those from this PR.

izqui:
Why is that?

cpurta:
Just seems like a lot of changes. Have also seen some debates in other repos to not include the lock files in commits. Guess that is just preference though.

izqui:
Paging in @sohkai and @bpierre for a better discussion on this

cpurta:
Added the test to all apps and is enable to run by lerna. Was not able to get it to properly run due to the compilation errors that are occurring during tests.

bpierre:
> Not sure if the package-lock.json files should have been included in this PR. Seems like I should remove those from this PR.

Hi @cpurta! Yes we are using lock files in aragon/aragon and aragon/ui, but we are prioritizing `yarn.lock` for now, because npm is having some issues with `npm link` that yarn doesn’t have.

Ideally, I think we should support both until the two projects agree on a common format, but keeping them in sync [need to be done manually](https://github.com/yarnpkg/yarn/issues/3614#issuecomment-344228157), which is too annoying IMO.

So, to answer the question: yes, having a `package-lock.json` is fine, and having a `yarn.lock` instead is even better if you are using yarn 😊

cpurta:
@bpierre, thanks for the clarification.

PR #77, opened 2018-01-21
Voting app

bpierre:
Missing:

- Create a new vote
- Some polishing (visual + performances)

## How to Run

You need to run simultaneously Aragon (wrapper) and the Voting app. Aragon UI also need to use a special branch.

### react-scripts

This step is [temporarily required](https://github.com/facebook/create-react-app/pull/3759) to render the fonts in the app.

- Checkout this branch: https://github.com/bpierre/create-react-app/tree/devserver-cors
- `cd packages/react-scripts`
- `yarn && yarn link && yarn build`

### aragon/aragon-apps

- Checkout this branch: https://github.com/aragon/aragon-apps/tree/voting-webapp
- `cd apps/voting/app`
- `yarn && && yarn link 'react-scripts' && yarn start`

### aragon/aragon

- Checkout this branch: https://github.com/aragon/aragon/tree/enable-aragonjs
- `yarn && yarn start`

### aragonOS template

Install the beta version of [ganache-cli](https://github.com/trufflesuite/ganache-cli):

- `yarn global add ganache-cli@beta`
- `ganache-cli --gasLimit 100000000`

In another shell session:

- Checkout this branch: https://github.com/aragon/aragon-apps/tree/dev-demo-data
- `yarn`
- `cd templates/dev`
- `yarn && yarn migrate`

Note: remember to checkout the `voting-webapp` again if needed (the dev template and the voting app are in the same repository, `aragon/aragon-apps`).

The last lines of the output should look like this:

```
DAO: 0x55cc683f7470b1f87ef1c3df88e99eabb7524850
DAO's voting app: 0xb725c48dd7a77656e92433b3573163aefe57d7bc
DAO's vault app: 0x8eb4df2a9a54fb18813af2eb9e2f0e670df51ee0
ENS: 0x7bdb37e2dc38d0a18a903aa318717aac408d53f6
```

Get the `DAO:` and `ENS:` addresses, and replace them in the [`src/App.js` file of the Aragon project](https://github.com/aragon/aragon/blob/b759743c26451db76e485b5864cdd59eb10aba8a/src/App.js#L22-L23).

The app should now be connected to the node.

## Visuals

image

image

<p align=center>

</p>

image


apps/voting/app/src/components/VotingRow.js
```diff
@@ -0,0 +1,133 @@
...
+
+const BarsGroup = styled.div`
+  width: 100%;
+`
+
+const Bar = styled.div`
+  margin-top: 20px;
+  &:first-child {
+    margin-top: 0;
+  }
```
sohkai: What do you think about reducing this into `BarsGroup`:

```
const BarsGroup = styled.div`
  width: 100%;

  ${ProgressBar}:not(:first-child) {
    margin-top: 20px;
  }
`
```

apps/voting/app/src/components/VotingRow.js
```diff
@@ -0,0 +1,133 @@
+import React from 'react'
+import styled from 'styled-components'
+import { Button, Countdown, TableCell, TableRow } from '@aragon/ui'
+import { VOTE_UNKNOWN } from '../vote-types'
+import ProgressBar from './ProgressBar'
+import VotingStatus from './VotingStatus'
+
+class VotingRow extends React.Component {
```
sohkai: We can do this later, but it'd be interesting to have a interval like `Countdown` that changes the style of a vote that was previously open but then expired while the app was open.

There's a lot of states for this though, e.g. where a vote might become approved while the app is open but we'd only learn of it via a state propagation from the wrapper.

apps/voting/app/src/components/VotingStatus.js
```diff
@@ -0,0 +1,44 @@
...
+import { theme, IconTime, IconCross, IconCheck } from '@aragon/ui'
+
+const STATUSES = {
+  timeout: { label: 'Time out', Icon: IconTime, color: theme.textSecondary },
+  approved: { label: 'Approved', Icon: IconCheck, color: theme.positive },
+  rejected: { label: 'Rejected', Icon: IconCross, color: theme.negative },
+}
+
+const getStatus = (votesYes, votesNo) => {
+  if (votesYes === 0 && votesNo === 0) return STATUSES.timeout
```
sohkai: Wouldn't the timeout depend on the vote's end time rather than not getting an votes?

apps/voting/app/src/components/VotingSummary.js
```diff
@@ -0,0 +1,150 @@
...
+import styled from 'styled-components'
+import { Motion, spring } from 'react-motion'
+import { SidePanel, Text, theme, spring as springConf } from '@aragon/ui'
+import VotingStatus from './VotingStatus'
+
+const { PANEL_INNER_WIDTH } = SidePanel
+
+const fast = springConf('fast')
+
+const VotingBar = ({ votesYes, votesNo, quorum }) => (
```
sohkai: Should this be renamed to `VotingSummary`?

apps/voting/app/src/components/VotingSummary.js
```diff
@@ -0,0 +1,150 @@
...
+    width: 10px;
+    height: 10px;
+    margin-right: 15px;
+    border-radius: 5px;
+    background: ${({ color }) => color};
+  }
+  span {
+    width: 35px;
+    color: ${theme.textSecondary};
+  }
```
sohkai: Should this be part of `&:before` instead?

sohkai:
@bpierre should we also upgrade `styled-components` here and in the wrapper to 3.0?

bpierre:
@sohkai Yes sure! 👍

apps/voting/app/src/App.js
```diff
@@ -0,0 +1,106 @@
...
+    const {
+      votes,
+      votingVisible,
+      currentVoting,
+      createVotingVisible,
+      votingSidebarOpened,
+    } = this.state
+    return (
+      <AragonApp publicUrl="/aragon-ui/">
+        <AppBar
```
sohkai: Should this also be fixed like in the finance app?

apps/voting/app/src/App.js
```diff
@@ -0,0 +1,107 @@
...
+    const vote = votes.find(vote => id === vote.id)
+    if (!vote) return
+    this.setState({
+      currentVoting: vote,
+      votingVisible: true,
+      votingSidebarOpened: false,
+    })
+  }
+  handleDeselectVote = () => {
+    this.setState({ votingVisible: false })
```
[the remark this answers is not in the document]
bpierre: Not yet, because we want to see it during the `SidePanel` closing transition (that’s why there is `votingVisible` instead of relying on `currentVoting` only).

apps/voting/app/src/App.js
```diff
@@ -0,0 +1,107 @@
...
+      createVotingVisible,
+      votingSidebarOpened,
+    } = this.state
+    return (
+      <AragonApp publicUrl="/aragon-ui/">
+        <AppBar
+          title="Voting"
+          endContent={
+            <Button mode="strong" onClick={this.handleCreateVoting}>
+              New Voting
```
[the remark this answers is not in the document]
bpierre: Thanks, I replaced all the instances of “voting” or “poll” by “vote” (except for `VotingButtons`).

apps/voting/app/src/components/VotingRow.js
bpierre: Yes I agree, it would be nice to make the UI react to any change in the state.

apps/voting/app/src/components/VotingRow.js
bpierre: ProgressBar is not a styled-component, so it would require to be wrapped in order to do that. I moved to use `:not()` though, thanks!

apps/voting/app/src/components/VoteSummary.js
```diff
@@ -0,0 +1,166 @@
...
+    <Motion
+      defaultStyle={{ progress: 0 }}
+      style={{ progress: spring(Number(ready), fast) }}
+    >
+      {({ progress }) => (
+        <Main>
+          <Header>
+            <span>
+              <Label>
+                Quorum: <strong>{Math.round(votingPowerSupport * 100)}%</strong>{' '}
```
sohkai: I think this should be `votesYeaPct`?

apps/voting/app/src/vote-types.js
```diff
@@ -1,3 +1,7 @@
-export const VOTE_UNKNOWN = -1
-export const VOTE_NO = 0
-export const VOTE_YES = 1
+export const VOTE_ABSENT = Symbol('VOTE_ABSENT')
+export const VOTE_NAY = Symbol('VOTE_NAY')
+export const VOTE_YEA = Symbol('VOTE_YEA')
+
+export const VOTE_STATUS_ONGOING = Symbol('VOTE_STATUS_ONGOING')
+export const VOTE_STATUS_REJECTED = Symbol('VOTE_STATUS_REJECTED')
+export const VOTE_STATUS_ACCEPTED = Symbol('VOTE_STATUS_ACCEPTED')
```
sohkai: Just thinking these might look better as two object groups: `VOTE_TYPE` and `VOTE_STATUS`.

apps/voting/app/src/components/NewVotePanelContent.js
```diff
@@ -0,0 +1,57 @@
...
+
+const Label = styled(Text).attrs({
+  smallcaps: true,
+  color: theme.textSecondary,
+})`
+  display: block;
+  margin-bottom: 10px;
+`
+
+const Part = styled.div`
```
sohkai: I think `Label` and `Part` aren't being used here.

apps/voting/app/src/components/VotePanelContent.js
```diff
@@ -11,42 +11,35 @@ import {
...
   }
 
-  const { minAcceptQuorumPct: quorum, creator, yea, nay, voters } = vote
-  const userVote = getAccountVote(user.address, voters)
-  const opened = isVoteOpen(vote, voteTime)
+  const { question, quorum, support, endDate } = vote
+  const { creator, yea, nay, totalVoters } = vote.vote
+
+  const creatorName = 'Robert Johnson' // TODO: get creator name
+  const accountVote = VOTE_ABSENT // TODO: detect if the current account voted
```
sohkai: This should make this a lot easier :)

https://github.com/aragon/aragon-apps/pull/103

bpierre: Totally! Thanks 💖

PR #81, opened 2018-01-30
Add Payroll app

bingen:


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+        auth(EMPLOYER_ROLE)
+    {
+        _addEmployee(
+            accountAddress,
+            initialYearlyDenominationSalary,
+            name,
+            getTimestamp()
+        );
+    }
+
```
izqui: I just thought that we could make this contract a forwarder, that forwards scripts to all employees in an org. We need to add the finance app to the blacklist so they can't steal all the moneyz though (See token manager forwarding!)

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+     * @param accountAddress Employee's address to receive Payroll
+     * @param initialYearlyDenominationSalary Employee's salary
+     * @param name Employee's name
+     * @param startDate It will actually set initial lastPayroll value
+     */
+    function addEmployeeWithNameAndStartDate(
+        address accountAddress,
+        uint256 initialYearlyDenominationSalary,
+        string name,
+        uint256 startDate
```
izqui: We need to pass the startDate as a parameter to employer role too

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+        employees[employeeId].denominationTokenSalary = yearlyDenominationSalary.toSecondDenominationToken();
+        yearlyTotalPayroll = yearlyTotalPayroll.add(employees[employeeId].denominationTokenSalary.toYearlyDenomination());
+    }
+
+    /**
+     * @dev Remove employee from Payroll
+     * @notice Updates also global Payroll salary sum
+     * @param employeeId Employee's identifier
+     */
+    function removeEmployee(uint256 employeeId) public auth(EMPLOYER_ROLE) {
```
izqui: We need to either force a payday() when the employee is removed or allow the employee to get payday() until removal after the fact

izqui: Lets make this a different role because `EMPLOYER_ROLE` will now have parameters specific to hiring. We can rename roles both also.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+        delete employeeIds[employees[employeeId].accountAddress];
+        delete employees[employeeId];
+        numEmployees--;
+    }
+
+    /**
+     * @dev Payable function to receive ETH
+     * @notice Payable function to receive ETH
+     */
+    function addFunds() public payable {
```
izqui: Are we keeping functions to add funds in this contract? Shouldn't that logic be deferred directly to the finance app with a escapeHatch to it here

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+        return true;
+    }
+
+    /**
+     * @dev Get number of employees in Payroll
+     * @notice Get number of employees in Payroll
+     * @return Number of employees
+     */
+    function getEmployeeCount() public view returns(uint256 count) {
+        count = numEmployees;
```
izqui: I like explicit returns better for situations like this one

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+
+        return true;
+    }
+
+    /**
+     * @dev Get number of employees in Payroll
+     * @notice Get number of employees in Payroll
+     * @return Number of employees
+     */
+    function getEmployeeCount() public view returns(uint256 count) {
```
izqui: style: `returns (uint256 count)`

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+        view
+        returns(
+            address accountAddress,
+            uint256 yearlyDenominationSalary,
+            string name,
+            uint lastAllocation,
+            uint lastPayroll
+        )
+    {
+        var employee = employees[employeeId];
```
izqui: Explicit types are better than relying on solc to infer it

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+
+        allocation = employee.allocation[token];
+    }
+
+    /**
+     * @dev payday To withdraw monthly payment by employee (the caller)
+     * @notice Only callable once a month. Assumed token has these standard checks implemented:
+     *         https://theethereum.wiki/w/index.php/ERC20_Token_Standard#How_Does_A_Token_Contract_Work.3F
+     */
+    function payday() public {
```
izqui: This function is buried too deep, we try to follow this as much as possible: http://solidity.readthedocs.io/en/develop/style-guide.html#order-of-functions

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+        var employee = employees[employeeIds[msg.sender]];
+        // check that employee exists (and matches)
+        require(employee.accountAddress == msg.sender);
+
+        allocation = employee.allocation[token];
+    }
+
+    /**
+     * @dev payday To withdraw monthly payment by employee (the caller)
+     * @notice Only callable once a month. Assumed token has these standard checks implemented:
```
izqui: Outdated comment

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+     *         https://theethereum.wiki/w/index.php/ERC20_Token_Standard#How_Does_A_Token_Contract_Work.3F
+     */
+    function payday() public {
+        var employee = employees[employeeIds[msg.sender]];
```
izqui: explicit type

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+        employee.lastPayroll = getTimestamp();
+    }
+
+    /**
+     * @dev Change employee account address. To be called by Employer.
+     * @notice Change employee account address
+     * @param employeeId Employee's identifier
+     * @param newAddress New address to receive payments
+     */
+    function changeAddressByOwner(uint256 employeeId, address newAddress) public auth(EMPLOYER_ROLE) {
```
izqui: I don't like this because it allows the employer to 'steal' funds from an employee that hasn't done payday() in a while

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+     *         Only callable once every 6 months
+     * @param tokens Array with the tokens to receive, they must belong to allowed tokens for employee
+     * @param distribution Array (correlated to tokens) with the proportions (integers over 100)
+     */
+    function determineAllocation(address[] tokens, uint256[] distribution) public {
+        var employee = employees[employeeIds[msg.sender]];
+        // check that employee exists (and matches)
+        require(employee.accountAddress == msg.sender);
+        // check that enough time has gone by
+        require(getTimestamp() > employee.lastAllocation && getTimestamp() - employee.lastAllocation > 15768000); // half a year in seconds
```
izqui: lets remove the last allocation time

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+            accountAddress: accountAddress,
+            denominationTokenSalary: initialYearlyDenominationSalary.toSecondDenominationToken(),
+            lastAllocation: 0,
+            lastPayroll: startDate,
+            name: name
+        });
+        // Ids mapping
+        employeeIds[accountAddress] = employeeId;
+        // update global variables
+        yearlyTotalPayroll = yearlyTotalPayroll.add(employees[employeeId].denominationTokenSalary.toYearlyDenomination());
```
izqui: Do we need to make this a storage value? Seems like something that would be expensive to compute otherwise (loops involved) but do we use it in the logic of the contract anywhere or is it just 'called' to it?

future-apps/payroll/contracts/Payroll.sol
izqui: Also should be external

future-apps/payroll/contracts/Payroll.sol
bingen: So, you mean allowing employees to change allocation whenever they want?

future-apps/payroll/contracts/Payroll.sol
bingen: Well, yes, but then should we remove the account address parameter from the addEmployee function and let the employee to set it? This would complicate too much the flow, I think. Anyway, I agree with removing this function, as employee can do it so it's not necessary.

future-apps/payroll/contracts/Payroll.sol
bingen: Yes, the storage was intended to avoid the loop. We use it for `calculatePayrollRunway` and `calculatePayrollBurnrate`.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,569 @@
...
+    /**
+     * @dev Add tokens to the allowed set
+     * @notice Add tokens to the allowed set
+     * @param _allowedTokens Array of new tokens allowed for payment
+     */
+    function addAllowedTokens(address[] _allowedTokens) public auth(EMPLOYER_ROLE) {
+        for (uint i = 0; i < _allowedTokens.length; i++) {
+            if (!allowedTokens[_allowedTokens[i]]) {
+                allowedTokens[_allowedTokens[i]] = true;
+                allowedTokensArray.push(_allowedTokens[i]);
```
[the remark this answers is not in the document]
bingen: It's used for loops in `payday` and `calculatePayrollRunway` functions. Indeed, doesn't look like very gas efficient but I wasn't able to come up with a better solution. I'll think about it...

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,472 @@
...
+        string name;
+    }
+
+    uint128 public nextEmployee; // starts at 1
+    mapping(uint128 => Employee) private employees;
+    mapping(address => uint128) private employeeIds;
+
+    Finance public finance;
+    ERC20 public denominationToken;
+    EtherToken public etherToken;
```
izqui: No longer needed in this app

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,472 @@
...
+
+    event EmployeeAdded(
+        uint128 employeeId,
+        address accountAddress,
+        uint256 initialYearlyDenominationSalary,
+        string name,
+        uint256 startDate
+    );
+
+    event Fund (address sender, address token, uint amount, uint balance, bytes data);
```
izqui: `event Fund(` no spaces before args please

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,472 @@
...
+    }
+
+    /**
+     * @dev Sends ETH to Finance. This contract should never receive funds,
+     *      but in case it happens, this function allows to recover them.
+     * @notice Allows to send ETH from this contract to Finance, to avoid locking them in contract forever.
+     */
+    function escapeHatch() external {
+        // convert ETH to EtherToken
+        etherToken.wrapAndCall.value(this.balance)(address(finance), "Adding Funds");
```
izqui: I think it makes more sense to move this to the Finance app and remove all ether logic from this contract.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -0,0 +1,472 @@
...
+        runScript(_evmScript, input, blacklist);
+    }
+
+    function isForwarder() public pure returns (bool) {
+        return true;
+    }
+
+    function canForward(address _sender, bytes) public view returns (bool) {
+        // check that employee exists (and matches)
+        return (employees[employeeIds[_sender]].accountAddress == msg.sender);
```
izqui: `== _sender`

future-apps/payroll/contracts/Payroll.sol
bingen: So you mean sending ETH to finance? Right now Finance doesn't have a `payable` function. Are you thinking of adding one to wrap ETH and send it to Vault?

izqui: yes

Issue #83, opened 2018-02-08
Finance: receiveApproval should check from sender


sohkai:
From @ewingrj:

> https://github.com/aragon/aragon-apps/blob/master/apps/finance/contracts/Finance.sol#L146 should have `require(msg.sender == token)`

bingen:
Closed by #85

PR #85, opened 2018-02-12
finance: Restrict receiveApproval to be called only by token

bingen:
Otherwise, `receiveApproval` can move anyone's money that has given a 0xFFF approval to the org.
Fixes issue #83 
Compare to PR #84

### DIFF
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -152,6 +152,7 @@ contract Finance is AragonApp, ERC677Receiver {
         transitionsPeriod
         external
     {
+        require(msg.sender == _token);
         ERC20 token = ERC20(msg.sender);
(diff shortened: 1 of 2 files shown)


izqui:
OMG @sohkai it is impossible to match PR sizing to your liking ;)

Tests for happy and failure path are needed and then we can merge

bingen:
Haha, @sohkai sorry my bad: I meant to pick only one or the other, not both, but I should have explained better. Of course we can do both, but I think it's redundant. Let me know which option you guys like better, I'll add the test to the chosen one and close the discarded one (if any).

sohkai:
> it is impossible to match PR sizing to your liking ;)

;)

Ohh, I thought both were going in; I'd say this is still a good sanity check to do (if the token doesn't send itself as the `token` it's probably an attack).

PR #87, opened 2018-02-12
DevTemplate: Test IPFS deployment for Vault and Voting and split MiniMeFactory for gas reasons

izqui:


templates/dev/migrations/2_deploy.js
```diff
@@ -5,20 +5,25 @@ const daoFactoryMigration = require('@aragon/os/migrations/3_factory')
 
 const DevTemplate = artifacts.require('DevTemplate')
 const Voting = artifacts.require('@aragon/apps-voting/contracts/Voting')
+const Vault = artifacts.require('@aragon/apps-voting/contracts/Vault')
+const MiniMeTokenFactory = artifacts.require('@aragon/os/lib/minime/MiniMeTokenFactory')
 
-const appURI = appName => 'file:' + path.resolve('../../', 'apps', appName)
+const votingIpfs = 'ipfs:QmV5sEjshcZ6mu6uFUhJkWM5nTa53wbHfRFDD4Qy2Yx88m'
+const vaultIpfs = 'ipfs:QmPhd9aQGGoHJNcVVAmbzudShvhpYmyKxLWgqLoqSGD81G'
```
sohkai: Maybe we can put these into a constants file (e.g. `dev/ipfs/latest.js`) somewhere so they're easy to find / change?

izqui: This is a super dev environment. This will change as soon as `aragon-dev-cli` is ready.

PR #88, opened 2018-02-12
Tokens template

izqui:
Easy way to deploy tokens to be used in test networks.

Will need to publish a version of the npm package after merge.

### DIFF
--- a/templates/tokens/migrations/2_deploy.js
+++ b/templates/tokens/migrations/2_deploy.js
@@ -1,0 +1,25 @@
+const path = require('path')
+
+const TokenFactory = artifacts.require('TokenFactory')
+
+const tokens = [
+    { name: 'Aragon', symbol: 'ANT' },
+    { name: 'Decentraland', symbol: 'MANA' },
+    { name: 'Bitconnect', symbol: 'BCC' },
+    { name: 'Spankchain', symbol: 'SPANK' },
+    { name: 'Status', symbol: 'SNT' },
+    { name: 'District0x', symbol: 'DNT' },
+    { name: '0x', symbol: 'ZRX' },
+    { name: 'MakerDAO', symbol: 'MKR' },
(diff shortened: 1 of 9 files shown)


templates/tokens/migrations/2_deploy.js
```diff
@@ -0,0 +1,25 @@
...
+]
+
+module.exports = async (deployer, network, accounts) => {
+    const fac = await TokenFactory.new()
+
+    console.log('fac', fac.address)
+    console.log(await Promise.all(tokens.map(async ({ name, symbol }) => {
+        const receipt = await fac.newToken(name, symbol)
+        return receipt.logs.filter(l => l.event == 'DeployToken')[0].args.token
+    })))
```
sohkai: I'm wondering if we can write to `templates/tokens/index.js` with this so it can automatically update it on each migration (assuming we'll be doing this more than once). It'd also be nice if the `index.js` had the token name associated with the address.

Here's what I'm thinking:

```diff
module.exports = async (deployer, network, accounts) => {
    const fac = await TokenFactory.new()
 
    console.log('fac', fac.address)
+   console.log(await Promise.all(tokens.map(async token => {
+        const { name, symbol } = token
         const receipt = await fac.newToken(name, symbol)
+        const addr = receipt.logs.filter(l => l.event == 'DeployToken')[0].args.token
+        token.address = addr
+        return addr
    })))
+
+   const indexObj = {
+     [network]: { // Assuming network === 'rinkeby'
+       tokens,
+       factory: fac,
+     }
+   }
+
+   const indexFile = 'module.exports = ' + JSON.stringify(indexObj, null, 2) // could also use https://github.com/yeoman/stringify-object if you wanted single quotes
+   fs.writeFileSync('../../index.js', indexFile)
}
```

izqui: Good idea, added it

PR #92, opened 2018-02-15
Issue89: change indentation from 4 to 2 spaces (.js .sol .json)

renexdev:


renexdev:
@izqui

bingen:
I think we agreed to wait until we have all other PRs closed. Besides, this is now mixing 4 and 2 spaces.

renexdev:
Sorry I've already pushed on Friday ... It's ok for me if we wait other PR's and then we add the new indentations....

bingen:
Thanks, sounds good to me. I guess we should try to schedule it somehow, otherwise new PRs could come in. Let's wait for @izqui to orchestrate it a bit.

renexdev:
Agree, In the meantime fortunately there is a lot to learn and do =)

izqui:
Given the solidity convention is to use 4 spaces, I'm closing this PR. Specially since given all the changes since it was open, we would need to completely re-do it. Thanks for the contribution regardless @renexdev.

PR #93, opened 2018-02-15
Pin truffle to 4.0.5

sohkai:
Truffle@4.0.6 upgrades solidity to 0.4.19, which breaks our fixed contracts.


bingen:
This looks good to me. Payroll app was already pinned, but to 4.0.4. For the sake of unicity we could update to 4.0.5 too.

PR #91, opened 2018-02-14
Implement and deploy Aragon 0.5 Beta templates

izqui:
See [readme](https://github.com/aragon/aragon-apps/blob/beta-template/templates/beta/readme.md)

PR #95, opened 2018-02-16
Dev template: log installed apps

sohkai:
Makes it easier to Do Things<sup>TM</sup> with the DAO, e.g. set up new votes during the migration.


izqui:
Cool, we should do this in the beta templates in #91 too

PR #97, opened 2018-02-17
Dev template migration: add some votes

bpierre:


sohkai:
Wondering if this would be better as a "demo" template instead? We could leave the "dev" one as a blank state and use the "demo" one when we want pre-populated state.

bpierre:
@sohkai Good idea! I moved it to a demo template.

PR #98, opened 2018-02-20
voting: Add custom duration to new votes

bingen:
If not provided, global value will be added like before.

### DIFF
--- a/apps/voting/contracts/Voting.sol
+++ b/apps/voting/contracts/Voting.sol
@@ -27,6 +27,7 @@ contract Voting is IForwarder, AragonApp {
     struct Vote {
         address creator;
         uint64 startDate;
+        uint64 endDate;
         uint256 snapshotBlock;
         uint256 minAcceptQuorumPct;
         uint256 yea;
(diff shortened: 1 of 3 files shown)


bingen:
So, an organization could end up having several Voting apps, each one with several Votes inside? This sounds like a very bad UX to me. But of course if that's the underlying design principle I will stick to it.
About the min quorum, although it's unrelated to this PR, I would have thought the opposite, that the idea was to be able to increase it for very critical votes where larger consensus is wanted. Good to know what was the intention.

izqui:
> an organization could end up having several Voting apps, each one with several Votes inside? This sounds like a very bad UX to me

That's the cost for having tiny simple apps that do one thing only, but really well. By composing many apps with some ACL rules you can build an infinitely complex corporate structure. 

Also the design of the core contracts shouldn't impact the UX, the UI will be built to make this easy to use, but our job is to make robust smart contracts, not contracts that make the frontend easy to build 😉

izqui:
If there are no further comments, I think we should close this PR

bingen:
Really good point about the danger of creating short-time votes! Although for that I would just define a minimum global value and allow to extend it for each particular one.
I don't think that allowing to change the voting period breaks the old Unix principle, actually the opposite. But I agree that on the trade-off between robustness and flexibility we should favour the first one.
Anyway, fine to close this one.

PR #99, opened 2018-02-22
voting: Check minAcceptQuorum

bingen:
It doesn't make sense a Vote with minimum threshold of 0 (it would
always pass).
Besides, added some tests for wrong initializations.

### DIFF
--- a/apps/voting/test/voting.js
+++ b/apps/voting/test/voting.js
@@ -238,6 +238,36 @@ contract('Voting App', accounts => {
         })
     })
 
+    context('wrong initializations', () => {
+        beforeEach(async() => {
+            const n = '0x00'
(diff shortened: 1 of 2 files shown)


apps/voting/test/voting.js
```diff
@@ -238,6 +238,36 @@ contract('Voting App', accounts => {
...
+
+        it('fails if min support is greater than 100', () => {
+            const neededSupport = pct16(20)
+            const minimumAcceptanceQuorum = pct16(50)
+            return assertRevert(async() => {
+                await app.initialize(token.address, neededSupport, minimumAcceptanceQuorum, votingTime)
+            })
+        })
+
+        it('fails if min acceptance quorum is greater than min support', () => {
```
sohkai: I think l255 and this line's description should be flipped?

bingen: Ha, good catch, thanks!

PR #102, opened 2018-02-23
[WIP] Move Registry app from aragonlabs/registry to future-apps

izqui:
Basically copies @onbjerg's registry app from https://github.com/aragonlabs/registry for easy use from aragon-apps templates to build a TCR template.

Also minor changes so it works with aragonOS 3.

Needs testing, documentation and review.

### DIFF
--- a/future-apps/registry/README.md
+++ b/future-apps/registry/README.md
@@ -1,7 +1,0 @@
-# dai.how
-
-## Build
-
-```bash
-npm i && truffle migrate && npm run build
-```
--- a/future-apps/registry/README.md
+++ b/future-apps/registry/README.md
@@ -1,0 +1,7 @@
+# dai.how
+
+## Build
+
+```bash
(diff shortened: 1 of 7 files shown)


onbjerg:
that's a complex smart contract, I approve

izqui:
We should refactor it soon, the technical debt this contract will bring is going to be a huge weight

tarrencev:
I've been thinking through the implementation of an app that uses this registry as a way to deduplicate data between multiple instances of the app (since it is content addressed). I think in order for this to work the registry needs to implement reference counting. Otherwise it's possible to have app1 and app2 both depending on some entry in the registry and app1 naively deleting the entry while app2 still depends on it. Does that make sense or am I missing something?

izqui:
Not super relevant to have in aragon-apps given that the TCR implementation has been tabled for now.

Issue #49, opened 2018-01-04
Enhance bylaw logic for Voting app to allow more customized quorum and approval rules


stellarmagnet:
## Problem
Based on the Voting specification: http://wiki.aragon.one/dev/apps/voting/

It seems that DAOs will be forced to measure "quorum" based on the number of acceptance votes, as opposed to more standard logic for what "quorum" actually means. 

Whenever you use "Quorum" terminology in the Aragon app, it will be very confusing if you are treating Quorum to mean "minimum acceptance quorum", as this does not match up to the definition of quorum for organizations/businesses, which is:

"Fixed minimum number of eligible members or stockholders (shareholders) who must be present (physically or by proxy) at a meeting before any official business may be transacted or a decision taken therein becomes legally binding." Read more here: http://www.businessdictionary.com/definition/quorum.html

## Solution
In addition to this problem with the terminology, this quorum/acceptance logic should be more flexible. I see three use cases for how a DAO can formulate metrics for approving proposals:

### Disable Quorum
If a DAO does not care to consider "Quorum" as a metric, then it can be disabled in the bylaws, and Approval percentage based on total token supply can be used. Quorum terminology should never be used in DAOs that do not care about true Quorum requirements. [this is equivalent to minimum acceptance quorum]

### Enable Quorum
Whenever Quorum is enabled, this means the Approval Percentage is based on the all of the tokens that voted, and not on the total token supply.

### Modified Quorum
If a DAO wants to be more flexible with how proposals are passed, then allow a feature called "Modified Quorum" which will serve as a conditional logic to determine when the Quorum metric should be considered and how it is treated in relation to Approval Percentage. For example, the DAO can define two ways for proposals to pass. After the time period has ended, if either condition is met, the proposal is approved:
- if Quorum percentage is met & Approval Percentage (based on quorum) is met
- if Yes votes = Quorum Percentage * Approval Percentage [this is equivalent to the "minimum acceptance quorum" variable used in the current Aragon logic]

Here is an example of how you can accommodate for the third use case. When I have more time I can try to simplify the language & mockup a screen that would accommodate to all 3 use cases I described above, but if you shared your current Bylaw screens (if you have any), that'd be helpful.

---
![image](https://user-images.githubusercontent.com/2584493/34556538-8c5961f2-f0eb-11e7-9358-1f053e80d70c.png)

lyricalpolymath:
I like the direction and I would take it a little bit further:
[...]

PR #105, opened 2018-02-24
Voting quorum

jvluso:
#49


izqui:
Hi, first thanks for your contribution! :) 

However this particular issue is something we feel strongly against. Please refer to this conversation to see why we won't add this functionality even though it seems a good and obvious idea: https://github.com/aragon/aragon-apps/pull/98

bingen:
Kind of late, sorry, but also I think this is redundant, if I understood correctly the purpose of it. Right now we have `Support` and `Quorum`:
```
S = Yay's / Vote's voters
Q = Yay's / Total Voters
```
What you are adding is, let's call it, Participation:
```
P = Vote's voters / Total Voters
```
But it turns out that `Q = S * P`.

PR #109, opened 2018-02-28
Token Manager

bpierre:
Normal Mode:

image

Group Mode:

image


izqui:
Awesome work @bpierre! We can remove everything regarding 'wrapped tokens' as that concept no longer exists :)

onbjerg:
Looks amazing! Where does the token names come from?

izqui:
`Foo` should come from the token symbol which should be provided as app context.

sohkai:
Oh snap, I thought the distribution graph wouldn't make it for this release 🎉!

apps/token-manager/app/public/index.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    <meta name="theme-color" content="#000000">
+    <title>Aragon Voting</title>
```
sohkai: Aragon Token Manager

apps/token-manager/app/src/components/HolderRow.js
```diff
@@ -0,0 +1,40 @@
...
+
+class HolderRow extends React.Component {
+  static defaultProps = {
+    name: '',
+    balance: 0,
+    groupMode: false,
+  }
+  static ActionLabel = styled.span`
+    margin-left: 15px;
+  `
```
sohkai: Do you prefer this to having an outside component (or was there another advantage to adding it inside the component)?

This seems more complicated to me, especially with the `Self` reference.

apps/token-manager/app/src/components/SideBar.js
```diff
@@ -0,0 +1,198 @@
...
+  }
+}
+
+const Main = styled.aside`
+  flex-shrink: 0;
+  flex-grow: 0;
+  width: 260px;
+  margin-left: 30px;
+  min-height: 100%;
+  li {
```
sohkai: Maybe move this down into `InfoRow` and `StakesListItem`?

apps/token-manager/app/src/components/HolderRow.js
bpierre: Yes I was trying this, I was waiting for your feedback ha ha 😆

I’ll go back to the previous style.

The idea was to move everything inside the class, to make it clearer when we have multiple components in a single file (`HolderRow` was in `screens/Holders.js`, then I moved it).

But I agree it’s not as readable, and it’s probably a good thing to make multiple component files look bad, so we are pushed to use a single component per file.

sohkai: > Make it clearer when we have multiple components in a single file

I like solving this by just moving out those multiple components into other files :).

PR #111, opened 2018-03-01
[WIP] Templates tests

bingen:
There are a couple of issues yet, and I could add some more tests, but you can start reviewing if you want.


package.json
```diff
@@ -17,7 +17,7 @@
   "scripts": {
     "install": "npm run bootstrap",
     "bootstrap": "lerna bootstrap --nohoist=@aragon/os",
-    "test": "lerna run --scope=@aragon/apps-* --concurrency=1 --stream test",
+    "test": "lerna run --scope=@aragon/apps-* --scope=@aragon/templates-beta --concurrency=1 --stream test",
```
izqui: are we sure we can pass 2 scope arguments? worried it might override it

templates/beta/contracts/MultisigTemplate.sol
```diff
@@ -4,25 +4,51 @@ import "./BetaTemplateBase.sol";
...
+            10**18
+        );
 
         uint256 multisigSupport = neededSignatures * 10 ** 18 / signers.length;
-        voting.initialize(token, multisigSupport, multisigSupport, uint64(-1));
+        voting.initialize(
+            token,
+            multisigSupport,
+            multisigSupport,
+            uint64(157680000) // 5 years
```
izqui: you can use the `5 years` constant directly

bingen: Fixed in #112

package.json
bingen: Yes, it worked for me. I actually added it for lint and coverage too. See first one for instance:
```
lerna success run Ran npm script 'lint' in packages:
lerna success - @aragon/apps-fundraising
lerna success - @aragon/apps-payroll
lerna success - @aragon/apps-finance
lerna success - @aragon/apps-token-manager
lerna success - @aragon/apps-vault
lerna success - @aragon/apps-voting
lerna success - @aragon/templates-beta
```

bingen:
Btw, I'm closing this PR as it's going to be replaced by another one once #112 is merged so I can add tests on top of it.

PR #112, opened 2018-03-02
Template fixes

bingen:
Minor fixes for Beta Templates:
- Add solium
- Add InstalledApp event
- Fix overflow bug in Beta Base for Finance period bug
- Fix decimals bug in Multisig Token
- Fix overflow bug in Multisig for voting period


templates/beta/contracts/BetaTemplateBase.sol
```diff
@@ -42,43 +52,118 @@ contract BetaTemplateBase {
         appIds = _appIds;
     }
 
-    function createDAO(string name, MiniMeToken token, address[] holders, uint256[] stakes, uint256 _maxTokens) internal returns (Voting) {
+    function latestVersionAppBase(bytes32 appId) public view returns (address base) {
+        Repo repo = Repo(PublicResolver(ens().resolver(appId)).addr(appId));
+        (,base,) = repo.getLatest();
+
+        return base;
```
sohkai: Don't need the return here; base is automatically returned.

templates/beta/.soliumrc.json
```diff
@@ -0,0 +1,21 @@
...
+    "camelcase": ["error"],
+    "uppercase": 0,
+    "no-empty-blocks": ["error"],
+    "no-unused-vars": ["error"],
+    "quotes": ["error"],
+    "indentation": 0,
+    "whitespace": ["error"],
+    "deprecated-suicide": ["error"],
+    "pragma-on-top": ["error"],
+    "security/enforce-explicit-visibility": ["error"]
```
sohkai: It might be useful to include `"arg-overflow": ["error", 8]` to this list, to avoid having to change all the functions calls into multi-line calls.

templates/beta/contracts/BetaTemplateBase.sol
```diff
@@ -42,43 +52,118 @@ contract BetaTemplateBase {
...
+            true
+        );
 
         for (uint256 i = 0; i < holders.length; i++) {
             tokenManager.mint(holders[i], stakes[i]);
         }
 
         // inits
-        finance.initialize(vault, etherToken, uint64(-1)); // yuge period
+        finance.initialize(vault, etherToken, uint64(-1) - uint64(now)); // yuge period
```
sohkai: 👍Awesome! Using safemath in the Finance app also makes a lot of sense to catch stuff like this.

templates/beta/contracts/Imports.sol
```diff
@@ -1,10 +0,0 @@
-pragma solidity 0.4.18;
-
-// HACK to workaround truffle artifact loading on dependencies
-
-import "@aragon/os/contracts/factory/ENSFactory.sol";
-import "@aragon/os/contracts/factory/APMRegistryFactory.sol";
-import "@aragon/os/contracts/factory/EVMScriptRegistryFactory.sol";
-import "@aragon/os/contracts/common/EtherToken.sol";
-
-contract Imports {}
```
sohkai: Is this workaround not needed anymore?

templates/beta/contracts/MultisigTemplate.sol
```diff
@@ -4,25 +4,53 @@ import "./BetaTemplateBase.sol";
...
         }
 
         MiniMeToken token = popTokenCache(msg.sender);
-        Voting voting = createDAO(name, token, signers, stakes, 1);
+        Voting voting = createDAO(
+            name,
+            token,
+            signers,
+            stakes,
+            10**18
```
sohkai: Mmm I think the intent was to create a non-transferrable "multisig" token here (so `maxToken <= 1`). That's why there was no decimals in the token and all the stakes were 1.

Was there a problem initializing it like it was before?

templates/beta/contracts/BetaTemplateBase.sol
bingen: Well, I just copy-pasted it from below, changing the order of the functions, but I can change that too.

templates/beta/.soliumrc.json
bingen: Is this the style we use? I copied the file from `voting` app. I see we only have that `arg-overflow` directive in `finance`. Should I add it to all apps? I absolutely fine with any style, but I think it should be consistent across all our repos/apps.

templates/beta/contracts/MultisigTemplate.sol
bingen: Yes, the problem is because of this line:
https://github.com/aragon/aragon-apps/blob/master/templates/beta/contracts/MultisigTemplate.sol#L25
As we are dividing, we need decimals.

templates/beta/.soliumrc.json
izqui: The issue I have with `arg-overflow` is that there are instances where it doesn't make sense since the line is perfectly readable, and splitting into lines actually makes the code worse. I support making it 8 arguments or so as @sohkai proposes and then having a common-sense/line character count rule.

templates/beta/contracts/BetaTemplateBase.sol
```diff
@@ -42,43 +52,118 @@ contract BetaTemplateBase {
...
-        acl.createPermission(ANY_ENTITY, voting, voting.CREATE_VOTES_ROLE(), voting);
-        acl.createPermission(voting, voting, voting.MODIFY_QUORUM_ROLE(), voting);
-
-        acl.createPermission(finance, vault, vault.TRANSFER_ROLE(), voting);
-        acl.createPermission(voting, finance, finance.CREATE_PAYMENTS_ROLE(), voting);
-        acl.createPermission(voting, finance, finance.EXECUTE_PAYMENTS_ROLE(), voting);
-        acl.createPermission(voting, finance, finance.DISABLE_PAYMENTS_ROLE(), voting);
-        acl.createPermission(voting, tokenManager, tokenManager.ASSIGN_ROLE(), voting);
-        acl.createPermission(voting, tokenManager, tokenManager.REVOKE_VESTINGS_ROLE(), voting);
+        acl.createPermission(
```
izqui: This is a clear case where the line split makes the template less readable. From having a list of rules with the same arguments roughly in the same location visually, to something impossible to reason.

templates/beta/contracts/BetaTemplateBase.sol
izqui: I think explicit returns are a good idea besides one liners :)

templates/beta/contracts/BetaTemplateBase.sol
```diff
@@ -42,43 +52,118 @@ contract BetaTemplateBase {
         appIds = _appIds;
     }
 
-    function createDAO(string name, MiniMeToken token, address[] holders, uint256[] stakes, uint256 _maxTokens) internal returns (Voting) {
+    function latestVersionAppBase(bytes32 appId) public view returns (address base) {
```
izqui: I understand why this has been reordered to be here. But I wonder whether we should, in exceptions like this one, keep the main contract (even if internal functions) on top and then start with the rule. Actually `latestVersionAppBase` is only a public function for convenience reasons (as it could perfectly be internal and the frontend could calculate it or use another contract for the purpose.)

templates/beta/contracts/MultisigTemplate.sol
izqui: Actually and sorry for being wrong the other day, I think the voting app should work even with simpler numbers because of how it does the calculations https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol#L277

templates/beta/.soliumrc.json
bingen: Ok, adding that rule, also to other apps for consistency.

templates/beta/contracts/BetaTemplateBase.sol
bingen: Ok, reverting last commit then ;-)

templates/beta/contracts/BetaTemplateBase.sol
bingen: Ok, reverting those changes. I actually agree in this case, I just changed it because of solium complains ;-)

templates/beta/contracts/BetaTemplateBase.sol
bingen: `solium` throws only a warning, so putting it back to its original position.

templates/beta/contracts/Imports.sol
bingen: I guess so. I removed it because `solium` was complaining for having an empty contract here.

templates/beta/contracts/MultisigTemplate.sol
bingen: Sorry for being wrong first ;-) And thanks @sohkai for pointing it out! Reverting those changes too.

templates/beta/contracts/BetaTemplateBase.sol
sohkai: We can also ignore specific lines (or files) for solium: see [docs](http://solium.readthedocs.io/en/latest/user-guide.html#configuring-with-comments).

templates/beta/contracts/Imports.sol
sohkai: If it's working, it'll be awesome to get rid of these hacks. Otherwise, we can ask solium to ignore this file.

bingen: Tests were passing without it.

templates/beta/contracts/BetaTemplateBase.sol
bingen: Nice, added to remove the annoying warning too. Thanks!

PR #115, opened 2018-03-05
Local deployment of beta templates using Docker

izqui:
Trying it out:

- Install [Docker CE](https://docs.docker.com/install/)
- `cd templates/beta && npm run docker:run`
- Outputted ENS address has to be provided to the client
- That's really it 🦅🚀

(Removes templates/beta build artifacts from git)


templates/beta/package.json
```diff
@@ -1,22 +1,28 @@
...
   "devDependencies": {
     "@aragon/id": "^1.2.0",
     "eth-ens-namehash": "^2.0.8",
-    "truffle": "^4.0.5",
+    "truffle": "4.0.5",
     "truffle-hdwallet-provider": "0.0.3"
   },
   "dependencies": {
-    "@aragon/os": "^3.0.3",
+    "@aragon/os": "https://github.com/aragon/aragonOS.git#docker-deploy",
```
bingen: I don't think that's a good idea. What if `docker-deploy` branch gets deleted?

izqui: good catch, was for local testing. releasing aOS version now that has been merged.

izqui:
@bingen We need the build directory or at least the ABI for both templates that get exposed through `index.js`

bingen:
Oh, ok, I was not aware of that. Then I'm going to change PR #117 as I removed all under `build` folder.

izqui:
Cool. Maybe there is a cleaner solution we can do as part of `prepublish` and just extract the ABIs

izqui:
Just did exactly that @bingen see e44fd

.dockerignore
```diff
@@ -0,0 +1,19 @@
...
+*/**/scTopics
+
+
+apps/**/artifact.json
+future-apps/**/artifact.json
+
+# lerna
+lerna-debug.log
+
+# truffle build artifacts
```
bingen: Can not be `**/*/build` added here too?
What about `.git`?

templates/beta/scripts/deploy.sh
```diff
@@ -0,0 +1,18 @@
...
+export ENS=$(npm run deploy:devnet:ens | tail -n 1) # get last line of output
+echo "Using ENS ${ENS}"
+npm run deploy:devnet:apm
+# extract and set ENS
+
+cd ../aragon-id
+npm run deploy:devnet
+
+cd ../aragon-apps
+env DEBUG=true npm run publish:devnet
```
bingen: This is not working for me if I try to build (it worked when I used docker registry published image). I realized that there this line is commented out:
```
% docker run --env-file .env --network beta_devnet --network beta_ipfs --link ipfs aragon/templates-beta cat scripts/deploy.sh
#!/bin/bash

set -e

cd ../../../aragonOS
export ENS=$(npm run deploy:devnet:ens | tail -n 1) # get last line of output
echo "Using ENS ${ENS}"
npm run deploy:devnet:apm
# extract and set ENS

cd ../aragon-id
npm run deploy:devnet

cd ../aragon-apps
# env DEBUG=true npm run publish:devnet

cd templates/beta
npm run deploy:devnet
```

bingen: This is not working for me if I try to build (it worked when I used docker registry published image). I realized that there this line is commented out:
```
% docker run --env-file .env --network beta_devnet --network beta_ipfs --link ipfs aragon/templates-beta cat scripts/deploy.sh
#!/bin/bash

set -e

cd ../../../aragonOS
export ENS=$(npm run deploy:devnet:ens | tail -n 1) # get last line of output
echo "Using ENS ${ENS}"
npm run deploy:devnet:apm
# extract and set ENS

cd ../aragon-id
npm run deploy:devnet

cd ../aragon-apps
# env DEBUG=true npm run publish:devnet

cd templates/beta
npm run deploy:devnet
```

.dockerignore
izqui: good idea

izqui: done

templates/beta/scripts/deploy.sh
izqui: okey done

PR #122, opened 2018-03-07
Bump aragon/os version in templates

sohkai:
At least 3.0.6 is required for the changes in `Voting` and `Finance` to compile.

### DIFF
--- a/templates/beta/package.json
+++ b/templates/beta/package.json
@@ -16,7 +16,7 @@
     "truffle-hdwallet-provider": "0.0.3"
   },
   "dependencies": {
-    "@aragon/os": "^3.0.3",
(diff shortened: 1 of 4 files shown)


templates/beta/package.json
```diff
@@ -16,7 +16,7 @@
     "truffle-hdwallet-provider": "0.0.3"
   },
   "dependencies": {
-    "@aragon/os": "^3.0.3",
+    "@aragon/os": "^3.0.8",
```
bingen: It's 3.0.6, right? Same for the other ones.

izqui: we are in 3.0.8 now, see #115

PR #123, opened 2018-03-07
Voting: add bg script

sohkai:
Requires #120 and #121 (and lots of aragon.js updates).

Nice to dos later:

* [ ] Use a safe div operator that returns `0` instead of `NaN` if dividing by 0.
* [ ] Send notifications


apps/voting/app/src/script.js
```diff
@@ -0,0 +1,154 @@
...
+ *                     *
+ *       Helpers       *
+ *                     *
+ ***********************/
+
+function loadVoteData(voteId) {
+  return new Promise(resolve => {
+    combineLatest(
+      app.call('getVote', voteId),
+      app.call('getVoteMetaData', voteId)
```
izqui: contract function is `voting.getVoteMetadata`

PR #135, opened 2018-03-15
finance: Adapt to new Vault

bingen:
Besides, rename IConnector interface to IVaultConnector

### DIFF
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -15,7 +15,7 @@ contract Finance is AragonApp {
     using SafeMath for uint256;
     using SafeMath64 for uint64;
 
-    address constant ETH = address(0);
+    address constant public ETH = address(0);
     uint64 constant public MAX_PAYMENTS_PER_TX = 20;
     uint64 constant public MAX_PERIOD_TRANSITIONS_PER_TX = 10;
     uint64 constant public MAX_UINT64 = uint64(-1);
@@ -147,7 +147,6 @@ contract Finance is AragonApp {
         ERC20(_token).approve(address(vault), _amount);
         // finally we can deposit them
         vault.deposit(_token, this, _amount, new bytes(0));
-        //vault.deposit(_token, msg.sender, _amount, new bytes(0));
     }
 
     /**
@@ -409,7 +408,7 @@ contract Finance is AragonApp {
         return settings.periodDuration;
     }
 
-    function getBudget(address _token) transitionsPeriod public view returns (uint256 budget, bool hasBudget, uint256 remainingBudget) {
+    function getBudget(address _token) transitionsPeriod public returns (uint256 budget, bool hasBudget, uint256 remainingBudget) {
         budget = settings.budgets[_token];
         hasBudget = settings.hasBudget[_token];
         remainingBudget = _getRemainingBudget(_token);
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -15,6 +15,7 @@ contract Finance is AragonApp {
     using SafeMath for uint256;
     using SafeMath64 for uint64;
 
+    address constant ETH = address(0);
     uint64 constant public MAX_PAYMENTS_PER_TX = 20;
     uint64 constant public MAX_PERIOD_TRANSITIONS_PER_TX = 10;
     uint64 constant public MAX_UINT64 = uint64(-1);
@@ -101,8 +102,8 @@ contract Finance is AragonApp {
      *      to Vault.
      * @notice Allows to send ETH from this contract to Vault, to avoid locking them in contract forever.
      */
-    function escapeHatch() public payable {
-        vault.deposit.value(msg.value)(address(0), msg.sender, 0, new bytes(0));
+    function () public payable {
+        vault.deposit.value(this.balance)(ETH, msg.sender, this.balance, new bytes(0));
     }
 
     /**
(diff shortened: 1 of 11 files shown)


apps/finance/contracts/Finance.sol
```diff
@@ -105,23 +102,20 @@ contract Finance is AragonApp, ERC677Receiver {
      * @notice Allows to send ETH from this contract to Vault, to avoid locking them in contract forever.
      */
     function escapeHatch() public payable {
-        // convert ETH to EtherToken
-        etherToken.wrapAndCall.value(this.balance)(address(this), "Adding Funds");
+        vault.deposit.value(msg.value)(address(0), msg.sender, 0, new bytes(0));
```
izqui: Better to use an `ETH` constant than passing around `address(0)` everywhere

izqui: Value must be `msg.value`, this code should only work when `msg.value == 0`

https://github.com/aragon/aragon-apps/pull/129/files#diff-9ebb8e9a2c4b877825d0aeece0fb63feR14

apps/finance/contracts/Finance.sol
```diff
@@ -139,14 +133,14 @@ contract Finance is AragonApp, ERC677Receiver {
...
-    function deposit(ERC20 _token, uint256 _amount, string _reference) external transitionsPeriod {
+    function deposit(address _token, uint256 _amount, string _reference) external transitionsPeriod {
         _recordIncomingTransaction(
             _token,
             msg.sender,
             _amount,
             _reference
         );
-        require(_token.transferFrom(msg.sender, address(vault), _amount));
+        vault.deposit(_token, msg.sender, _amount, new bytes(0));
```
izqui: So now before depositing tokens using Finance you need to create an allowance to the Vault? (otherwise there is no way this works haha). This should be made very clear in the documentation. Also pinging @onbjerg and @sohkai  who are working in the frontend and this could be a blocker

apps/finance/contracts/Finance.sol
```diff
@@ -166,32 +160,35 @@ contract Finance is AragonApp, ERC677Receiver {
         external
```
izqui: I would remove this function entirely now that we have the new vault

apps/finance/contracts/Finance.sol
```diff
@@ -166,32 +160,35 @@ contract Finance is AragonApp, ERC677Receiver {
...
-            amount,
-            string(data)
+            msg.sender,
+            _from,
+            _amount,
+            string(_userData)
         );
-        require(token.transfer(address(vault), amount));
-        return true;
+        vault.deposit(msg.sender, _from, _amount, _userData);
```
izqui: Doesn't finance need to create an allowance or setOperator here for this to work?

apps/finance/contracts/Finance.sol
```diff
@@ -325,12 +322,15 @@ contract Finance is AragonApp, ERC677Receiver {
         require(value > 0);
```
izqui: we can remove `ERC20 token = ERC20(_token);` above

apps/finance/test/finance.js
```diff
@@ -55,13 +60,13 @@ contract('Finance App', accounts => {
     })
 
     it('records ERC20 deposits', async () => {
-        await token1.approve(app.address, 5)
+        await token1.approve(vault.address, 5)
```
izqui: 👍

apps/finance/contracts/Finance.sol
```diff
@@ -105,23 +102,20 @@ contract Finance is AragonApp, ERC677Receiver {
      * @notice Allows to send ETH from this contract to Vault, to avoid locking them in contract forever.
      */
     function escapeHatch() public payable {
```
izqui: We should make this function `function () payable { vault.deposit.value(this.balance)(ETH, msg.sender, this.balance, new bytes(0)) } `

izqui: rather than escapeHatch

apps/finance/contracts/Finance.sol
bingen: Yes, this is not very user friendly, I'm changing it to get the tokens to finance app, allow to Vault and then deposit, in the way `receiveApproval` was done.

apps/finance/contracts/Finance.sol
bingen: :+1:

apps/finance/contracts/Finance.sol
bingen: Sure!

apps/finance/contracts/Finance.sol
bingen: Yes, well, as ERC777 is not part of the release and Vault doesn't fully support it yet, there was no tests for it .. so even the name and the signature are wrong! :(
Wouldn't something like this work?
```
ERC777(msg.sender).send(adress(vault), _amount, _userData);
```
Anyway, maybe I could just comment it out until we have `ERC777` on Vault.

apps/finance/contracts/Finance.sol
bingen: So true! :100:

apps/finance/contracts/Finance.sol
```diff
@@ -74,8 +73,7 @@ contract Finance is AragonApp, ERC677Receiver {
         mapping (address => bool) hasBudget;
     }
 
-    Vault public vault;
-    EtherToken public etherToken;
+    IVaultConnector public vault;
```
sohkai: Seems a bit odd that `vault` is an `IVaultConnector`; I haven't looked yet, but I guess `Vault` builds on that interface too?

apps/finance/contracts/Finance.sol
```diff
@@ -104,24 +102,21 @@ contract Finance is AragonApp, ERC677Receiver {
      *      to Vault.
      * @notice Allows to send ETH from this contract to Vault, to avoid locking them in contract forever.
      */
-    function escapeHatch() public payable {
-        // convert ETH to EtherToken
-        etherToken.wrapAndCall.value(this.balance)(address(this), "Adding Funds");
+    function () public payable {
+        vault.deposit.value(this.balance)(ETH, msg.sender, this.balance, new bytes(0));
```
sohkai: Somewhat confused about this—do we want this to be the fallback (rather than an explicit "escape hatch")?

Also, should it be `this` and not `msg.sender`, e.g.:

`vault.deposit.value(this.balance)(ETH, this, this.balance, new bytes(0));`

apps/finance/contracts/Finance.sol
```diff
@@ -139,60 +134,43 @@ contract Finance is AragonApp, ERC677Receiver {
...
             _reference
         );
-        require(_token.transferFrom(msg.sender, address(vault), _amount));
+        // first we need to get the tokens to Finance
+        ERC20(_token).transferFrom(msg.sender, this, _amount);
+        // and then approve them to vault
+        ERC20(_token).approve(address(vault), _amount);
+        // finally we can deposit them
+        vault.deposit(_token, this, _amount, new bytes(0));
+        //vault.deposit(_token, msg.sender, _amount, new bytes(0));
```
sohkai: Should this commented out line be removed?

apps/finance/contracts/Finance.sol
```diff
@@ -429,7 +409,7 @@ contract Finance is AragonApp, ERC677Receiver {
         return settings.periodDuration;
     }
 
-    function getBudget(address _token) transitionsPeriod public returns (uint256 budget, bool hasBudget, uint256 remainingBudget) {
+    function getBudget(address _token) transitionsPeriod public view returns (uint256 budget, bool hasBudget, uint256 remainingBudget) {
```
sohkai: This can't be `view`; `transitionsPeriod` potentially changes state.

apps/finance/test/finance.js
```diff
@@ -2,6 +2,9 @@ const { assertRevert, assertInvalidOpcode } = require('@aragon/test-helpers/asse
 const getBalance = require('@aragon/test-helpers/balance')(web3)
 
 const Vault = artifacts.require('Vault')
+const ETHConnector = artifacts.require('ETHConnector')
+const ERC20Connector = artifacts.require('ERC20Connector')
+//const IVaultConnector = artifacts.require('IVaultConnector')
```
sohkai: Should we remove this?

apps/finance/test/finance.js
```diff
@@ -13,10 +16,13 @@ contract('Finance App', accounts => {
     const periodDuration = 100
     const withdrawAddr = '0x0000000000000000000000000000000000001234'
 
+    const ETH='0x0'
```
sohkai: Can we get this from Finance as `Finance.ETH()` (if we made that public)?

apps/finance/contracts/Finance.sol
sohkai: Hmm, I think this was always the case, to be honest. We could get away with it in one call if the token had `approveAndCall` / ERC677, but I guess that's not in here anymore.

apps/finance/contracts/Finance.sol
sohkai: Pinging @izqui, who came up with this change [earlier](https://github.com/aragon/aragon-apps/pull/135#discussion_r175007166).

apps/finance/contracts/Finance.sol
bingen: Ya, it looks odd, but it's actually cool: `Vault` doesn't implement it, but fallback is doing `delegateFwd` to it. This way final functions can be called without the need of implementing them in `Vault`. It was @izqui 's suggestion, so pinging him here.

apps/finance/contracts/Finance.sol
bingen: sure, thanks!

apps/finance/contracts/Finance.sol
bingen: Oh, yes, thanks! I changed because I was having problems with some test, but reverting it now.

apps/finance/test/finance.js
bingen: :+1:

apps/finance/test/finance.js
bingen: I tried, but it doesn't work for me from Truffle. I get:
```
TypeError: Finance.ETH is not a function
```

apps/finance/contracts/Finance.sol
bingen: Yes, but what was weird (from UX perspective) was to approve to `Vault` but to send to `Finance`. Have a look at how it is now, please, and let me know your thoughts.

izqui:
I consider Brett's comments addressed. Merging to my branch

PR #136, opened 2018-03-16
Token Manager: add bg script

sohkai:
screen shot 2018-03-16 at 4 36 30 am

Holy crap, can you imagine this is coming live from a blockchain near you??

Currently, assigning tokens is failing due to https://github.com/aragon/aragon.js/issues/91; because we're using a different web3 provider for aragon.js, the wrong accounts are used to calculate transaction pathing (and we don't yet have a signer error state for when no paths are found... so it all breaks).

Requires #132, #133.


apps/token-manager/app/src/script.js
```diff
@@ -0,0 +1,166 @@
+import Aragon from '@aragon/client'
+import tokenSettings, { hasLoadedTokenSettings } from './token-settings'
+import minimeArtifact from '@aragon/os/build/contracts/MinimeToken.json'
```
sohkai: This would be much better if it were just the ABI, rather than the entire truffle artifact.

Pinging @ji to bring an `abi` export to aragonOS 😊

PR #141, opened 2018-03-16
Changes to radspec descriptions

bingen:


izqui:
why is it a WIP?

bingen:
Haha, because I had those issues testing the strings with radspec, so I was not sure if it was correct.

apps/finance/contracts/Finance.sol
```diff
@@ -134,7 +134,7 @@ contract Finance is AragonApp, ERC677Receiver {
 
     /**
     * @dev Deposit for ERC20 approved tokens
-    * @notice Send `_amount` `_token`
+    * @notice Send `_amount` `_token.symbol()`
```
onbjerg: Needs to be suffixed with `: string`, same goes for every other place you do `_token.symbol()`. This is because Radspec has no ABIs at all, so it needs to be inlined.

apps/voting/contracts/Voting.sol
```diff
@@ -76,7 +76,7 @@ contract Voting is IForwarder, AragonApp {
     }
 
     /**
-    * @notice Change minimum acceptance quorum to `_minAcceptQuorumPct`
+     * @notice Change minimum acceptance quorum to `(_minAcceptQuorumPct - _minAcceptQuorumPct % 10**14) / 10**16`
```
onbjerg: Exponentation is `^`, not `**`

PR #146, opened 2018-03-19
templates: Change geth docker image data path

bingen:
So it's the default and there's no need to add anything to `geth attach`.

### DIFF
--- a/templates/beta/docker-compose.yml
+++ b/templates/beta/docker-compose.yml
@@ -19,7 +19,7 @@ services:
   devnet:
     image: aragon/geth-devnet
     volumes:
-      - ${ETHDATA_LOCAL_PATH}:/root/.ethdata
+      - ${ETHDATA_LOCAL_PATH}:/root/.ethereum
     ports:
       - "8545:8545"
       - "8546:8546"


templates/beta/docker-compose.yml
```diff
@@ -19,7 +19,7 @@ services:
   devnet:
     image: aragon/geth-devnet
     volumes:
-      - ${ETHDATA_LOCAL_PATH}:/root/.ethdata
+      - ${ETHDATA_LOCAL_PATH}:/root/.ethereum
```
bpierre: Shall we set a user in aragon/geth-devnet, instead of running geth as root?

bingen: Yes, but that image (`aragon/geth-devnet`) Dockerfile is not in this repo, right? So that change needs to be made somewhere else.

bpierre: Yes sorry, I was talking about https://github.com/aragon/geth-devnet

bingen: So maybe we can just merge this one for now and leave that change for another PR

bpierre: Yes 👌

PR #153, opened 2018-03-19
Apps: publish 1.0.3

sohkai:
I've added the hand-fixed version of Finance's artifacts.json for now, so we can deploy off of that.


izqui:
why is the artifact committed just for finance?

sohkai:
Because the dev-cli's artifact includes wrong information (see https://github.com/aragon/aragon-dev-cli/issues/43)

PR #160, opened 2018-03-20
templates: Remove docker deploy image for beta templates

bingen:
Avoids the need to download or build a 3GB image just for deployment and connects to `geth-devnet` image from host directly. (On continued testing can fill up local hard disk. )
Besides, testing is way faster this way.

Needs:
https://github.com/aragon/aragonOS/pull/249
https://github.com/aragon/aragon-id/pull/13


templates/beta/test/beta.js
```diff
@@ -1,406 +0,0 @@
-const { assertRevert } = require('@aragon/test-helpers/assertThrow')
```
sohkai: Are these tests supposed to be removed? Does the docker.js tests supersede these?

bingen: Yes, that's the idea. Old `beta.js` was meant to be run on ganache, deploying everything from scratch. Now we use `scripts/deploy.sh` to test, which previously does all the needed deployment.

templates/beta/scripts/deploy.sh
```diff
@@ -10,6 +10,7 @@ npm run deploy:devnet:apm
 cd -
 
 cd node_modules/@aragon/id
+npm install
```
sohkai: Ahh, I think I need to push my local version of npm installing here.

PR #161, opened 2018-03-20
Voting: launch fixes

sohkai:
Fixes for launch:

* Removes dummy data
* Fixes NaNs from showing up
* Asks contracts about user balances and if they can vote
* Use rinkeby etherscan URL

### DIFF
--- a/apps/voting/app/src/components/VotePanelContent.js
+++ b/apps/voting/app/src/components/VotePanelContent.js
@@ -1,28 +1,81 @@
 import React from 'react'
+import PropTypes from 'prop-types'
 import styled from 'styled-components'
 import Blockies from 'react-blockies'
 import {
   Button,
   Info,
+  SafeLink,
   SidePanelSplit,
   SidePanelSeparator,
   Countdown,
   Text,
   theme,
 } from '@aragon/ui'
-import { VOTE_ABSENT, VOTE_NAY, VOTE_YEA } from '../vote-types'
+import { combineLatest } from '../rxjs'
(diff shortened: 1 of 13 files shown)


apps/voting/app/src/components/VotePanelContent.js
```diff
@@ -1,37 +1,63 @@
...
+    app: PropTypes.object.isRequired,
+  }
+  state = {
+    userBalance: null,
+    voterState: null,
+  }
+  componentDidMount() {
+    const { app, user, vote } = this.props
+    app
+      .call('getVoterState', vote.id, user)
```
sohkai: @onbjerg This doesn't seem to be working :(

onbjerg: Just leaving this here for note: we should check if `vote.id` even exists (it's not set in `marshallVote`)

onbjerg:
Can you also update to `@aragon/client` beta.6? It includes a new method we need to use for voting descriptions.

Few changes:

- If `vote.description` does not exist, the description section should be hidden.
- If vote has a script attached to it, set `vote.description` to the value returned by `app.describeScript(vote.script)`

`describeScript` takes in a callscript and returns an observable that emits a single array of objects, where each object is in the following shape:

```js
{
  to: string,
  data: string,
  ?description: string
}
```

The vote description should reflect the description of each item in this array. Note that some items may lack a description, if there is no Radspec string for the method being called on that particular address.

sohkai:
@onbjerg will be done in another PR.

PR #167, opened 2018-03-21
Finance load test tokens

sohkai:
Testnet tokens are now added to the balances list for each DAO automatically on Finance installation and every event thereafter.

### DIFF
--- a/apps/finance/app/src/script.js
+++ b/apps/finance/app/src/script.js
@@ -13,6 +13,8 @@ const app = new Aragon()
 
 // Hook up the script as an aragon.js store
 app.store(async (state, event) => {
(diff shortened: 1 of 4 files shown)


apps/finance/app/src/script.js
```diff
@@ -212,3 +217,36 @@ function marshallTransactionDetails({
     periodId: parseInt(periodId, 10),
   }
 }
+
+/**********************
+ *                    *
+ * RINKEBY TEST STATE *
+ *                    *
+ **********************/
+const TEST_TOKEN_ADDRS = [
```
izqui: Not a blocker for launch, but please import from `@aragon/templates-tokens`

PR #168, opened 2018-03-21
[WIP] Remove abis.json from beta templates

sohkai:
Update beta template's `abis.json` for new Vault and Finance.


bingen:
Wasn't this file supposed to be autogenerated (with `npm run prepublish`) and git-ignored?

sohkai:
Oh was it? I guess I'll add the .gitignore instead?

But we are meant to publish this, yeah?

bingen:
I don't think so, I think that's why `npm run prepublish` was added. Anyway we should either `git rm abis.json` or remove it from `.gitignore`.

izqui:
Close? @sohkai

bingen:
Shouldn't we do `git rm abis.json` before closing?

sohkai:
Yes, I'll do that now!

PR #172, opened 2018-03-21
Radspec voting

onbjerg:
Generates a description from the vote callscript if specified.

We might want to change the description format. Currently each step in the transaction path contains:

```js
{
  to: string,
  data: string,
  ?description: string,
  ?name: string,
  ?identifier: string
}
```

And we render each step on a new line in chronological order, where each like is ``<name> (<address>): <description>``, for example:

>Token Manager (0xbabe): Execute action as token holder
>Token Manager (0xdeadbeef): Mint 1 tokens for 0xebefbbabbe

Note that the above example is super hypothetical

### DIFF
--- a/apps/voting/app/package.json
+++ b/apps/voting/app/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "homepage": "/",
   "dependencies": {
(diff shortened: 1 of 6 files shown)


sohkai:
Tested now, 🤞

PR #178, opened 2018-03-21
Add test task to CI

bingen:


sohkai:
Do we need to rebase this on top of #180?

bingen:
Yes, good idea.

bingen:
Hm, now we have that weird gas error for Vault (it works with `truffle test` and `npm run coverage`, but not for `npm run test`). Will try to fix it.

bingen:
Vault tests are fixed now. I have another PR almost ready that uses docker geth image for vault tests, but relies on #160 so I'll submit it once that one is merged.

PR #180, opened 2018-03-21
Fix broken tests

bingen:
For token manager and voting.


bingen:
They were not passing, see: #178

PR #181, opened 2018-03-22
ethereumjs-abi is needed to test token-manager

gasolin:
Issue:

Can't run `truffle test` successfully because of `ethereumjs-abi` module is missing.

Solution:

Add the missing module via run the command in `apps/token-manager/`:

```sh
yarn add ethereumjs-abi
```

The `truffle test` is passed now.

FYI, `apps/finance` also has `ethereumjs-abi` dev-dependency
https://github.com/aragon/aragon-apps/blob/master/apps/finance/package.json#L43

### DIFF
--- a/apps/token-manager/package.json
+++ b/apps/token-manager/package.json
@@ -40,6 +40,7 @@
     "babel-preset-stage-3": "^6.17.0",
     "babel-register": "^6.26.0",
     "eth-gas-reporter": "^0.1.1",
+    "ethereumjs-abi": "^0.6.4",
     "ethereumjs-testrpc": "^6.0.2",
     "ganache-cli": "^6.0.3",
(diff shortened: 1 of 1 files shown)


bingen:
What problem exactly are you having? I just cloned a fresh copy of the repo, did `yarn install` and then `npm run test` both in token-manager and finance and it worked. I tried with `truffle test` too.

gasolin:
＠bingen I reproduced again by clone another repo and do `yarn install` & `npm run test`. The issue still the same. (in Elementary OS/Ubuntu 16.04 LTS)

bingen:
Hm, that's weird. I don't know why it doesn't happen to me. Can you copy-paste here the exact error you see? Do you have any linked packages in `node_modules/@aragon/` Thanks!

PR #183, opened 2018-03-22
Remove future-apps as lerna detected packages

izqui:
This will make `future-apps` not be run as part of the test suite and whenever we do `lerna exec --scope=@aragon/apps*`.

This is a recurring PITA, given that `future-apps` don't have the required quality for being a full app.


bingen:
Makes sense, especially during release times, but then we could add a task to `package.json` to manually check future apps, something like `npm run test:future`. This way we can easily check if we broke something in future apps during development process.

PR #204, opened 2018-03-28
finance: Record ETH incoming transaction

bingen:


sohkai:
Wondering if it'd be possible to add a test for this, where we send the finance contract some ether and then test that it's recorded a payment :).

PR #220, opened 2018-03-29
Save payment reference as transaction reference in one time payments

izqui:
We only save the payment reference in the tx reference in one time payments (that the payment data structure is not stored). When a saved payment is executed many times, the reference can be fetched from the `paymentId`.

Requires APM redeploy.


izqui:
Dupe #219

PR #232, opened 2018-04-03
Restrict some functions before app is initialized

### LINKED ISSUE #188: Add initialization checks to critical functions in apps

bingen:
Closes #188

### DIFF
--- a/apps/token-manager/test/tokenmanager.js
+++ b/apps/token-manager/test/tokenmanager.js
@@ -331,22 +331,24 @@ contract('Token Manager', accounts => {
     })
 
     context('app not initialized', async () => {
-
         it('fails to mint tokens', async() => {
             return assertRevert(async() => {
                 await tokenManager.mint(accounts[1], 1)
             })
         })
+
         it('fails to assign tokens', async() => {
             return assertRevert(async() => {
                 await tokenManager.assign(accounts[1], 1)
             })
         })
+
         it('fails to issue tokens', async() => {
             return assertRevert(async() => {
                 await tokenManager.issue(1)
             })
         })
+
         it('fails to burn tokens', async() => {
             return assertRevert(async() => {
                 await tokenManager.burn(accounts[0], 1)
--- a/apps/token-manager/test/tokenmanager.js
+++ b/apps/token-manager/test/tokenmanager.js
@@ -329,4 +329,28 @@ contract('Token Manager', accounts => {
             })
         })
     })
+
+    context('app not initialized', async () => {
+
+        it('fails to mint tokens', async() => {
+            return assertRevert(async() => {
+                await tokenManager.mint(accounts[1], 1)
+            })
+        })
+        it('fails to assign tokens', async() => {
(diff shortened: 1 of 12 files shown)


apps/token-manager/test/tokenmanager.js
```diff
@@ -329,4 +329,28 @@ contract('Token Manager', accounts => {
             })
         })
     })
+
+    context('app not initialized', async () => {
+
```
izqui: this newline here kills my eyes lol

barryWhiteHat:
@bingen its still possible to vote before `isInitalized` 

1. Could i vote for `_voteId=0` ?
2. Will it pass because `supportRequiredPct=0`
3. What will happen when it gets executed?

bingen:
@barryWhiteHat Have you tried? I just did but it reverted. Token is not set yet, so any call to it will fail. In particular, I think the following lines won't work:
1. `token.balanceOf` [here](https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol#L157)
2. In the [same line](https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol#L157), `vote.snapshotBlock` will be 0, so even if the token existed the condition would evaluate to false
3. As `vote.startDate` [here](https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol#L280) would be 0 too, unless `voteTime` is huge the condition would be false too.
4. [Here](https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol#L241) same comment as 1 before.
5. As in 2 before, if token existed stake at block 0 would be 0, so vote wouldn't pass because all calls to `_isValuePct` would have `_value = 0`. Therefore `canExecute` would return false too.
6. Finally, even if it passed, `executionScript` would be empty.

barryWhiteHat:
Did it revert because of the `isInitalized` modifier? I mean is it you intention that this call fails or does it fail just by accident because of reasons 1 to 6? I believe you should be explicit about things should fail as much as possible. 

Its good practice to have these things fail explicitly not because something else was not set correctly. Especially if this is meant to be a sample application for other developers to use. They need to understand that the deployment has to be done is a very careful way. AND they should think okay i better harden all my functions so that no one can call them before it gets initialized.

bingen:
> Did it revert because of the isInitalized modifier?
Are you talking about `vote`? It doesn't have that modifier, that's what you are actually asking for, right?

Well, anyway, I just realized that I was actually wrong, the call fails for a more important reason: vote with id 0 doesn't even exist. If app is not initialized, `votes` array is empty. You can check it adding this at the end of `vote.js`:
```
    it('gets vote id 0', async () => {
        console.log(await app.getVote(0))
    })
```
which will revert.
But I see your point. I think this belongs to a more broader discussion I want to have with @izqui and @sohkai to define a general criteria for all apps (e.g., protecting all public functions that modify state)

bingen:
Fixed in #307 . Thanks @barryWhiteHat !

PR #238, opened 2018-04-06
Voting: add signaling multi-option vote support, and some general refactor

### LINKED ISSUE #189: Allow to create a vote and not cast a vote immediately

### LINKED ISSUE #231: Voting: Add Multi-option support
For Signalling to be useful and be able to use the Voting app we need to add support for more than 2 options at the contract level. This will also require frontend changes to display votes properly.

## Spec modification proposal

- Add a `_newSignalingVote(bytes metadata, uint8 options)` function protected by the ACL under the same or a new role. The reason for adding a new function is that multi-option votes **CANNOT** have a script to be executed, because of the cognitive complexity of having different scripts be executed depending on the outcome. And the scamming potential if a no vote can mean different things.

- Signalling votes cannot be executed, which means that a vote cannot be prematurely closed. They will be open for the entire `voteTime`.

- Metadata is an arbitrary byte array that can follow one of the following 3 formats, being the 3rd one the fallback:
  - An `aragon.js` contentURI, currently `ipfs:` and `http:` are the only supported providers. The client will then fetch the voting metadata JSON file at that location (data avail problems).
  - The metadata JSON payload directly (expensive).
  - An arbitrary string that doesn't fall under the above categories which is just displayed as the question.

The format of the metadata JSON could be something like:

```js
{
  "specId": "metadata1.voting.aragonpm.eth",
  "metadata": {
     "question": "when binance?", // required
     "description": "We want to pay **$1m** for the listing", // optional, markdown formatted blob
     "descriptionSource" : "github:issue://aragon/governance#181", // overwrites description when loaded correctly
     "tags": ["exchange-talk", "extortion"],
     "options": ["Yes", "Meh", "Over my dead body"] // optional
  }
}
```

If the options array or an particular option is missing then label options as 'Option 1' to 'Option n'

Tagging @lkngtn and @luisivan for feedback and assigning @chris-remus as the owner of the issue for now.

izqui:
This PR adds multi-option votes for non-binding votes (no script execution) (closes #231)

It also no longer votes on vote creation when a vote is created directly (that is not through the forwarding interface, which still votes on creation) (closes #189)

It is also a general refactor in terms of style mainly. Tests have been minimally modified (apart from adding tests for the new logic).

### DIFF
--- a/apps/voting/test/voting.js
+++ b/apps/voting/test/voting.js
@@ -83,19 +83,19 @@ contract('Voting App', accounts => {
         it('deciding voting is automatically executed', async () => {
             const action = { to: executionTarget.address, calldata: executionTarget.contract.execute.getData() }
             const script = encodeCallScript([action])
-            const voteId = createdVoteId(await app.newVote(script, '', { from: holder50 }))
+            const voteId = createdVoteId(await app.forward(script, { from: holder50 }))
             assert.equal(await executionTarget.counter(), 1, 'should have received execution call')
         })
 
         it('execution scripts can execute multiple actions', async () => {
             const action = { to: executionTarget.address, calldata: executionTarget.contract.execute.getData() }
             const script = encodeCallScript([action, action, action])
-            const voteId = createdVoteId(await app.newVote(script, '', { from: holder50 }))
+            const voteId = createdVoteId(await app.forward(script, { from: holder50 }))
             assert.equal(await executionTarget.counter(), 3, 'should have executed multiple times')
         })
 
         it('execution script can be empty', async () => {
-            const voteId = createdVoteId(await app.newVote(encodeCallScript([]), '', { from: holder50 }))
+            const voteId = createdVoteId(await app.forward(encodeCallScript([]), { from: holder50 }))
         })
 
         it('execution throws if any action on script throws', async () => {
@@ -103,7 +103,7 @@ contract('Voting App', accounts => {
             let script = encodeCallScript([action])
             script = script.slice(0, -2) // remove one byte from calldata for it to fail
             return assertRevert(async () => {
-                await app.newVote(script, '', { from: holder50 })
+                await app.forward(script, { from: holder50 })
             })
         })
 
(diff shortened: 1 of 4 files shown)


apps/voting/test/voting.js
```diff
@@ -318,7 +418,7 @@ contract('Voting App', accounts => {
         })
 
         it('creating vote as holder executes vote', async () => {
```
bingen: I would change this description, as it's no longer the case when creating a "regular" vote.

apps/voting/test/voting.js
```diff
@@ -358,17 +458,18 @@ contract('Voting App', accounts => {
...
         it('creating vote as holder2 executes vote', async () => {
-            const voteId = createdVoteId(await app.newVote(EMPTY_SCRIPT, 'metadata', { from: holder2 }))
+            const voteId = createdVoteId(await app.forward(EMPTY_SCRIPT, { from: holder2 }))
             const [isOpen, isExecuted] = await app.getVote(voteId)
 
             assert.isFalse(isOpen, 'vote should be closed')
             assert.isTrue(isExecuted, 'vote should have been executed')
         })
     })
+
```
bingen: We could add a test to check that now when creating a vote, vote is not casted.

izqui: The test is implicit when state is checked. Because now holder31 is the creator of the vote but the initial state of the vote has 0 casted votes

bingen: Oh, yes, you are right, it's not created by `nonHolder` anymore.

izqui:
Closing as this will no longer happen in the Voting app, rather a separate Survey app (just for non-binding votes) will be created. This new app will also allow holders to divide their voting power among different options, which would require a major refactor in this app for something that wouldn't be allowed anyway in binary, binding votes.

Keeping the branch alive because I believe some tests/logic could be useful.

izqui:
A separate PR should be created to tackle #189 in the Voting app

PR #255, opened 2018-04-16
Run solidity-coverage from ganache-cli.sh

sohkai:
Simplifies our scripting a little bit.

### DIFF
--- a/shared/test-helpers/ganache-cli.sh
+++ b/shared/test-helpers/ganache-cli.sh
@@ -30,8 +30,11 @@ fi
 
 echo "Starting our own testrpc instance at port $testrpc_port"
 start_testrpc
+sleep 5
 
-if [ "$TRUFFLE_TEST" = true ]; then
+if [ "$SOLIDITY_COVERAGE" = true ]; then
+  ./node_modules/.bin/solidity-coverage "$@"
+elif [ "$TRUFFLE_TEST" = true ]; then
   truffle test --network rpc "$@"
-  kill -9 $testrpc_pid
 fi
+kill -9 $testrpc_pid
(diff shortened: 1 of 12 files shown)


shared/test-helpers/ganache-cli.sh
```diff
@@ -30,8 +30,11 @@ fi
 
 echo "Starting our own testrpc instance at port $testrpc_port"
 start_testrpc
+sleep 5
```
sohkai: 🤞 this fixes some of those connection issues.

shared/test-helpers/ganache-cli.sh
```diff
@@ -30,8 +30,11 @@ fi
...
 
-if [ "$TRUFFLE_TEST" = true ]; then
+if [ "$SOLIDITY_COVERAGE" = true ]; then
+  ./node_modules/.bin/solidity-coverage "$@"
+elif [ "$TRUFFLE_TEST" = true ]; then
   truffle test --network rpc "$@"
-  kill -9 $testrpc_pid
 fi
+kill -9 $testrpc_pid
```
sohkai: @bingen I'm pretty sure this should also kill the solidity-coverage testrpc instance?

bingen: I don't think so, as we have `set -o errexit` at the beginning, therefore if a process returns error it will exit before.

bingen: Btw, were you replying to this? ;-)
https://github.com/aragon/aragonOS/pull/291#issuecomment-381430736
It's in a different repo, but the same problem.

sohkai: Something different; was wondering if we're correctly capturing the testrpc instance solidity-coverage starts in `$testrpc_pid` (I don't see why it wouldn't, but you know solidity-coverage...)

sohkai:
Merging this so we can finally start cascading our merges :D.

The tests on CI are still failing due to templates-beta.

PR #256, opened 2018-04-16
Finance: check vault on initialization

sohkai:
> initialize check vault & etherToken != 0. Why not use kernel for vault & etherToken.

In the future, we could also use the kernel's default vault, if no vault is given.

TODO:

* [x] Wait until aragonOS is upgraded, and then rebase ontop of master
* [x] Use `IsContract` to check if the vault is a contract

Note that this is untestable due to the breaking Vault depedency :(.

### DIFF
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -123,6 +123,7 @@ contract Finance is AragonApp {
         initialized();
 
         require(_periodDuration >= 1 days);
+        require(_vault != address(0));
 
         vault = _vault;
 
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -6,6 +6,7 @@ pragma solidity 0.4.18;
 
 import "@aragon/os/contracts/apps/AragonApp.sol";
 
+import "@aragon/os/contracts/common/IsContract.sol";
 import "@aragon/os/contracts/lib/zeppelin/token/ERC20.sol";
 import "@aragon/os/contracts/lib/zeppelin/math/SafeMath.sol";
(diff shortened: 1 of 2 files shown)


apps/finance/contracts/Finance.sol
```diff
@@ -119,6 +119,7 @@ contract Finance is AragonApp {
         initialized();
 
         require(_periodDuration > 1);
+        require(_vault != address(0));
```
bingen: Maybe we could even check that `isContract`.

sohkai: Was thinking the same, but even with `isContract` we can't be sure that it's a Vault... so not sure it's worth the extra gas. We're also not checking that in a few places (e.g. for the kernel in the app proxies).

bingen: Well, I think the same reasoning would apply to several other places where we are using `isContract` and that even not making sure it is a Vault, at least we could avoid some more errors than only checking 0 address (so we increase safety although not up to 100%), and about the gas this function is called only once, but I'm fine leaving it this way.

sohkai: Thinking about it some more, safer is better :). Will wait until `IsContract` is released in a new aragonOS.

apps/finance/contracts/Finance.sol
```diff
@@ -15,7 +16,7 @@ import "@aragon/apps-vault/contracts/IVaultConnector.sol";
 import "@aragon/os/contracts/lib/misc/Migrations.sol";
 
 
-contract Finance is AragonApp {
+contract Finance is IsContract, AragonApp {
```
sohkai: Note that solidity fails to linearize the inheritance tree if it's `AragonApp, IsContract`.

bingen:
Incorporated into #425

PR #257, opened 2018-04-16
Finance: include payment repeat number in transactions

sohkai:
> Payment repeat number is not recorded in the transaction

### DIFF
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -51,6 +51,7 @@ contract Finance is AragonApp {
         uint256 amount;
         uint256 paymentId;
         string reference;
+        uint256 paymentRepeatNumber;
     }
 
     struct TokenStatement {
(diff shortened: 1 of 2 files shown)


apps/finance/contracts/Finance.sol
```diff
@@ -51,6 +51,7 @@ contract Finance is AragonApp {
         uint256 amount;
         uint256 paymentId;
         string reference;
+        uint256 paymentRepeatNumber;
```
bingen: A very picky one, I would put this one line above, under `paymentId`, for consistence with the rest of the code. ;-)

sohkai: Ah—this is the tricky bit; we can't do that safely if we want to ensure people can upgrade up, since the underlying storage slots would stay in the previous order, but we'd have inserted a new element.

Doing it this way wouldn't break access to other variables, but would mean that existing transactions would all have a repeat number of 0.

PR #259, opened 2018-04-18
Allow to create a vote and not cast a vote immediately

### LINKED ISSUE #189: Allow to create a vote and not cast a vote immediately

bingen:
Closes #189.
Mostly imported from commit https://github.com/aragon/aragon-apps/pull/238/commits/3dd43692350140b7307410bd77e7376f63e22900

### DIFF
--- a/apps/voting/contracts/Voting.sol
+++ b/apps/voting/contracts/Voting.sol
@@ -101,7 +101,7 @@ contract Voting is IForwarder, AragonApp {
      * @notice Create a new vote about "`_metadata`"
      * @param _executionScript EVM script to be executed on approval
      * @param _metadata Vote metadata
-     * @param _castVote Wether to also cast newly created vote
+     * @param _castVote Whether to also cast newly created vote
      * @return voteId id for newly created vote
      */
     function newVote(bytes _executionScript, string _metadata, bool _castVote) auth(CREATE_VOTES_ROLE) external returns (uint256 voteId) {
@@ -144,7 +144,7 @@ contract Voting is IForwarder, AragonApp {
     */
     function forward(bytes _evmScript) public {
         require(canForward(msg.sender, _evmScript));
-        uint256 voteId = _newVote(_evmScript, "", true);
+        _newVote(_evmScript, "", true);
     }
 
     function canForward(address _sender, bytes _evmCallScript) public view returns (bool) {
--- a/apps/voting/contracts/Voting.sol
+++ b/apps/voting/contracts/Voting.sol
@@ -19,7 +19,7 @@ contract Voting is IForwarder, AragonApp {
     uint256 public minAcceptQuorumPct;
     uint64 public voteTime;
 
-    uint256 constant public PCT_BASE = 10 ** 18;
+    uint256 constant public PCT_BASE = 10 ** 18; // 0% = 0; 1% = 10^16; 100% = 10^18 
 
     bytes32 constant public CREATE_VOTES_ROLE = keccak256("CREATE_VOTES_ROLE");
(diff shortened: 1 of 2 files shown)


apps/voting/contracts/Voting.sol
```diff
@@ -205,15 +210,6 @@ contract Voting is IForwarder, AragonApp {
...
         StartVote(voteId);
-
-        if (canVote(voteId, msg.sender)) {
-            _vote(
-                voteId,
-                true,
-                msg.sender,
-                true
-            );
-        }
```
sohkai: To avoid making someone do two calls to vote for a vote they create, maybe we could use a bool param to indicate whether they'd like to auto vote or not?

To avoid breaking changes, we could overload `newVote()` with a version that automatically passes true for auto voting.

apps/voting/contracts/Voting.sol
```diff
@@ -264,7 +260,7 @@ contract Voting is IForwarder, AragonApp {
         ExecuteVote(_voteId);
     }
 
-    function _isVoteOpen(Vote storage vote) internal view returns (bool) {
+    function _isVoteOpen(Vote vote) internal view returns (bool) {
```
sohkai: I'm pretty sure the `storage` is a small optimization to avoid copying the entire struct into memory when we call this function. I would hope the compiler errors if we don't pass in a storage variable to this function, but who knows with solc.

apps/voting/test/voting.js
```diff
@@ -83,27 +83,27 @@ contract('Voting App', accounts => {
         it('deciding voting is automatically executed', async () => {
             const action = { to: executionTarget.address, calldata: executionTarget.contract.execute.getData() }
             const script = encodeCallScript([action])
-            const voteId = createdVoteId(await app.newVote(script, '', { from: holder50 }))
+            const voteId = createdVoteId(await app.forward(script, { from: holder50 }))
```
sohkai: It's kind of weird to be replacing `newVote()` with the forwarding functionality; the two should be tested separately (I notice we have just one—"forwarding creates vote"—that should also test the forwarding automatically casts a vote).

Also see comment above on adding a param to `newVote()` instead :).

apps/voting/contracts/Voting.sol
```diff
@@ -94,7 +94,18 @@ contract Voting is IForwarder, AragonApp {
...
     function newVote(bytes _executionScript, string _metadata) auth(CREATE_VOTES_ROLE) external returns (uint256 voteId) {
-        return _newVote(_executionScript, _metadata);
+        return _newVote(_executionScript, _metadata, true);
+    }
+
+    /**
+     * @notice Create a new vote about "`_metadata`"
+     * @param _executionScript EVM script to be executed on approval
+     * @param _metadata Vote metadata
+     * @param _castVote Wether to also cast newly created vote
```
sohkai: Small typo: `Whether`

apps/voting/contracts/Voting.sol
```diff
@@ -126,13 +138,13 @@ contract Voting is IForwarder, AragonApp {
...
     /**
-    * @notice Creates a vote to execute the desired action
+    * @notice Creates a vote to execute the desired action, and casts a support vote
     * @dev IForwarder interface conformance
     * @param _evmScript Start vote with script
     */
     function forward(bytes _evmScript) public {
         require(canForward(msg.sender, _evmScript));
-        _newVote(_evmScript, "");
+        uint256 voteId = _newVote(_evmScript, "", true);
```
izqui: Why do you save the `voteId` here?

PR #280, opened 2018-04-19
TokenManager: let Minime token handle transferable-ness

sohkai:
See https://github.com/aragon/aragon-apps/pull/280#issuecomment-383041298.

~Adds a check to make sure the TokenManager is initialized to be non-transferable if the underlying minime token is also non-transferable.~

~I didn't force equality because the converse is less strict—even if the token was originally transferable, initializing the TokenManager as non-transferable would effectively make the token also non-transferable.~

~Fixes~

> ~check token.transfersEnabled on init~

### DIFF
--- a/apps/token-manager/contracts/TokenManager.sol
+++ b/apps/token-manager/contracts/TokenManager.sol
@@ -62,10 +62,14 @@ contract TokenManager is ITokenController, AragonApp { // ,IForwarder makes cove
         )
         onlyInit external
     {
-
         initialized();
 
         require(_token.controller() == address(this));
+        if (!_token.transfersEnabled()) {
+            // If the token itself isn't transferable, make sure this app is also initialized to not
+            // be transferable
+            require(!_transferable);
+        }
 
         token = _token;
         transferable = _transferable;
--- a/apps/token-manager/contracts/TokenManager.sol
+++ b/apps/token-manager/contracts/TokenManager.sol
@@ -16,7 +16,6 @@ contract TokenManager is ITokenController, AragonApp { // ,IForwarder makes cove
     using SafeMath for uint256;
 
     MiniMeToken public token;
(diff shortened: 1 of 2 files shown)


apps/token-manager/contracts/TokenManager.sol
```diff
@@ -62,10 +62,14 @@ contract TokenManager is ITokenController, AragonApp { // ,IForwarder makes cove
...
     {
-
         initialized();
 
         require(_token.controller() == address(this));
+        if (!_token.transfersEnabled()) {
+            // If the token itself isn't transferable, make sure this app is also initialized to not
+            // be transferable
+            require(!_transferable);
+        }
```
bingen: Is the possibility of using a transferable token as non-transferable supposed to be allowed? What's the use case for that? Isn't it a little bit confusing?
Otherwise, couldn't we just get rid of `_transferable` param on `initialize` function and just use the one from `_token`?

sohkai: What we could do is set the token's transferability in the TokenManager's `initialize()` based on `_transferable`.

Need to think about the implications of that... need to make sure the minting, assigning, etc. still works if we do this.

sohkai:
4fe922b replaces all the `transferable` logic in the TokenManager and instead piggybacks on the Minime token to handle that.

@bingen @izqui The only API change that's affected is the inability for holders of non-transferable tokens to transfer tokens back to the TokenManager (however, the TokenManager can still burn those tokens directly): see https://github.com/aragon/aragon-apps/pull/280/files#diff-17c8771ab9c99065348cbba1b3fba658L79 and https://github.com/aragon/aragon-apps/pull/280/files#diff-17c8771ab9c99065348cbba1b3fba658R104. I'd argue that's more consistent, and fits the use case better.

PR #287, opened 2018-04-24
Create new Survey app, for signaling, non-binding, votes

bingen:
Closes issue #231.


apps/survey/contracts/Survey.sol
```diff
@@ -0,0 +1,186 @@
...
+        return _newSurvey(_metadata, _options);
+    }
+
+    /**
+    * @notice Vote option #`_option` in survey #`_surveyId`.
+    * @dev If voting option is 0, that's like cancelling the previously casted vote
+    * @param _surveyId Id for survey
+    * @param _option Index of supported option
+    */
+    function voteOption(uint256 _surveyId, uint256 _option) public {
```
izqui: Should allow to specify how many tokens the voter is casting for the option. We should also have a multi-cast function so it can be done in just one tx.

See general review comment for more context.

bingen:
You are right! I just forgot, sorry.

apps/survey/contracts/Survey.sol
```diff
@@ -0,0 +1,186 @@
...
+
+import "@aragon/os/contracts/apps/AragonApp.sol";
+
+import "@aragon/os/contracts/lib/minime/MiniMeToken.sol";
+import "@aragon/os/contracts/lib/zeppelin/math/SafeMath.sol";
+import "@aragon/os/contracts/lib/zeppelin/math/SafeMath64.sol";
+import "@aragon/os/contracts/lib/misc/Migrations.sol";
+
+
+contract Surveying is AragonApp {
```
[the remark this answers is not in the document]
bingen: True. It was initially `Survey` (that's why it doesn't match now), but it was throwing a warning, and as `Voting` is not `Vote` I thought this was a good idea.

apps/survey/contracts/Surveying.sol
```diff
@@ -0,0 +1,240 @@
...
+            _voteOption(_surveyId, _options[i], _stakes[i]);
+            // let's avoid repeating an option by
+            // making sure that ascending order is preserved in options array
+            require(survey.voters[msg.sender].options[i] < survey.voters[msg.sender].options[i+1]);
+            totalVoted = totalVoted.add(_stakes[i]);
+        }
+
+        // compute and register non used tokens
+        // implictly we are doing require(totalVoted <= voterStake) too:
+        survey.voters[msg.sender].stakes[0] = voterStake.sub(totalVoted);
```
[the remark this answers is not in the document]
bingen: Not sure what you mean by "magic number" but I wouldn't say this is one. As stated in a previous comment (and in your next one), first index (`0`) for `options` and `stakes` arrays is reserved for `NO_VOTE`, but `0` here means the index, not the value. Although it certainly work too, I would find confusing using `NO_VOTE` here instead of `0`.

apps/survey/contracts/Surveying.sol
```diff
@@ -0,0 +1,240 @@
...
+    * @dev Empty options and stakes arrays allow to remove previous votes
+    * @param _surveyId Id for survey
+    * @param _options Array with indexes of supported options
+    * @param _stakes Number of tokens assigned to each option
+    */
+    function voteOptions(uint256 _surveyId, uint256[] _options, uint256[] _stakes) public {
+        require(_options.length == _stakes.length);
+        require(canVote(_surveyId, msg.sender));
+
+        uint256 i;
```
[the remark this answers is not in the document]
bingen: Because it's being used twice, to avoid using 2 variables. Doing `for (uint256 i =..` in each for loop was throwing an error (`DeclarationError: Identifier already declared`).

apps/survey/contracts/Surveying.sol
```diff
@@ -0,0 +1,240 @@
...
+        uint256 participationPct = survey.participation.mul(PCT_BASE) / survey.votingPower;
+        return participationPct >= survey.minAcceptParticipationPct;
+    }
+
+    function _newSurvey(string _metadata, uint256 _options) isInitialized internal returns (uint256 surveyId) {
+        surveyId = surveys.length++;
+        Survey storage survey = surveys[surveyId];
+        survey.creator = msg.sender;
+        survey.startDate = uint64(now);
+        survey.options = _options;
```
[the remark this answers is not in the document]
bingen: How would check then that casted votes are right, i.e., that `_option <= survey.options`?

apps/survey/contracts/Surveying.sol
```diff
@@ -0,0 +1,240 @@
...
+        Survey storage survey = surveys[_surveyId];
+
+        require(_option != NO_VOTE && _option <= survey.options);
+        require(_stake > 0);
+
+
+        // register voter amount
+        survey.voters[msg.sender].options.push(_option);
+        survey.voters[msg.sender].stakes.push(_stake);
+
```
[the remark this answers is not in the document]
bingen: The problem here was that the function was wrongly set as `public` instead of `internal`. With this function being internal I don't think a situation where those array lengths are different can be reached, right?

apps/survey/contracts/Surveying.sol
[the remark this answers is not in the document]
izqui: You can name it differently, `for (uint j = 1; ...)`. There is no benefit in just using one variable

apps/survey/contracts/Surveying.sol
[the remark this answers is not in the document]
izqui: It wouldn't matter. Voting in an option without metadata could be interpreted as a null vote.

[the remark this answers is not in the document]
izqui: Actually implementing metadata is required for the name of the options, so I would suggest removing the options uint from the survey storage

izqui:
Looks like coverage is failing:

```
lerna ERR! sh: 1: ./node_modules/@aragon/test-helpers/run-coverage.sh: not found
```

apps/survey/contracts/Surveying.sol
```diff
@@ -0,0 +1,256 @@
...
+    MiniMeToken public token;
+    uint256 public minAcceptParticipationPct;
+    uint64 public surveyTime;
+
+    struct OptionCast {
+        uint256 optionId;
+        uint256 stake;
+    }
+
+    /* To allow multiple option votes.
```
izqui: Weird comment format, please start with '/**'

apps/survey/contracts/Surveying.sol
```diff
@@ -0,0 +1,240 @@
...
+
+import "@aragon/os/contracts/apps/AragonApp.sol";
+
+import "@aragon/os/contracts/lib/minime/MiniMeToken.sol";
+import "@aragon/os/contracts/lib/zeppelin/math/SafeMath.sol";
+import "@aragon/os/contracts/lib/zeppelin/math/SafeMath64.sol";
+import "@aragon/os/contracts/lib/misc/Migrations.sol";
+
+
+contract Surveying is AragonApp {
```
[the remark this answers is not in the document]
izqui: 'Surveying' feels weird, lets rename to 'Survey' pls. Do we need to rename the struct?

apps/survey/arapp.json
```diff
@@ -0,0 +1,20 @@
+{
+  "appName": "survey.aragonpm.eth",
+  "version": "1.0.1",
```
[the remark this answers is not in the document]
izqui: ?

apps/survey/contracts/Surveying.sol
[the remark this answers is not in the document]
izqui: there is no issue with that and no benefit in reusing the value. it is just confusing, you can rename the second for loop counter to `j`

apps/survey/contracts/Surveying.sol
```diff
@@ -0,0 +1,256 @@
...
+            // reset previously voted options, if any
+            delete survey.voters[msg.sender];
+        }
+
+        uint256 totalVoted = 0;
+        // reserve first index for NO_VOTE
+        survey.voters[msg.sender].castedVotes[0] = OptionCast({optionId: NO_VOTE, stake: 0});
+        for (i = 0; i < _options.length; i++) {
+            // index 0 in options array is reserved for NO_VOTE, so index will be 1 more
+            _voteOption(_surveyId, i+1, _options[i], _stakes[i]);
```
izqui: `i + 1`

izqui: You can just generate the `OptionCast` instance here and past that to the internal function

apps/survey/contracts/Surveying.sol
bingen: Weird?
http://solidity.readthedocs.io/en/v0.4.23/layout-of-source-files.html?highlight=comments#comments

bingen:
> Do we need to rename the struct?
I'll do, to avoid a Warning. Any suggestion?

apps/survey/contracts/Surveying.sol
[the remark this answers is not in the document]
bingen: What's the benefit of allowing wrong/null options? Will we notify the user that one of chosen options was invalid and therefore didn't count?

bingen:
> Looks like coverage is failing:

Hm, it's weird, it works for me locally (and I tried with a fresh clone) and also here:
https://travis-ci.org/aragon/aragon-apps/jobs/373506739
I'll try to figure out what's going on.

apps/survey/contracts/Survey.sol
```diff
@@ -0,0 +1,250 @@
...
+        votingPower = survey.votingPower;
+        participation = survey.participation;
+        options = survey.options;
+    }
+
+    function getSurveyMetadata(uint256 _surveyId) public view returns (string) {
+        return surveys[_surveyId].metadata;
+    }
+
+    function getVoterState(uint256 _surveyId, address _voter) public view returns (uint256[] options, uint256[] stakes) {
```
sohkai: Wondering if this should be make `external` instead.

I think we keep most of the getters public so that inherited contracts can use them too (usually as a convenience), but I'm worried about this one because it has a potentially unbounded for-loop.

An inherited contract has access to read this data using `surveys[_surveyId].voters[_voter].castedVotes` (and they probably don't need the entire array duplicated for them when they do want to read some part of it).

To me, this is something that just shouldn't be used internally.

apps/survey/contracts/Survey.sol
```diff
@@ -0,0 +1,250 @@
...
+        if (surveys[_surveyId].voters[_voter].optionsCastedLength == 0) {
+            return (new uint256[](0), new uint256[](0));
+        }
+
+        options = new uint256[](surveys[_surveyId].voters[_voter].optionsCastedLength + 1);
+        stakes = new uint256[](surveys[_surveyId].voters[_voter].optionsCastedLength + 1);
+        for (uint256 i = 0; i <= surveys[_surveyId].voters[_voter].optionsCastedLength; i++) {
+            options[i] = surveys[_surveyId].voters[_voter].castedVotes[i].optionId;
+            stakes[i] = surveys[_surveyId].voters[_voter].castedVotes[i].stake;
```
sohkai: It might be nice to use a variable for this particular vote, so it's not getting repeated all the time, e.g.

```sol
MultiOptionVote storage vote = surveys[_surveyId].voters[_voter]
```

apps/survey/contracts/Survey.sol
```diff
@@ -0,0 +1,250 @@
...
+    */
+    function initialize(
+        MiniMeToken _token,
+        uint256 _minAcceptParticipationPct,
+        uint64 _surveyTime
+    ) onlyInit external
+    {
+        initialized();
+
+        require(_minAcceptParticipationPct > 0 && _minAcceptParticipationPct <= PCT_BASE);
```
sohkai: I wouldn't necessarily require min participation to be above 0.

Depends on what the goals are for setting this (most interesting would be if another contract wanted to invoke some action based on the voting results if the participation's been reached), but I could see people not requiring any participation at times.

Also depends on the goals, but it might be interesting declaring the min participation as part of a new survey rather than globally (mostly if there's not going to be a consistent default).

apps/survey/contracts/Survey.sol
```diff
@@ -0,0 +1,250 @@
...
+        survey.participation = survey.participation.add(totalVoted);
+    }
+
+    /**
+    * @notice Vote option #`_option` in survey #`_surveyId`.
+    * @dev It will use the whole balance. If voting option is 0, that's like cancelling the previously casted vote
+    * @param _surveyId Id for survey
+    * @param _option Index of supported option
+    */
+    function voteOption(uint256 _surveyId, uint256 _option) public {
```
sohkai: Any time where we use `_option`, I think it'd be more clear if we used `_optionId`.

apps/survey/contracts/Survey.sol
```diff
@@ -0,0 +1,250 @@
...
+
+    struct OptionCast {
+        uint256 optionId;
+        uint256 stake;
+    }
+
+    /** To allow multiple option votes.
+     * Index 0 will always be for NO_VOTE option
+     * option Ids must be in ascending order
+     */
```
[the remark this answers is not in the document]
bingen: Yes, I was using /* ... */ but @izqui asked me to change it. See his comment here:
https://github.com/aragon/aragon-apps/pull/287#discussion_r185331000
As I have said before, we should have a clear guide of our style, especially when it differs from "official" Solidity one (those comments are perfectly valid according to it: http://solidity.readthedocs.io/en/v0.4.23/layout-of-source-files.html#comments) and enforced with `lint` as much as possible. I feel these kind of discussions distracting from the important things and sometimes even a little bit frustrating.
I personally like `/* ... */` for multi line and `//` for one line, but as I have said before too, I will be glad to adapt to any agreed style if it's clearly stated somewhere.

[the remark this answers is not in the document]
sohkai: Totally agree, although I don't think the linters are super smart about this right now.

Let's use `/*` or `//` for multi line blocks, but definitely not `///` or `/**` since those are **explicitly** for natspec.

apps/survey/contracts/Survey.sol
```diff
@@ -0,0 +1,250 @@
...
+    * @param _options Number of options voters can decide between
+    * @return surveyId id for newly created survey
+    */
+    function newSurvey(string _metadata, uint256 _options) auth(CREATE_SURVEYS_ROLE) external returns (uint256 surveyId) {
+        return _newSurvey(_metadata, _options);
+    }
+
+    /**
+    * @notice Vote for multiple options in survey #`_surveyId`.
+    * @dev Empty options and stakes arrays allow to remove previous votes
```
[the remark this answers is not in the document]
izqui: yes pls, great idea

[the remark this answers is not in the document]
izqui: Then `voteOptions()` can just force a reset before

apps/survey/contracts/Survey.sol
bingen: See this comment too: https://github.com/aragon/aragon-apps/pull/287#discussion_r184844630

apps/survey/contracts/Survey.sol
```diff
@@ -0,0 +1,250 @@
...
+        uint256 stake;
+    }
+
+    /** To allow multiple option votes.
+     * Index 0 will always be for NO_VOTE option
+     * option Ids must be in ascending order
+     */
+    struct MultiOptionVote {
+        uint256 optionsCastedLength;
+        mapping (uint256 => OptionCast) castedVotes;
```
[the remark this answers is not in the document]
bingen: Yes, see comment here: https://github.com/aragon/aragon-apps/pull/287#discussion_r184844754
Adding the comment for now. Let me know if you want me to revert it back to an array.

apps/survey/contracts/Survey.sol
```diff
@@ -0,0 +1,250 @@
...
+        }
+    }
+
+    function getOptionSupport(uint256 _surveyId, uint256 _option) public view returns (uint256) {
+        return surveys[_surveyId].votes[_option];
+    }
+
+    function isParticipationAchieved(uint256 _surveyId) public view returns (bool) {
+        SurveyStruct storage survey = surveys[_surveyId];
+        // votingPower is always > 0
```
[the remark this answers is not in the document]
bingen: Ok, about the assertion, but with the comment here I want to emphasize that the division is correct as the denominator can't be zero.

apps/survey/contracts/Survey.sol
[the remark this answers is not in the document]
sohkai: Ah hmm, what about just having an array of `OptionCast`s (rather than two separate arrays)?

apps/survey/test/survey.js
```diff
@@ -281,4 +281,25 @@ contract('Survey app', accounts => {
       })
     })
   })
+
+  context('worng supply token', () => {
```
sohkai: Small typo 😄

apps/survey/contracts/Survey.sol
sohkai: Cool, that's a good point. We can always use super small participation requirements anyway, so it's not a big deal :).

apps/survey/contracts/Survey.sol
```diff
@@ -204,21 +225,22 @@ contract Survey is AragonApp {
...
         }
     }
 
-    function getOptionSupport(uint256 _surveyId, uint256 _option) public view returns (uint256) {
-        return surveys[_surveyId].votes[_option];
+    function getOptionSupport(uint256 _surveyId, uint256 _optionId) public view returns (uint256) {
+        return surveys[_surveyId].optionVotes[_optionId];
     }
 
     function isParticipationAchieved(uint256 _surveyId) public view returns (bool) {
```
sohkai: I'd rename `survey.minAcceptParticipationPct` to just `survey.minParticipationPct`, since there's no real "acceptance".

apps/survey/contracts/Survey.sol
```diff
@@ -45,8 +46,8 @@ contract Survey is AragonApp {
         uint256 participation;                  // tokens that casted a vote
         string metadata;
 
-        mapping (uint256 => uint256) votes;     // option -> voting power for option
-        mapping (address => MultiOptionVote) voters;    // voter -> options voted, with its stakes
+        mapping (uint256 => uint256) optionVotes;     // option -> voting power for option
```
sohkai: You're probably going to kill me 🙈, but what do you think about just renaming this `optionPower`? `optionVotes` gets me too confused with `votes` half the time.

apps/survey/contracts/Survey.sol
```diff
@@ -98,84 +99,104 @@ contract Survey is AragonApp {
         return _newSurvey(_metadata, _options);
     }
 
+    /**
+     * @notice Reset previously casted vote in survey #`_surveyId`, if any.
+     * @param _surveyId Id for survey
+     */
+    function resetVote(uint256 _surveyId) public {
```
sohkai: Something I forgot about earlier; should we be using the `isInitialized` modifier for all these public functions?

apps/survey/contracts/Survey.sol
```diff
@@ -98,84 +99,104 @@ contract Survey is AragonApp {
...
+        require(canVote(_surveyId, msg.sender));
+
+        SurveyStruct storage survey = surveys[_surveyId];
+        MultiOptionVote storage previousVote = survey.votes[msg.sender];
+        if (previousVote.optionsCastedLength == 0) {
+            return;
```
sohkai: What do you think about using the opposite conditional here, and putting the rest of the logic in that?

E.g.

```
if (previousVote.optionsCastedLength > 0) {
    ... // reset their vote
}
```

I find this structure lets you reason about the functionality a lot easier, especially at a glance. This also lets us avoid conditional comments, like `reset previously voted options, if any`.

apps/survey/contracts/Survey.sol
```diff
@@ -98,84 +99,104 @@ contract Survey is AragonApp {
...
-            for (uint256 i = 1; i <= previouslyVoted.optionsCastedLength; i++) {
-                survey.votes[previouslyVoted.castedVotes[i].optionId] = survey.votes[previouslyVoted.castedVotes[i].optionId].sub(previouslyVoted.castedVotes[i].stake);
-            }
-            // remove previous vote participation
-            survey.participation = survey.participation.sub(voterStake.sub(previouslyVoted.castedVotes[0].stake));
-
-            // reset previously voted options, if any
-            delete survey.voters[msg.sender];
-        }
+        // revert previous votes, if any (also checks if canVote)
```
sohkai: What do you think about making `canVote` a modifier and then using an internal `_resetVote()` function that both `voteOptions()` and `resetVote()` use?

I think it's better to make these requirements as explicit as possible.

apps/survey/contracts/Survey.sol
```diff
@@ -98,84 +99,104 @@ contract Survey is AragonApp {
...
         uint256 totalVoted = 0;
-        // reserve first index for NO_VOTE
-        survey.voters[msg.sender].castedVotes[0] = OptionCast({optionId: NO_VOTE, stake: 0});
-        for (uint256 j = 0; j < _options.length; j++) {
-            require(_options[j] != NO_VOTE && _options[j] <= survey.options);
-            require(_stakes[j] > 0);
+        // reserve first index for ABSTAIN_VOTE
+        survey.votes[msg.sender].castedVotes[0] = OptionCast({optionId: ABSTAIN_VOTE, stake: 0});
+        for (uint256 j = 1; j <= _optionIds.length; j++) {
+            // parameter arrays don't have abstain vote, while survey ones do, so we subtract 1 here to the index
```
sohkai: Had to read this a few times to make sense of it, might be more clear being explicit: `Voters don't specify that they're abstaining, but we still keep track of this by reserving the first index of a survey's votes. We subtract 1 from the indexes of the arrays passed in by the voter to account for this.`

Maybe `j` might also be more descriptive as `optionIndex`?

apps/survey/contracts/Survey.sol
```diff
@@ -98,84 +99,104 @@ contract Survey is AragonApp {
...
+            uint256 optionId = _optionIds[j-1];
+            uint256 stake = _stakes[j-1];
+            require(optionId != ABSTAIN_VOTE && optionId <= survey.options);
+            require(stake > 0);
 
             // register voter amount
-            // index 0 in options array is reserved for NO_VOTE, so index will be 1 more
-            survey.voters[msg.sender].castedVotes[j + 1] = OptionCast({optionId: _options[j], stake: _stakes[j]});
+            // index 0 in options array is reserved for ABSTAIN_VOTE, so index will be 1 more
```
sohkai: Think this comment isn't needed anymore

apps/survey/contracts/Survey.sol
```diff
@@ -98,84 +99,104 @@ contract Survey is AragonApp {
...
 
             // let's avoid repeating an option by
             // making sure that ascending order is preserved in options array
-            // index 0 in options array is reserved for NO_VOTE, so here we are comparing
-            // the option just added (_options[j] = options[j+1]) with the previous one (options[j])
-            require(survey.voters[msg.sender].castedVotes[j].optionId < survey.voters[msg.sender].castedVotes[j + 1].optionId);
-            totalVoted = totalVoted.add(_stakes[j]);
+            // index 0 in options array is reserved for ABSTAIN_VOTE, so here we are comparing
+            // the option just added with the previous one
+            require(survey.votes[msg.sender].castedVotes[j - 1].optionId < survey.votes[msg.sender].castedVotes[j].optionId);
```
sohkai: I think we can actually move this check up to where the other requires are, since we already know `survey.votes[msg.sender].castedVotes[j].optionId`. Then that block will read like we're checking to make sure we can actually add the next option a voter gave.

apps/survey/contracts/Survey.sol
```diff
@@ -98,84 +99,104 @@ contract Survey is AragonApp {
...
+        uint256 voterStake = token.balanceOfAt(msg.sender, survey.snapshotBlock);
+        survey.votes[msg.sender].castedVotes[0].stake = voterStake.sub(totalVoted);
 
         // register number of options voted
-        survey.voters[msg.sender].optionsCastedLength = _options.length;
+        survey.votes[msg.sender].optionsCastedLength = _optionIds.length;
 
         // add voter tokens to participation
         survey.participation = survey.participation.add(totalVoted);
+        require(survey.participation <= survey.votingPower);
```
sohkai: What do you think about making this an `assert()`? If this isn't true, I feel like something horrible has gone wrong (rather than input checking).

apps/survey/test/survey.js
bingen: 'CTRL+t' is the best shortcut ever for me! Haha.

apps/survey/contracts/Survey.sol
bingen: Well, it was initially this way but I changed it to avoid one unnecessary indentation, which TBH I like it better (and it's a common pattern). I see that condition a little bit like the `require` checks at the beginning of a lot of our functions. So the idea is:
```
function doSomething() {
    // check that you actually can do something or have something to do

    // actually
    // do
    // something
}
```
So when I'm reading the code inside the function I can forget about the conditions, I know everything is fine.
Besides, although it's not much the case here (just one `for`) it helps a lot if the main logic has other conditions or logic structures inside.
But anyway, changing it, I said I would adapt to any style you like ;-)
The only thing I don't get is why should I remove that comment. The function will do exactly the same, so it will only reset previously voted options if any existed (otherwise it will do nothing). I mean, I can remove the comment if yo don't like it, but the same way I could remove it now as the function currently is.

apps/survey/contracts/Survey.sol
```diff
@@ -0,0 +1,250 @@
...
+
+        require(_minAcceptParticipationPct > 0 && _minAcceptParticipationPct <= PCT_BASE);
+
+        token = _token;
+        minAcceptParticipationPct = _minAcceptParticipationPct;
+        surveyTime = _surveyTime;
+    }
+
+    /**
+    * @notice Change minimum acceptance participation to `(_minAcceptParticipationPct - _minAcceptParticipationPct % 10^14) / 10^16`%
```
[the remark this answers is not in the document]
bingen: Well, I think the precedence is right, what I'm trying to do is remove decimals. So imagine that we have 66.66% and 100 instead of 10^14, and therefore `_minAcceptParticipationPct` is 6666. So we would do: 6666 - 6666 % 100 = 6666 - 66 = 6600. So when we divide by 100 it's exact. The error is that we have 10^14 and 10^16. Fixing it.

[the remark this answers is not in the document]
bingen: Now I realize that my idea was to keep 2 decimals, that's why I had 10^14 and 10^16, but they were in reverse order.

apps/survey/contracts/Survey.sol
sohkai: For the comment, the `, if any` part makes it sound conditional, but now it isn't :). It'll only reach that block if there were previous votes.

apps/survey/contracts/Survey.sol
bingen: No problem, next time just don't give me options to choose! Hahaha.

apps/survey/contracts/Survey.sol
bingen: I don't get it, so as a general function comment it will apply the same (actually both versions of the function do exactly the same, right?). So *if* there is *any* previous vote, it will reset it, otherwise it won't do anything. 
Maybe it's my English ;-), anyway, removing it, it's just a comment, so it's supposed to be there to help, not to confuse.

bingen: Oh, damn it! Now I see it! You were talking about the comment *inside* the function, while I was talking about the one in the header! Haha. Sorry for the confusion!

apps/survey/contracts/Survey.sol
bingen: Oops, nope. We don't have `survey.votes[msg.sender].castedVotes[optionIndex]` yet. I just moved it a little bit above.

bingen: Oops, sorry, I read your comment too fast. You were absolutely right. Thanks for your fix!!!

PR #289, opened 2018-04-25
Prune old unnecessary dependencies

sohkai:


sohkai:
Only test that's failing is the templates-beta test, so I'm merging.

PR #292, opened 2018-04-28
Replace Vault with old Vault (+ simple ether support)

izqui:
Given the amount of complexity the new Vault introduces and the fact that we won't be using much of the functionality for some time, I propose migrating to a slightly modified version of the old vault.

The API is still the same as the other one and the old vault now can support Ether natively too.

I still think the new Vault is superior and I will continue working on it for staking. This can get audited when the AN is being audited. DAOs can always upgrade to the new Vault in the future.


apps/finance/contracts/Finance.sol
```diff
@@ -108,15 +108,15 @@ contract Finance is AragonApp {
             this.balance,
             "Ether transfer to Finance app"
         );
-        vault.deposit.value(this.balance)(ETH, msg.sender, this.balance, new bytes(0));
+        vault.deposit.value(this.balance)(ETH, this, this.balance);
```
sohkai: Just to give some visibility: I think it's safe to pass along `msg.sender` here if we remove the check in the vault (see below)

apps/vault/contracts/Vault.sol
```diff
@@ -1,118 +1,103 @@
...
-    event NewTokenStandard(uint32 indexed erc, uint32 indexed interfaceDetectionERC, bytes4 indexed interfaceID, address connector);
+    event Transfer(address indexed token, address indexed to, uint256 amount);
+    event Deposit(address indexed token, address indexed sender, uint256 amount);
 
-    function Vault() public {
-        erc20ConnectorBase = new ERC20Connector();
-        ethConnectorBase = new ETHConnector();
+    /**
+    * @dev When used behind an AppProxy, this fallback is never executed
+    *      as it will be intercepted by the Proxy (see aragonOS#281)
```
sohkai: Isn't it still possible to trigger this fallback by sending > 10k gas but no data in a call to the proxy? We should probably reword to `On a normal send() or transfer() this fallback is never executed as it will be intercepted by the Proxy (see aragonOS#281)`

apps/vault/contracts/Vault.sol
```diff
@@ -1,118 +1,103 @@
...
+    * @param token Address of the token being transferred
+    * @param from Entity that currently owns the tokens
+    * @param value Amount of tokens being transferred
+    */
+    function deposit(address token, address from, uint256 value) payable public {
+        require(value > 0);
+
+        if (token == ETH) {
+            // Deposit is implicit in this case
+            require(msg.sender == from && msg.value == value);
```
sohkai: If we remove the `msg.sender == from` check, we can pass along the actual sender when other apps deposit to the vault (see Finance, above).

apps/vault/package.json
```diff
@@ -40,6 +41,6 @@
     "truffle-hdwallet-provider": "0.0.3"
   },
   "dependencies": {
-    "@aragon/os": "^3.1.1"
+    "@aragon/os": "^3.1.9"
   }
```
sohkai: To consider: we should either update the version to this package to be `1.0.1` or `3.0.0`.

I prefer `1.0.1` for the similar interface, despite the opportunity to line up the versions with our `arapp.json` file, which we'd need to increment as well to `3.0.0` to deploy to rinkeby.

apps/finance/contracts/Finance.sol
bingen: What would be the point of passing `msg.sender`?

apps/vault/contracts/Vault.sol
bingen: Actually, should we allow to execute the fallback if it's not behind a proxy? :thinking: What would be the use case for it?

apps/finance/contracts/Finance.sol
izqui: Then we need senders to create the allowance to the Vault, which would allow to bypass the Finance deposit by calling deposit directly in the vault.

apps/vault/package.json
izqui: I imagine you are referring to `@aragon/apps-vault`'s version and not aOS. If so, I support keeping it in version `1.0.1` or `1.1.0` given the nature of the change, even though we didn't break the API.

apps/vault/contracts/Vault.sol
izqui: I'm having some doubts about this. Given that people may create an allowance greater than the funds they want to deposit to the vault (patterns like `0xffff...` allowances and ERC777 'operators' are being used more and more), this presents the risk that someone from the DAO may trigger the function to deposit more tokens than intended by the owner. 

If we leave it like this, then the owner of the tokens must always send the message for the vault to pull the tokens.

apps/finance/contracts/Finance.sol
sohkai: @bingen passing `msg.sender` preserves the actual sender of the funds (rather than finance) when the vault emits its `Deposit` event.

@izqui this would only be for ETH, so no allowance required.

apps/vault/contracts/Vault.sol
sohkai: I think it's still valuable to have the fallback, in the event someone does use the vault without a proxy. Without it, `send()` and `transfer()` wouldn't work.

apps/vault/contracts/Vault.sol
sohkai: This would only be for ETH, so no allowances are necessary.

apps/finance/contracts/Finance.sol
bingen: Yes, it's just that I don't see the point of it, as this is kind of a escape hatch mechanism, so I think it makes sense to see funds coming from Finance. Finance is already recording incoming transaction with the actual sender, so it can be tracked there. But no strong opinion about it, I'm fine if you want to change it.

apps/vault/contracts/Vault.sol
bingen: So, is our Vault supposed to be intended for use without a proxy? This way, couldn't it happen that funds are sent to the base Vault (in an scenario where a proxy is being used) and then are lost?

sohkai: > So, is our Vault supposed to be intended for use without a proxy?

I'd say it's "intended" in a very light way; nothing should break if it's not behind a proxy (things only break if it, or a proxy, is not attached to a kernel).

> couldn't it happen that funds are sent to the base Vault (in an scenario where a proxy is being used) and then are lost

Potentially, if someone got the wrong address for their vault, somehow. We could protect against this by adding an initialization step to the vault, where it only accepts deposits and transfers if it's been initialized.

apps/vault/contracts/Vault.sol
izqui: I agree this check is not needed for ETH, but we should have the `require(msg.sender == from)` check for token transfers to avoid the scenario I explain in my previous comment

izqui: I agree this check is not needed for ETH, but we should have the require(msg.sender == from) check for token transfers to avoid the scenario I explain in my previous comment

sohkai: With https://github.com/aragon/aragon-apps/pull/292#discussion_r196114741, I'm hesitating on this too. I think we can leave this as-is so that the vault logs who was the last "hop" in sending it money.

apps/vault/contracts/Vault.sol
bingen: I like this last option, although I still don't see the point of using a proxy-less Vault, plus the fact (as @izqui said) that anybody can initialize it. Maybe we could do a special initialization for base Vault that forbids deposits, although I think it's unnecessary because as I said before I don't get the use case.

apps/vault/package.json
izqui: I think this could be one of the reasons the CI is failing, in Finance the vault dependency was set to `^2.0.1` and the vault we had in the repo was `1.1.0`.  The issue could be that given that the local version is not able to satisfy the vault dependency, it lerna then tries to fetch it from NPM. This is what it looks like it is happening in the CI as it keeps complaining about the solidity issue we solved here: https://github.com/aragon/aragon-apps/pull/292/commits/3757ec1d526a65cb73b87de8a0ea8bc9e77187b7.

There is another issue with setting the vault version to `1.1.0`, because `2.0.1` is already published in NPM https://www.npmjs.com/package/@aragon/apps-vault and we won't be able to overwrite it when the time comes.

I propose jumping to vault `3.0.0` given that we ended up breaking the API.

sohkai:
On https://github.com/aragon/aragon-apps/pull/292#discussion_r196757403 (using this vault without a proxy):

With the onset of https://github.com/aragon/aragonOS/pull/355#issuecomment-408856961, let's make this `Vault` app only usable behind a proxy.

sohkai:
Superseded by #425.

PR #293, opened 2018-04-28
Graduate Payroll to a proper app status

izqui:
Lets use this PR to review it again before merging


izqui:
Can we increase the coverage @bingen?

bingen:
Sure!

bingen:
I was really surprised of that low coverage (I seemed to recall that this app had a good coverage). It turns out that the problem is again `npm run coverage` not failing on errors (because of lerna?):
https://travis-ci.org/aragon/aragon-apps/jobs/372342379#L1980
... which is actually way worse, as tests are not passing, haha.
Anyway, I'll fix it.

bingen:
And we have the same problem for `npm run test`:
https://travis-ci.org/aragon/aragon-apps/jobs/372342377#L1240

bingen:
Coverage was indeed 100% for this app. A recent change in Finance was making tests fail. It should be good now. We still need to fix the CI (to actually fail when tests are not passing).

izqui:
Closing as we can do this as part of https://github.com/aragon/aragon-apps/pull/361

PR #299, opened 2018-05-10
Templates: Expose IPFS 8080 port

bingen:
This way we can access it from outside and use it with the local apps
generated with aragon-dev-cli.


sohkai:
Dumb question: do we also need to add it to the `expose` list?

bingen:
I think we don't:
https://docs.docker.com/compose/compose-file/#expose
We do want that port in the host machine so the browser can retrieve content. I can confirm that it was working fine for me this way.

PR #300, opened 2018-05-10
Voting minor improvements

bingen:
- Remove unused first index for Votes
- Minor style change
- Simplify `_isValuePct` function

### DIFF
--- a/apps/voting/contracts/Voting.sol
+++ b/apps/voting/contracts/Voting.sol
@@ -71,6 +71,8 @@ contract Voting is IForwarder, AragonApp {
         supportRequiredPct = _supportRequiredPct;
         minAcceptQuorumPct = _minAcceptQuorumPct;
         voteTime = _voteTime;
+
+        votes.length += 1;
     }
 
     /**
--- a/apps/voting/contracts/Voting.sol
+++ b/apps/voting/contracts/Voting.sol
@@ -282,13 +282,16 @@ contract Voting is IForwarder, AragonApp {
     * @dev Calculates whether `_value` is at least a percentage `_pct` of `_total`
     */
     function _isValuePct(uint256 _value, uint256 _total, uint256 _pct) internal pure returns (bool) {
-        if (_value == 0 && _total > 0)
+        if (_total == 0) {
+            return true;
+        }
+
+        if (_value == 0) {
             return false;
+        }
 
-        uint256 m = _total.mul(_pct);
-        uint256 v = m / PCT_BASE;
+        uint256 computedPct = _value.mul(PCT_BASE) / _total;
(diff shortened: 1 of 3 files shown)


apps/voting/contracts/Voting.sol
```diff
@@ -277,20 +275,23 @@ contract Voting is IForwarder, AragonApp {
...
     /**
     * @dev Calculates whether `_value` is at least a percentage `_pct` of `_total`
     */
     function _isValuePct(uint256 _value, uint256 _total, uint256 _pct) internal pure returns (bool) {
-        if (_value == 0 && _total > 0)
+        if (_total == 0) {
+            return true;
+        }
+
+        if (_value == 0) {
```
bingen: I think we should remove this one, but I didn't want to modify original function's behaviour in this commit. What do you think @izqui @sohkai ?

sohkai: I think you can remove this condition; if `_value == 0`, the rest will evaluate to false anyway.

sohkai: I think you can remove this condition; if _value == 0, the rest will evaluate to false anyway.

bingen: Not if `_pct` is zero too. It's an edge case, but if I wanted to keep the function equivalent I had to add this condition. So we could re-formulate the question as: "If `_pct` is zero, should a `_value` of zero return true or false?" I think it should return true, but the original function would return false.

apps/voting/contracts/Voting.sol
```diff
@@ -277,20 +275,23 @@ contract Voting is IForwarder, AragonApp {
...
+        return uint64(now) < vote.startDate.add(voteTime) && !vote.executed;
     }
 
     /**
     * @dev Calculates whether `_value` is at least a percentage `_pct` of `_total`
     */
     function _isValuePct(uint256 _value, uint256 _total, uint256 _pct) internal pure returns (bool) {
-        if (_value == 0 && _total > 0)
+        if (_total == 0) {
+            return true;
```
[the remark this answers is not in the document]
bingen: I agree, but I didn't want to modify function behaviour, like here: https://github.com/aragon/aragon-apps/pull/300#discussion_r187476157
Let's make a decision about that one and I'll push another commit with both.

sohkai:
For https://github.com/aragon/aragon-apps/pull/300#discussion_r187894800 and https://github.com/aragon/aragon-apps/pull/300#discussion_r187790576 (modifying the behaviour of `_isValuePct`), I consider both bugs of the original implementation.

We should fix them and add tests to make sure they stay that way :).

PR #301, opened 2018-05-13
Remove unnecessary babel and webpack in non-UI package.jsons

### LINKED ISSUE #34: Remove babel
When https://github.com/aragon/aragon-core/pull/154 is merged and a new aragon-core release is made, we need to remove unused dependencies from all apps packages.

sohkai:
Closes #34.

Also webpack was causing errors with react-scripts when you use `lerna bootstrap` :(.


sohkai:
@bingen would be great to get a sanity check; do the tests still pass on your side after this? You'll probably have to do a `lerna clean && lerna bootstrap` before checking.

PR #302, opened 2018-05-13
Use relative paths for frontend assets

sohkai:
Just found this for creating relative paths for assets: https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#serving-the-same-build-from-different-paths

Some brief testing convinces me this will probably solve a lot of our asset loading problems with IPFS, but would like another set of eyes :).

### DIFF
--- a/apps/finance/app/package.json
+++ b/apps/finance/app/package.json
@@ -54,5 +54,6 @@
     "last 1 edge versions",
     "not ie > 0",
     "not op_mini all"
-  ]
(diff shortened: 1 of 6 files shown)


sohkai:
We're actually secretly using the hash mode in the web version for things like `aragon run` anyway ;). Assuming this works, we may be able to get out of hosting our apps like `voting.aragonpm.com`.

PR #305, opened 2018-05-14
Remove migrations import in app contracts

sohkai:
Is there a reason they're imported?

Vault's is removed in https://github.com/aragon/aragon-apps/pull/304.


izqui:
It is required in case we need to migrate the apps, but right now I believe we are just doing `Contract.new()` everywhere

izqui:
Actually we added some more convenient and simple deploy scripts for apps: https://github.com/aragon/aragon-apps/blob/master/apps/survey/package.json#L8

izqui:
Actually we added some more convenient and simple deploy scripts for apps: https://github.com/aragon/aragon-apps/blob/master/apps/survey/package.json#L8. 

I have been thinking for a while how much I'd love to remove truffle migrations from every repo.

PR #307, opened 2018-05-14
Add isInitialized to all state-changing public functions

### LINKED ISSUE #303: Double check isInitialized is included for all state-changing public functions

bingen:
Closes #303.

### DIFF
--- a/apps/token-manager/contracts/TokenManager.sol
+++ b/apps/token-manager/contracts/TokenManager.sol
@@ -154,7 +154,7 @@ contract TokenManager is ITokenController, AragonApp { // ,IForwarder makes cove
     * @param _holder Address getting vesting revoked
     * @param _vestingId Numeric id of the vesting
     */
-    function revokeVesting(address _holder, uint256 _vestingId) authP(REVOKE_VESTINGS_ROLE, arr(_holder)) external {
+    function revokeVesting(address _holder, uint256 _vestingId) authP(REVOKE_VESTINGS_ROLE, arr(_holder)) isInitialized external {
         TokenVesting storage v = vestings[_holder][_vestingId];
         require(v.revokable);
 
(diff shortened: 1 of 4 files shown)


apps/token-manager/contracts/TokenManager.sol
```diff
@@ -351,7 +351,7 @@ contract TokenManager is ITokenController, AragonApp { // ,IForwarder makes cove
     * @param _owner The address that sent the ether to create tokens
     * @return True if the ether is accepted, false for it to throw
     */
-    function proxyPayment(address _owner) payable public returns (bool) {
+    function proxyPayment(address _owner) payable isInitialized public returns (bool) {
```
sohkai: I think we don't need to have `isInitialized` on this, since it's not a `view` simply out of conformance to the minime interface.

apps/token-manager/contracts/TokenManager.sol
```diff
@@ -365,7 +365,7 @@ contract TokenManager is ITokenController, AragonApp { // ,IForwarder makes cove
     * @param _amount The amount in the `approve()` call
     * @return False if the controller does not authorize the approval
     */
-    function onApprove(address _owner, address _spender, uint _amount) public returns (bool) {
+    function onApprove(address _owner, address _spender, uint _amount) isInitialized public returns (bool) {
```
sohkai: Same as above.

apps/vault/contracts/Vault.sol
```diff
@@ -54,7 +54,7 @@ contract Vault is VaultBase {
         connectors[ETH] = ethConnector;
     }
 
-    function () payable public {
+    function () payable isInitialized public {
```
sohkai: Let's ignore the Vault for now, since we'll be changing it pretty soon.

apps/token-manager/contracts/TokenManager.sol
bingen: Ya, I was in doubt about these ones, but I thought it was more aligned with the purpose of the issue: some day we could change the functions to modify state. Actually, most of the changes in this PR are not needed. But I agree with your comment, so removing it.

PR #309, opened 2018-05-14
Fix CI

bingen:
Failed tests were returning 0 (success) due to `ganache-cli.sh` use.


apps/voting/package.json
```diff
@@ -40,7 +40,7 @@
     "babel-register": "^6.26.0",
     "eth-gas-reporter": "^0.1.1",
     "ganache-cli": "^6.1.0",
-    "solidity-coverage": "0.3.5",
+    "solidity-coverage": "0.4.3",
```
sohkai: Oh wow, does this just work now? I remember having to do some funny things in order for this upgrade to work properly (https://github.com/sc-forks/solidity-coverage/issues/182).

We also might as well specify `0.4.15` as the version: https://github.com/sc-forks/solidity-coverage/blob/master/CHANGELOG.md#0415--2018-03-28

PR #312, opened 2018-05-16
Survey script

derekalia:
Updated the Aragon Survey app. 
Added => scripts.js & survey-settings.js


apps/survey/app/src/script.js
```diff
@@ -0,0 +1,154 @@
...
+ *                     *
+ *   Event Handlers    *
+ *                     *
+ ***********************/
+
+async function startSurvey(state, { surveyId }) {
+  const transform = ({ data, ...survey }) => ({
+    ...survey,
+    data: { ...data, executed: true },
+  })
```
sohkai: I think this bit is wrong; `startSurvey()` should probably just be inserting the new survey rather setting `executed: true`.

apps/survey/app/dist/aragon-ui/188027b3d4ea587a.svg
```diff
@@ -0,0 +1 @@
...-6.732 6.733-6.732 2.273 0 3.668.637 5.011 1.773l-1.446 1.722c-1-.86-1.998-1.395-3.65-1.395-2.412 0-4.237 2.101-4.237 4.598v.034c0 2.687 1.774 4.65 4.443 4.65 1.222 0 2.342-.396 3.15-.982V25.56l2.222-1.48v5.475c-1.291 1.103-3.134 2.032-5.441 2.032-4.098 0-6.785-2.91-6.785-6.698M67.82 18.551l-5.734 13.121h2.342l1.343-3.134.844-2.031 2.238-5.218 2.256 5.218.827 2.031 1.326 3.134h2.428l-5.734-13.121zm32.132-.258l-5.751 13.121h2.359l1.343-3.151.827-2.032 2.238-5.2 2.256 5.2.844 2.032 1.326 3.151h2.41l-5.734-13.121zm33.888-.138c-4.011 0-6.835 3.065-6.835 6.733v.034c0 3.668 2.79 6.699 6.801 6.699 4.013 0 6.837-3.065 6.837-6.733v-.035c0-3.667-2.807-6.698-6.802-6.698zm4.392 6.785c0 2.53-1.808 4.597-4.391 4.597s-4.443-2.1-4.443-4.632v-.034c0-2.532 1.808-4.598 4.391-4.598s4.443 2.1 4.443 4.632v.035zm16.152-6.561v9.006l-6.991-9.006h-2.118v13.035h2.255v-9.281l7.198 9.281h1.912V18.38zm-66.847 8.093c1.825-.516 3.117-1.808 3.117-3.943v-.035c0-1.136-.396-2.1-1.085-2.806-.844-.827-2.118-1.31-3.771-1.31h-5.82v13.036h2.29V20.462h3.34c.293 0 .551.035.792.07 1.206.206 1.912.93 1.912 2.083v.034c0 1.326-1.05 2.17-2.687 2.17h-1.274l-.895 2.032h1.687l3.203 4.546h2.704l-3.513-4.925z" fill="#FFF"/></g></svg>
```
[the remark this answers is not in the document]
derekalia: thanks 👍

apps/survey/app/src/script.js
derekalia: Removed. Thanks for catching that.

sohkai: A _slightly_ better way would be to pass the entire object through (avoids a new object if the transpiler / compiler isn't smart and is cleaner in intent); I've been thinking this is probably what the `transform` argument should default to anyway.

It'd look like:

```js
const transform = survey => survey
```

[See an example in the voting app's script](https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/script.js#L58).

future-apps/payroll/.gitignore
```diff
@@ -0,0 +1,7 @@
+.cache
```
sohkai: Hmm, this should probably be in the payroll app's PR :).

apps/survey/app/src/script.js
```diff
@@ -0,0 +1,154 @@
...
+/***********************
+ *                     *
+ *   Event Handlers    *
+ *                     *
+ ***********************/
+
+async function startSurvey(state, { surveyId }) {
+  const transform = ({ data, ...survey }) => ({
+    ...survey,
+    data: { ...data },
```
sohkai: A _slightly_ better way would be to pass the entire object through (avoids a new object if the transpiler / compiler isn't smart and is cleaner in intent); I've been thinking this is probably what the `transform` argument should default to anyway.

It'd look like:

```js
const transform = survey => survey
```

[See an example in the voting app's script](https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/script.js#L58).

apps/survey/app/dist/aragon-ui/188027b3d4ea587a.svg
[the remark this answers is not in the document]
sohkai: Hmm, since it's already been checked in, you might have to forcibly remove them from the repo even though they're now added to the gitignore.

future-apps/payroll/.gitignore
derekalia: Agreed. I noticed that after I pushed 😅

PR #317, opened 2018-05-17
Pin @aragon/os and other contract packages

sohkai:
Contract repos should always be pinned for real contracts.

I've left them unpinned in the templates that are meant to be used only for testing (rather than deployed).


sohkai:
Closing this and will do later when we bump all the apps to 3.1.7+

PR #318, opened 2018-05-17
Add local migrations file in apps for testing

sohkai:
Fixes the tests from being broken due to a missing migrations contract.


sohkai:
Will be done later.

PR #320, opened 2018-05-18
Update licenses to include both GPL and AGPL

sohkai:
Not sure if this really matters, and the way the repo is set up makes this a bit of a pain, but I've tried to make it so that both GPL and AGPL are included.

If you like this, I'll do it for the rest of the apps too.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -659,3 +659,688 @@ specific requirements.
 if any, to sign a "copyright disclaimer" for the program, if necessary.
 For more information on this, and how to apply and follow the GNU AGPL, see
 <http://www.gnu.org/licenses/>.
+
+----------------------------------------------------------------------------
+
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <http://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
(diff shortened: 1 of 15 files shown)


LICENSE
```diff
@@ -659,3 +659,687 @@ specific requirements.
 if any, to sign a "copyright disclaimer" for the program, if necessary.
 For more information on this, and how to apply and follow the GNU AGPL, see
 <http://www.gnu.org/licenses/>.
+
+----------------------------------------------------------------------------
```
sohkai: I've just concated the GPL license after, hopefully that's OK.

LICENSE
```diff
@@ -659,3 +659,687 @@ specific requirements.
...
+  The GNU General Public License does not permit incorporating your program
+into proprietary programs.  If your program is a subroutine library, you
+may consider it more useful to permit linking proprietary applications with
+the library.  If this is what you want to do, use the GNU Lesser General
+Public License instead of this License.  But first, please read
+<http://www.gnu.org/philosophy/why-not-lgpl.html>.
+
+----------------------------------------------------------------------------
+
+Note:
```
sohkai: And this is for individual files to declare that they're GPL-3.0 rather than AGPL-3.0 (basically any deployed contracts).

LICENSE
```diff
@@ -659,3 +659,688 @@ specific requirements.
...
+Unless otherwise specified, individual files are licensed as AGPL-3.0-or-later.
```
sohkai: **NOTE**: I made this line up and have no idea if it's enough.

PR #328, opened 2018-05-22
Change vote function comment to 'yea' from 'yay'

Kyrrui:
Throughout the file, "Yea" is used as the spelling, which accurately aligns with the way it is spelled in US politics https://www.senate.gov/general/Features/votes.htm ,  so it should be 'yea' here instead of 'yay'


sohkai:
Thanks @Kyrrui! This is definitely something we missed :).

PR #335, opened 2018-05-24
Survey: emit current option's power

sohkai:
Makes it much, much easier on the frontend to calculate historical information about how an option was voted in a survey.

The voter's stake can still be found either by looking at the contract (if it hasn't been changed), or by looking at the events between two voting events.


sohkai:
@bingen Good point, for some reason I thought we were hard capped at 4 parameters for an event but that's only for indexed params. I'll add it back.

PR #337, opened 2018-05-24
Survey background script with more data

sohkai:
Also modifies the frontend schema to be consistent with what the background script will give.


bpierre:
✨ 😊 👍

PR #342, opened 2018-05-26
Voting: fix VotePanel not using balanceOfAt()

sohkai:
Oops, forgot to use `balanceOfAt()` earlier 😅


sohkai:
@bpierre Feel free to merge and then build on top of this :).

izqui:
I added information about what block the snapshot was taken and fixed the copy being `You will cast 1 votes`.

PR #349, opened 2018-05-31
 Fix token calculation in survey voting to force integers

sohkai:


sohkai:
Haha I think everyone was just dragging it all the way to the right to get 100%

PR #101, opened 2018-02-23
Staking app

izqui:
Implements a generic Staking primitive that can be used to build things such as the Aragon Network v2 or TCRs.

Inspired by @onbjerg's [aragonlabs' staking app](https://github.com/aragonlabs/staking) and Harbour's [Stakebank](https://github.com/HarbourProject/stakebank)

Conforms to [ERC900](https://github.com/ethereum/EIPs/issues/900) staking interface.

Needs tests, documentation and review.

PR #351, opened 2018-06-01
Staking: Address PR #101 comments

bingen:
And add tests.

TODO: fix issue with `timeTravel` and solidity coverage (it works fine with `npm run test`)


future-apps/staking/contracts/Staking.sol
```diff
@@ -97,6 +104,10 @@ contract Staking is ERCStaking, AragonApp {
     }
   }
 
+  function lockIndefinitely(uint256 amount, address unlocker, bytes32 metadata, bytes data) public returns(uint256 lockId) {
```
izqui: Is there any reason for adding a function for this? See a limited usecase and people can do it themselves anyway

future-apps/staking/contracts/Staking.sol
```diff
@@ -68,25 +74,26 @@ contract Staking is ERCStaking, AragonApp {
...
     require(stakingToken.transferFrom(msg.sender, this, amount));
-    processStake(acct, amount, data);
+
+    // process Stake
+    accounts[acct].amount = accounts[acct].amount.add(amount);
+
+    Staked(acct, amount, totalStakedFor(acct), data);
 
     if (stakeScript.length > 0) {
       runScript(stakeScript, data, new address[](0));
```
izqui: I'm having second thoughts on whether we should add `stakingToken` to the blacklist for all script executions

future-apps/staking/contracts/Staking.sol
```diff
@@ -108,9 +119,13 @@ contract Staking is ERCStaking, AragonApp {
...
+    returns(uint256 lockId)
   {
+    // lock 0 tokens makes no sense
+    require(amount > 0);
+
     Lock memory newLock = Lock(amount, Timespan(lockEnds, TimeUnit(lockUnit)), unlocker, metadata);
-    uint256 lockId = accounts[msg.sender].locks.push(newLock) - 1;
+    lockId = accounts[msg.sender].locks.push(newLock) - 1;
 
     Locked(msg.sender, lockId, metadata);
```
izqui: We should log the amount of locked tokens

future-apps/staking/contracts/Staking.sol
```diff
@@ -131,15 +146,30 @@ contract Staking is ERCStaking, AragonApp {
...
-    lock(amount, lockUnit, lockEnds, unlocker, metadata, lockData);
+    return lock(amount, lockUnit, lockEnds, unlocker, metadata, lockData);
   }
 
-  function unlockAll(address acct) external {
+  function unlockAllOrNone(address acct) external {
     while (accounts[acct].locks.length > 0) {
-      if (canUnlock(acct, 0)) {
-        unlock(acct, 0);
+      uint256 lastAccountLock = accounts[acct].locks.length - 1;
```
izqui: I like the idea of this function, but we shouldn't re-implement unlock here. Why not just iterate over all as in `unlockAll`?

future-apps/staking/contracts/Staking.sol
```diff
@@ -158,12 +188,14 @@ contract Staking is ERCStaking, AragonApp {
...
 
   function moveTokens(address from, address to, uint256 amount) authP(GOD_ROLE, arr(from, to, amount)) external {
-    require(accounts[from].amount >= amount);
+    // move 0 tokens makes no sense
+    require(amount > 0);
 
-    accounts[from].amount -= amount;
-    accounts[to].amount += amount;
+    // make sure we don't move locked tokens, to avoid inconsistency
+    require(unlockedBalanceOf(from) >= amount);
```
izqui: I understand where you are coming from, but I think move tokens should be able to move them regardless of locks.

future-apps/staking/contracts/Staking.sol
```diff
@@ -68,25 +74,26 @@ contract Staking is ERCStaking, AragonApp {
...
   }
 
-  function processStake(address acct, uint256 amount, bytes data) internal {
-    accounts[acct].amount += amount; // overflow check in token contract
-    assert(accounts[acct].amount >= amount);
-
-    Staked(acct, amount, totalStakedFor(acct), data);
-  }
-
   function unstake(uint256 amount, bytes data) authP(UNSTAKE_ROLE, arr(amount)) checkUnlocked(amount) public {
```
izqui: We should add a function to `unlockAndUnstake` that uses the `unlockAll` function

future-apps/staking/package.json
```diff
@@ -3,10 +3,23 @@
...
+  "scripts": {
+    "lint": "solium --dir ./contracts",
+    "test": "TRUFFLE_TEST=true npm run ganache-cli:dev",
+    "test:gas": "GAS_REPORTER=true npm test",
+    "coverage": "./node_modules/@aragon/test-helpers/run-coverage.sh",
+    "console": "node_modules/.bin/truffle console",
+    "ganache-cli:dev": "./node_modules/@aragon/test-helpers/ganache-cli.sh",
+    "ganache-cli:coverage": "SOLIDITY_COVERAGE=true npm run ganache-cli:dev"
+  },
   "author": "Aragon Institution MTU",
```
izqui: Aragon One AG

future-apps/staking/package.json
```diff
@@ -3,10 +3,23 @@
...
   "license": "GPLv3",
```
izqui: GPL-3.0

future-apps/staking/test/staking.js
```diff
@@ -0,0 +1,480 @@
...
+      const r = await app.lock(amount / 2, TIME_UNIT_SECONDS, initialTimestamp + time, other, '', '')
+      const lockId = getEvent(r, 'Locked', 'lockId')
+
+      // check lock values
+      const lock = await app.getLock(owner, lockId)
+      assert.equal(lock[0], amount / 2, "locked amount should match")
+      assert.equal(lock[1], TIME_UNIT_SECONDS, "lock time unit should match")
+      assert.equal(lock[2], initialTimestamp + time, "lock time end should match")
+      assert.equal(lock[3], other, "unlocker should match")
+      assert.equal(lock[4], "0x0000000000000000000000000000000000000000000000000000000000000000", "lock metadata should match")
```
izqui: Remove magic number to a const

future-apps/staking/test/staking.js
```diff
@@ -0,0 +1,480 @@
...
+    it('locks using blocks', async () => {
+      const mockCounter = await getContract('Counter').new()
+      const amount = 100
+      const blocks = 10
+      await token.approve(app.address, amount)
+      await app.stake(amount, '')
+      const initialBlockNumber = await getBlockNumber()
+      const r = await app.lock(amount / 2, TIME_UNIT_BLOCKS, initialBlockNumber + blocks, other, '', '')
+      const lockId = getEvent(r, 'Locked', 'lockId')
+      await timeTravel(2000)
```
izqui: Why is the timetravel required checking for block locking logic?

future-apps/staking/test/staking.js
```diff
@@ -0,0 +1,480 @@
...
+
+      // can not unlock
+      assert.isFalse(await app.canUnlock(owner, lockId))
+      assert.equal((await app.unlockedBalanceOf(owner)).valueOf(), amount / 2, "Unlocked balance should match")
+      assert.equal((await app.locksCount(owner)).valueOf(), parseInt(lockId, 10) + 1, "last lock id should match")
+
+      // we just send dumb transactions to increment block number
+      let blockNumber
+      do {
+        await mockCounter.increment()
```
izqui: Nice hack 😂🙏 Though this is going to slow down the test unnecessarily. Two possible solutions:

- Make the lock shorter, 2 or 3 blocks should be ok.
- Mock block numbers like we do with timestamps (I thought we did this all over the system, but apparently just in Finance https://github.com/aragon/aragon-apps/blob/master/apps/finance/test/mocks/FinanceMock.sol)

Also rather than introducing a random contract, you can perform ETH value transfers to the same account `web3.eth.sendTransaction({ from: owner, to: owner: value: 1, gasPrice: 1 })` to bump blocks.

future-apps/staking/test/staking.js
```diff
@@ -0,0 +1,480 @@
...
+        blockNumber = await getBlockNumber()
+      } while( blockNumber <= initialBlockNumber + blocks)
```
izqui: Also I hadn't seen this in a while 😉. We could just do a `for (let x of new Array(blocks))` without the block number check, as ganache always creates a new block per tx submitted.

If we want these tests to properly run in geth or parity, we should await for the block to change before continuing sending txs

future-apps/staking/test/staking.js
```diff
@@ -0,0 +1,480 @@
...
+
+      await timeTravel(time + 1)
+      // unlock
+      await app.unlockAll(owner)
+
+      assert.equal((await app.unlockedBalanceOf(owner)).valueOf(), amount / 2, "Unlocked balance should match")
+      assert.equal((await app.locksCount(owner)).valueOf(), 1, "there shouldn't be locks")
+    })
+
+    it('fails trying to unlockAllOrNone', async () => {
```
izqui: 'fails to unlockAllOrNone if a lock cannot be unlocked'

future-apps/staking/test/staking.js
```diff
@@ -0,0 +1,480 @@
...
+    })
+
+    it('fails moving more tokens than staked', async () => {
+      const amount = 100
+      await token.approve(app.address, amount)
+      await app.stake(amount, '')
+      return assertRevert(async () => {
+        await app.moveTokens(owner, other, amount + 1)
+      })
+    })
```
izqui: Add a test for failing to move more tokens from an account than unlocked.

future-apps/staking/test/staking.js
```diff
@@ -0,0 +1,480 @@
...
+      await app.stake(amount, '')
+
+      const finalOwnerBalance = parseInt((await token.balanceOf(owner)).valueOf(), 10)
+      const finalStakingBalance = parseInt((await token.balanceOf(app.address)).valueOf(), 10)
+      assert.equal(finalOwnerBalance, initialOwnerBalance - amount, "owner balance should match")
+      assert.equal(finalStakingBalance, initialStakingBalance + amount, "Staking app balance should match")
+      assert.equal((await app.totalStakedFor(owner)).valueOf(), amount, "staked value should match")
+      assert.equal(await executionTarget.stakeCounter(), 1, 'should have received execution call')
+    })
+
```
izqui: Add test for stakeAndLock triggering both scripts

future-apps/staking/contracts/Staking.sol
bingen: Well, I thought it may be convenient for the user to avoid the need of `uint8(TimeUnit.Seconds), MAX_UINT64`, but I can remove it if you don't think so.

future-apps/staking/contracts/Staking.sol
bingen: But the (correct me if I'm wrong) the following situation may happen: an account could have more locked tokens than staked. So that account could try to stake more tokens for whatever reason but still wouldn't have available staked tokens. Even more, `unlockedBalanceOf` could revert trying to do the subtraction. I think that leaving such an inconsistent state where locked tokens amount can be greater than staked amount can give a lot of problems and be very confusing for the user. Let me put it this way: how can you have locked tokens that you don't actually have?
Another option would be to allow `moveTokens` to unlock first, but then it should require `UNLOCK_ROLE`. Maybe we can do an `unlockAndMove` function.

future-apps/staking/contracts/Staking.sol
bingen: Well, yes, I'm "re-implementing", but it's only 2 things: remove last array element and emitting event.
I don't want to call `unlock` because then, in case of success, I would be repeating the `canUnlock` and `lastAccountLock` things. Yes, not big deal, but I find it ugly.
With respect `unlockAll`, I'm iterating over all of them, but with a `while` instead of a `for`, which I find nicer. In `unlockAll` I can not do the same because in the end the array may not be empty.

future-apps/staking/contracts/Staking.sol
bingen: Done.

future-apps/staking/test/staking.js
bingen: Done.

future-apps/staking/test/staking.js
bingen: Done.

PR #356, opened 2018-06-05
Update test sync deploy.js

EvertonMelo:
sync after callback


EvertonMelo:
@izqui 
Sorry, teste only rss.. not Mainnet. rss

PR #360, opened 2018-06-06
Survey: Add Jest and some tests for math-utils.js

bpierre:


apps/survey/app/src/math-utils.js
```diff
@@ -49,21 +49,31 @@ function highestValueIndex(values) {
     .sort((v1, v2) => v2.value.minus(v1.value))[0].index
 }
 
-export function scaleBigNumberValuesSet(values, total = new BigNumber(100)) {
+// Scale a set of 0 => 1 values to equal `total`.
```
sohkai: Maybe `[0, 1] values` would work better as math notation? Or even a set of values summing to 1?

apps/survey/app/src/math-utils.test.js
```diff
@@ -0,0 +1,55 @@
...
+    ],
+    [
+      'should correct the values if their sum is less than 1',
+      [[0.1, 0.1], '1000'],
+      ['900', '100'],
+    ],
+    [
+      'should correct the values if their sum is more than 1',
+      [[1.1, 0.3], '1000'],
+      ['700', '300'],
```
sohkai: These two behaviours seem like they'd be surprising—given `[0.1, 0.1]`, I would expect `['100', '100']`, and given `[1.1, 0.3]` I would expect an error similar to the `'too far from 1'` error.

bpierre: This is to correct any number precision issue when summing the values to get `1`, but I agree it is not ideal.

Do you think it would be a good idea to check if `Math.abs(1 - sum) < 0.0001`?

- A value within that range would be accepted as `1`.
- A value below would give a result below 100% (`[0.1, 0.1]` => `['100', '100']`).
- A value above would trigger a `too far from 1` error.

sohkai: @bpierre I think those rules make a lot of sense, especially if we let the user configure the precision they need.

PR #383, opened 2018-06-19
Payroll review

bingen:
Addressing @sohkai review:

Big:
​
- [X] _payTokens() is at risk of re-entrancy: if an employee sets their address to have a fallback that re-enters the payroll app, both of payday() and removeEmployee() will give that employee more funds than they should
- [X] Should determineAllocation() clear the employee's allocation mapping before setting it (otherwise old allocations could still exist)?
- [X] Setting allocation: time between could be parameterized in the initialize (https://github.com/aragon/aragon-apps/blob/master/future-apps/payroll/contracts/Payroll.sol#L253) (But I think this isn't actually true... there's no speed bump for setting allocations afaik?)
​
Small:
​
- [X] Events: some of the parameters could be indexed
- [X] ADD_EMPLOYEE_ROLE: should this also pass the address (useful if you wanted to whitelist / blacklist certain addresses, e.g. those who hold a token?)
- [x] Maybe another role should be added for setEmployeeSalary?
- [X] Setting salary: seems weird that the arguments are in yearly time, when the storage is in second time. Could push that responsibility to a frontend instead?
- [X] Could make require(employeeIds[employees[employeeId].accountAddress] != 0) an employeeExists modifier
- [ ] escapehatch might not be necessary anymore; token escape hatch could also be comboed into one function
- [X] https://github.com/aragon/aragon-apps/blob/master/future-apps/payroll/contracts/Payroll.sol#L255 should probably say integers summing to 100?
- [X] We could make a small uint8 safemath implementation and use it instead of checking asserts (could even push it to zeppelin)
- [X] https://github.com/aragon/aragon-apps/blob/master/future-apps/payroll/contracts/Payroll.sol#L316 could use a delete instead
- [x] getEmployee(): could have an overload for getting the employee by address
- [X] getAllocation() is missing a return natspec description: https://github.com/aragon/aragon-apps/blob/master/future-apps/payroll/contracts/Payroll.sol#L350
- [X] https://github.com/aragon/aragon-apps/blob/master/future-apps/payroll/contracts/Payroll.sol#L367 has a small typo ("token")
- [x] https://github.com/aragon/aragon-apps/blob/master/future-apps/payroll/contracts/Payroll.sol#L427 is technically unnecessary (it's defaulted to false)
- [ ] If we're going to use uint32 to iterate over allowedTokensArray (https://github.com/aragon/aragon-apps/blob/master/future-apps/payroll/contracts/Payroll.sol#L428), we should probably also check that the length does not overflow when we push to it (and add a comment or assertion in the iteration); given gas costs, I'd suggest even lowering that allowed space to uint16 or something.
- [x] Sprinkle some radspec
- [X] Add isInitialized to state modifying functions

### DIFF
--- a/future-apps/payroll/arapp.json
+++ b/future-apps/payroll/arapp.json
@@ -16,7 +16,7 @@
         }
     ],
     "roles": [
-        { "name": "Add employees", "id": "ADD_EMPLOYEE_ROLE", "params": ["Initial yearly payroll", "Start date"] },
+        { "name": "Add employees", "id": "ADD_EMPLOYEE_ROLE", "params": ["Account address", "Initial yearly payroll", "Start date"] },
         { "name": "Remove employees", "id": "REMOVE_EMPLOYEE_ROLE" },
         { "name": "Allowed Tokens Manager", "id": "ALLOWED_TOKENS_MANAGER_ROLE", "params": "Token" },
         { "name": "Change Price Feed", "id": "CHANGE_PRICE_FEED_ROLE", "params": ["Old Price Feed", "New Price Feed"] },
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -50,15 +50,15 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
     address[] private allowedTokensArray;
 
     event AddEmployee(
-        uint128 employeeId,
-        address accountAddress,
+        uint128 indexed employeeId,
+        address indexed accountAddress,
         uint256 initialYearlyDenominationSalary,
         string name,
         uint256 startDate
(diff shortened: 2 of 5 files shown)


izqui:
Adding to brett's review, it would be great to have events more things:

- `addAllowedToken`
- `determineAllocation`
- `setEmployeeSalary`
- `changeAddressByEmployee`
- `removeEmployee
- One event every time a token is sent to an employee

Other minor things:
- We can now remove the `Fund` event as it is not being used and remove `depositToFinance` and `scapeHacth`
- In `determineAllocation`, we can remove this comment now too `Only callable once every 6 months`
- If an employees haven't set an allocation, it is impossible to remove them. This creates a weird race condition.

sohkai:
Superseded by #479.

PR #386, opened 2018-06-25
Add build command to main app repos

izqui:
This allows `apm publish` to build the frontend of the apps so just one command is required.

In case of the vault app that doesn't have a frontend, we just create the directory to avoid the issue of trying to publish an inexistent directory.

### DIFF
--- a/apps/finance/package.json
+++ b/apps/finance/package.json
@@ -5,6 +5,7 @@
   "scripts": {
     "postinstall": "cd app && npm install",
     "compile": "truffle compile",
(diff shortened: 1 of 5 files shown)


sohkai:
> In case of the vault app that doesn't have a frontend, we just create the directory to avoid the issue of trying to publish an inexistent directory.

Does `apm publish` always expect to publish out of `app/build`? Seems like a weird hack to get around the tooling. Maybe `apm publish` should detect if an app has a frontend or not via the `manifest.json`'s `start_url`?

Issue #390, opened 2018-07-01
Payroll: bug changing employee salary


bingen:
When changing an employee salary, next withdraw would use the new amount also for the remaining period before the change.

izqui:
I think both this issue and #389 can be solved in a very similar way. We should add a new variable to the employee struct that can keep track of the amount of tokens (denominated in the denomination token) that is owed to the employee, `accruedValue` as suggested by @sohkai. There could also be a role to automatically add more value to this counter for an employee, which would be useful for reimbursements. 

If we do this, we need to stop removing employees from storage when they are terminated and just have a way to flag when was the moment when they stopped accruing value.

sohkai:
Merged with #479.

PR #394, opened 2018-07-09
Payroll accrued value

### LINKED ISSUE #389: Payroll: Add termination date for employees
If an employee didn't have an allocation set, trying to remove him would fail. The proposed solution is to add a termination date field to Employee object and allow the employee to withdraw up to that date in cas of termination.

bingen:
Closes #389 and #390


future-apps/payroll/test/payroll_accrued_value.js
```diff
@@ -0,0 +1,110 @@
...
+  })
+
+  it('Fails trying to re-enable employee', async () => {
+    const timestamp = parseInt(await payroll.getTimestampPublic.call(), 10)
+    await payroll.terminateEmployeeNow(employeeId1)
+    await payroll.mockSetTimestamp(timestamp + 500);
+    return assertRevert(async () => {
+      await payroll.terminateEmployee(employeeId1, timestamp + SECONDS_IN_A_YEAR)
+    })
+    assert.isTrue(0)
```
izqui: Did you forget to remove this?

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -212,34 +217,79 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
         authP(ADD_EMPLOYEE_ROLE, arr(employees[employeeId].accountAddress, denominationSalary, 0))
     {
+        // forbid to change salary in the past
+        require(changeDate >= getTimestamp());
+
+        // Add owed salary to employee's accrued value
+        uint256 owed = _getOwedSalary(employeeId, changeDate);
+
+        employees[employeeId].lastPayroll = changeDate;
+        _addAccruedValue(employeeId, owed);
```
izqui: This will make all the salary that an employee would be getting during the period between when the function is called and the `changeDate` immediately available, with no option to be taken back if the employee is terminated. A joke such as 'we will raise your salary to 1 USD in 2100' could completely deplete the DAO funds as the employee could get 80 years worth of their previous salary immediately.

We can either remove the possibility to set a custom `changeDate` (and always change at this moment) or add a lot of extra logic for handling different salary periods. I like the former.

future-apps/payroll/test/payroll_accrued_value.js
bingen: Absolutely, sorry.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -511,6 +586,31 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
             SendPayroll(employee.accountAddress, token, tokenAmount);
             somethingPaid = true;
         }
+
+        // try to remove employee
+        if (employees[employeeId].endDate <= getTimestamp() && employees[employeeId].accruedValue == 0) {
```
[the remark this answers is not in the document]
bingen: Ok, changing it. That was actually the reason. What's the advantage of not deleting it?

future-apps/payroll/contracts/Payroll.sol
bingen: Good catch! This is so wrong!!

bingen: It's certainly an edge case, and it wouldn't happen with a proper/regular management, but it's really dangerous and misleading. 
I liked the idea of being able to set the change date in advance and with exactitude, but I guess it can be done right after the date has passed and then adjust the exact amount with `addAccruedValue` function if needed.

future-apps/payroll/contracts/Payroll.sol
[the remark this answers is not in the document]
bingen: I guess at least we need to remove the account address, to be able to reuse it, so I would remove it all unless there is some reason to keep it.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -461,6 +523,12 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
         nextEmployee++;
     }
 
+    function _addAccruedValue(uint128 employeeId, uint256 amount) internal {
+        employees[employeeId].accruedValue = employees[employeeId].accruedValue.add(amount);
```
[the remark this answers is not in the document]
bingen: Hm, interesting. You are right, but I guess that even setting the `accruedValue` to such an amount (`2^128`) would still block the employee in most cases as there wouldn't be enough funds to execute `payday`. I guess the best solution would be to allow to withdraw as much as possible up to owed amount, but that would make logic even way more complex, taking into account the multi-currency support.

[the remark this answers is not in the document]
bingen: I was thinking about limiting it up to the yearly salary. I think this function was meant to be used for things like expenses, bonuses, etc. They are usually less than one year of salary, but even if it's more but not much more, it could be done using several transactions. So it would cover most of the use case being quite safe... The problem is that I realized the same problem happens with the salary itself.

future-apps/payroll/contracts/Payroll.sol
bingen: Another option could be to store last salary change date and not to allow another change if employee hasn't called `payday` after the last one. I guess in regular use cases payday should happen way more often than salary changes. But maybe it's still too much complexity.

sohkai:
Superseded by #479.

PR #395, opened 2018-07-12
Explicitly define state visibility

sohkai:
Explicitly set the state to be `internal`, since these either have custom getters or have indices start at `1`.

### DIFF
--- a/apps/token-manager/contracts/TokenManager.sol
+++ b/apps/token-manager/contracts/TokenManager.sol
@@ -38,8 +38,8 @@ contract TokenManager is ITokenController, AragonApp, IForwarder {
         bool revokable;
     }
 
-    mapping (address => TokenVesting[]) vestings;
-    mapping (address => bool) everHeld;
+    mapping (address => TokenVesting[]) internal vestings;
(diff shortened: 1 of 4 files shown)


apps/token-manager/contracts/TokenManager.sol
```diff
@@ -38,8 +38,8 @@ contract TokenManager is ITokenController, AragonApp, IForwarder {
         bool revokable;
     }
 
-    mapping (address => TokenVesting[]) vestings;
-    mapping (address => bool) everHeld;
+    mapping (address => TokenVesting[]) internal vestings;
+    mapping (address => bool) internal everHeld;
```
bingen: I guess it's not a big deal (there is `allHolders`) but I don't see a getter for this one.

sohkai: Good point, this can be public :).

sohkai:
> Why does it matter if indices start at 1?

It makes the automatic getter more awkward, but usually these are returning structs anyway and are better served with a custom getter.

PR #402, opened 2018-07-16
Finance: remove limit on transitions per transaction in transitionsPeriod modifier

sohkai:
From https://github.com/ConsenSys/aragon-apps-audit-repo/issues/29:

> Remove the MAX_PERIOD_TRANSITION_PER_TX. Items 1 and 2 above together should make this arbitrary cutoff unnecessary.

--------------

If there are too many periods to transition, an OOG error will revert all changed state, including in-flight transfers of ETH or tokens.

However, this does mean the frontend may have a harder time estimating gas costs.

### DIFF
--- a/apps/finance/test/finance.js
+++ b/apps/finance/test/finance.js
@@ -359,10 +359,11 @@ contract('Finance App', accounts => {
         })
 
         context('many accounting period transitions', () => {
-            let maxTransitions = 0
+            // Arbitrary number of max transitions to simulate OOG behaviour with transitionsPeriod
+            const maxTransitions = 20
 
             beforeEach(async () => {
-                maxTransitions = await app.MAX_PERIOD_TRANSITIONS_PER_TX().then(bn => bn.toNumber())
+                await app.mock_setMaxPeriodTransitions(maxTransitions)
                 await app.mock_setTimestamp(time + (maxTransitions + 2) * periodDuration)
             })
 
--- a/apps/finance/test/mocks/FinanceMock.sol
+++ b/apps/finance/test/mocks/FinanceMock.sol
@@ -3,8 +3,12 @@ pragma solidity 0.4.18;
 import "../../contracts/Finance.sol";
 
 contract FinanceMock is Finance {
-    uint _mockTime = now;
+    uint256 mockTime = now;
+    uint mockMaxPeriodTransitions = MAX_UINT;
 
-    function getTimestamp() internal view returns (uint256) { return _mockTime; }
-    function mock_setTimestamp(uint i) public { _mockTime = i; }
+    function mock_setMaxPeriodTransitions(uint256 i) public { mockMaxPeriodTransitions = i; }
+    function mock_setTimestamp(uint256 i) public { mockTime = i; }
+
+    function getMaxPeriodTransitions() internal view returns (uint256) { return mockMaxPeriodTransitions; }
+    function getTimestamp() internal view returns (uint256) { return mockTime; }
 }
--- a/apps/finance/test/mocks/FinanceMock.sol
+++ b/apps/finance/test/mocks/FinanceMock.sol
@@ -4,7 +4,7 @@ import "../../contracts/Finance.sol";
 
 contract FinanceMock is Finance {
     uint256 mockTime = now;
-    uint mockMaxPeriodTransitions = MAX_UINT;
+    uint256 mockMaxPeriodTransitions = MAX_UINT;
 
     function mock_setMaxPeriodTransitions(uint256 i) public { mockMaxPeriodTransitions = i; }
     function mock_setTimestamp(uint256 i) public { mockTime = i; }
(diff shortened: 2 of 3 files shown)


apps/finance/test/mocks/FinanceMock.sol
```diff
@@ -3,8 +3,12 @@ pragma solidity 0.4.18;
 import "../../contracts/Finance.sol";
 
 contract FinanceMock is Finance {
-    uint _mockTime = now;
+    uint256 mockTime = now;
+    uint mockMaxPeriodTransitions = MAX_UINT;
```
bingen: (Nitpicking, or missing something) Why are you changing `mockTime` to `uint256` but using `uint` for `mockMaxPeriodTransitions`?

apps/finance/test/finance.js
```diff
@@ -359,10 +359,11 @@ contract('Finance App', accounts => {
         })
 
         context('many accounting period transitions', () => {
-            let maxTransitions = 0
+            // Arbitrary number of max transitions to simulate OOG behaviour with transitionsPeriod
+            const maxTransitions = 20
 
             beforeEach(async () => {
-                maxTransitions = await app.MAX_PERIOD_TRANSITIONS_PER_TX().then(bn => bn.toNumber())
+                await app.mock_setMaxPeriodTransitions(maxTransitions)
```
bingen: Do we need this mock function? Can we just set a big enough time forward and let it actually run out of gas?

sohkai:
@bingen 

> I guess the advantage is that now it will only fail when it has to, while before we may be forcing it to fail unnecessarily. Is that it?

Yes, that's the intent. I think we chose 20 pretty arbitrarily, but it should be possible to advance more than that given the cost of another iteration is now pretty cheap.

Some "scientific" gas usage tests in remix (with `0.4.18+commit.9cf6e910.Emscripten.clang`):

```solidity
pragma solidity 0.4.18;

library SafeMath {
  function sub(uint256 a, uint256 b) internal pure returns (uint256) {
    require(b <= a);
    return a - b;
  }
}

library SafeMath64 {
  function add(uint64 a, uint64 b) internal pure returns (uint64) {
    uint64 c = a + b;
    require(c >= a);
    return c;
  }
}

contract Finance {
    using SafeMath for uint256;
    using SafeMath64 for uint64;

    struct TokenStatement {
        uint256 expenses;
        uint256 income;
    }

    struct Period {
        uint64 startTime;
        uint64 endTime;
        uint256 firstTransactionId;
        uint256 lastTransactionId;

        mapping (address => TokenStatement) tokenStatement;
    }
    
    Period[] internal periods; // first index is 0
    uint64 constant public MAX_UINT64 = uint64(-1);
    
    event NewPeriod(uint256 indexed periodId, uint64 periodStarts, uint64 periodEnds);
    
    function Finance() public {
        _newPeriod(uint64(getTimestamp()));
    }

    function transitionAccountingPeriod(uint256 _transitions) public returns (bool success) {
        Period storage currentPeriod = periods[currentPeriodId()];
        uint256 timestamp = getTimestamp();

        // Transition periods if necessary
        while (timestamp > currentPeriod.endTime) {
            if (_transitions == 0) {
                // Required number of transitions is over allowed number, return false indicating
                // it didn't fully transition
                return false;
            }
            _transitions = _transitions.sub(1);

            // If there were any transactions in period, record which was the last
            // In case 0 transactions occured, first and last tx id will be 0
            if (currentPeriod.firstTransactionId != 0) {
                currentPeriod.lastTransactionId = 1; // fudge it
            }

            // new period starts at end time + 1
            currentPeriod = _newPeriod(currentPeriod.endTime.add(1));
        }

        return true;
    }
    
    function currentPeriodId() public view returns (uint256) {
        return periods.length - 1;
    }
    
    function _newPeriod(uint64 _startTime) internal returns (Period storage) {
        uint256 newPeriodId = periods.length++;

        Period storage period = periods[newPeriodId];
        period.startTime = _startTime;

        // fudge this
        uint64 endTime = _startTime + 2 - 1;
        if (endTime < _startTime) { // overflowed
            endTime = MAX_UINT64;
        }
        period.endTime = endTime;

        NewPeriod(newPeriodId, period.startTime, period.endTime);

        return period;
    }
    
    function getTimestamp() internal view returns (uint256) { return now; }
}
```

One transition costs 57771 gas, two cost 92863, ten cost 193545, and so it's around ~30-35k gas per transition. This effectively means someone could transition ~250 times in a single block, if they had to (hopefully not... with these gas prices).

apps/finance/test/finance.js
sohkai: Wanted to do that, but was afraid of the fragility it would introduce into the tests (we'd have to guess how many transitions would force a OOG, and if it ever changed in the future, our tests could break).

apps/finance/test/mocks/FinanceMock.sol
sohkai: Ohh, good point!

Issue #375, opened 2018-06-13
Finance: Payment struct has confusing name


sohkai:
`Payment`s are only created for recurring payments. Suggested to name to something more clear, e.g. `RecurringPayment`.

IvanTheGreatDev:
I think `RecurringPayments` is the best name since they occur at a set interval. 

Does this also mean `mapping (uint256 => Payment) internal payments;` is changed to `recurringPayments`?

I think leaving the rest of the functions as payment is fine, I'm just not sure about the above.

sohkai:
Given that this didn't make it to 0.6, any external interfaces will likely be difficult to upgrade afterwards (the event names, at least, will be difficult to upgrade).

Just renaming the struct and the mapping would be a good start :). We will likely handle the rest of the nomenclature in the frontend when it grows to have more budgeting / etc features.

PR #410, opened 2018-07-30
Use new uint64 helpers for time

bingen:
Depends on aragonOS pull #375.

TODO:

- [ ] Upgrade aragonOS package version, once that PR is merged and new version is published.


bingen:
Incorporated into #425

PR #412, opened 2018-07-31
Survey: fix calculation when votes are reset

sohkai:
Discovered while debugging Memefactory's survey DAO. I didn't factor in resetting votes earlier when doing calculations for the history chart.

Also adds `logIndex` as a way to make sure we don't work on the same event.

**Breaking change**: all past survey DAOs use `0` to signify sparseness in their history whereas `-1` is now used (so we can properly account for when a vote is reset back to 0). Although the data model has changed, almost no users on mainnet should be impacted by this change, at least visually.

Before:

After:

### DIFF
--- a/apps/survey/app/src/script.js
+++ b/apps/survey/app/src/script.js
@@ -89,7 +89,10 @@ async function initialize(tokenAddr) {
 async function createStore(token, tokenSettings) {
   const { decimals: tokenDecimals, symbol: tokenSymbol } = tokenSettings
   return app.store(
-    async (state, { blockNumber, event, returnValues, transactionIndex }) => {
+    async (
+      state,
+      { blockNumber, event, logIndex, returnValues, transactionIndex }
+    ) => {
       let nextState = {
         ...state,
         // Fetch the app's settings, if we haven't already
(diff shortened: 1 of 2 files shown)


apps/survey/app/src/script.js
```diff
@@ -108,11 +111,30 @@ async function createStore(token, tokenSettings) {
...
+              transactionIndex,
+              logIndex
+            )
+            break
+          case 'ResetVote':
+            // We have to manually calculate the option power left when a vote is reset
+            returnValues = {
+              ...returnValues,
+              // TODO: use BN.js instead
+              optionPower: String(
```
sohkai: This is not really nice, but I didn't want to deal with the hassle of adding BN to this script (it should do all math via BN) until we completely refactor `bignumber` to `BN.js` in this app.

2color: [bignumber.js](https://github.com/MikeMcl/bignumber.js/) allows doing arithmetic and also recommends instantiating with a string for numbers of limited precision. Which problem do you aim so solve by switching to `BN.js`?

sohkai: `BN.js` is faster and more widely used, compared to `bignumber.js`. This app should eventually be converted to completely use `BN.js` for all math operations in the future.

PR #409, opened 2018-07-20
Staking 4 deferred lock

bingen:
Allow to create locks for the future, for simultaneous voting in PLCR. 

Also fixes #406

PR #415, opened 2018-08-06
[WIP] Staking checkpointing

izqui:
Implements a Staking sub-app that checkpoints a stake holder's balance on every stake/unstake action, conforming to the optional history methods of [ERC900](https://eips.ethereum.org/EIPS/eip-900).

This is a required primitive for implementing stake-based voting with any ERC20 (w/o Minime) in a governance model that doesn't require locking tokens. An interesting combination would be a Voting app that requires a stake-lock in order to perform certain actions (create a proposal or vote favorably).

I'm currently thinking about whether we should move locking functionality off the main Staking contract in a similar fashion, as it would make it easier for us to implement different types of locking without adding exponential complexity, like it is currently happening with `overlocking` in #409 (cc @bingen)

To do:
- [x] Test `StakingHistory` with existing `staking.js` tests (✅ all passing)
- [x] Test history-specific functionality
- [ ] (?) MiniMe interface wrapping for automatic compatibility with current MiniMe-compatible voting apps.

### DIFF
--- a/future-apps/staking/contracts/Checkpointing.sol
+++ b/future-apps/staking/contracts/Checkpointing.sol
@@ -10,7 +10,9 @@ library Checkpointing {
     Checkpoint[] history;
   }
 
-  function add(History storage self, uint128 value, uint128 time) internal {
+  uint256 constant MAX_UINT128 = uint256(uint128(-1));
+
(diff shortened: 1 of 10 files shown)


bingen:
Closed in favor of https://github.com/aragon/staking/

PR #422, opened 2018-08-28
Migrate to aragonOS 4

### LINKED ISSUE #315: Update to latest @aragon/os
I have been debugging for a couple hours today and the issue is that while we are bumping aragonOS versions because the templates are still in the old version of aragonOS some stuff was breaking (discovered because of a mismatch filtering for a topic)

### LINKED ISSUE #372: Fix CI

### LINKED ISSUE #385: Vault should revert on deposit if Kernel reference hasn't been set
When https://github.com/aragon/aragonOS/issues/325 is implemented, depositing to the Vault should revert if it is not running behind a proxy (detectable when there is no kernel reference), as assets sent there won't be recoverable as the `auth` modifier will revert.

We can prevent ETH from being deposited in regular ways (unless ETH is sent to the address before the Vault is deployed, or sent with a `selfdestruct`) and ERC20s when deposited using the `deposit` function (cannot avoid people sending tokens directly to the vault address).

### LINKED ISSUE #418: Try to get rid of compiling warnings
Main rationale:
https://github.com/ethereum/solidity/issues/4315#issuecomment-398466651
Besides, it would make easier to spot compiling errors while developing, that sometimes get buried among long list of warnings.

sohkai:
Fixes #418, fixes #315, fixes #372, fixes #385.

A small migration guide for the contracts (I'll flesh this out in https://github.com/aragon/aragonOS/issues/404) ([see changes](https://github.com/aragon/aragon-apps/commit/3e891e73ea23d85735c00b3707980799112db2bc)):

- Upgrade to `pragma solidity 0.4.24`
- Upgrade to `truffle@4.1.14` and `solidity-coverage@0.5.8`; doing so requires that any test contracts be placed in `contracts/`
- All apps are required to be initialized before they can be used
- Unnecessary to use `isInitialized` modifier if the `auth` or `authP` modifier is used
- Changed import locations of `@aragon/os/contracts/lib/zeppelin/{token,math}` to be `@aragon/os/contracts/lib/{token,math}`
- Removed token implementations (`StandardToken` and `MinimeToken`; [`MinimeToken` was added directly as a shared module in this repo])(https://github.com/aragon/aragon-apps/commit/2db53ea682008902000ac2528eb35fb7b18a36ef))
- Apps inheriting from `AragonApp` are automatically petrified upon deployment, so any tests directly using the base contract will fail (e.g. Finance and Survey)

I've opted to spread out the testing for each app as a separate job in travis so we can see which apps fail, rather than lerna failing on the first app.

------------

Todo:

- [x] Update and merge #292
- [x] Fix tests for Finance and Survey by using proxied versions (see Token Manager, Voting)
- [x] Fix compile warnings
- [x] Upgrade and fix linter warnings (will be getting `solium-config-aragon` soon on npm...)
- [x] Update and merge #256, #410 
- [x] Add new tests for app lifecycle (e.g. checking that the base is immediately petrified and unusable)
- [ ] Fix visibility order of constants
- [ ] Upgrade to `aragonOS@4.0.0-beta.1`
- [ ] Make sure #385 is covered


shared/minime/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "@aragon-apps/minime",
```
sohkai: Note that I've opted to use the prefix `@aragon-apps` for private packages that should only be shared inside this repo.

shared/minime/contracts/MinimeToken.sol
```diff
@@ -0,0 +1,575 @@
+pragma solidity ^0.4.24;
```
sohkai: Curious... [there's a minime v2 from last year](https://github.com/Giveth/minime/blob/master/contracts/MiniMeToken.sol); should we update to that?

apps/voting/contracts/test/mocks/VotingMock.sol
```diff
@@ -0,0 +1,20 @@
...
+
+import "../../Voting.sol";
+
+
+contract VotingMock is Voting {
+    /** Ugly hack to work around this issue:
+     * https://github.com/trufflesuite/truffle/issues/569
+     * https://github.com/trufflesuite/truffle/issues/737
+     */
+    function newVoteWithCast(bytes _executionScript, string _metadata, bool _castVote) auth(CREATE_VOTES_ROLE) external returns (uint256 voteId) {
```
sohkai: For some odd reason, the newer version of truffle only exposes `newVote(bytes, string)` rather than the old version which only exposed `newVote(bytes, string, bool)`.

apps/finance/contracts/Finance.sol
```diff
@@ -2,24 +2,22 @@
...
+import "@aragon/os/contracts/lib/math/SafeMath.sol";
+import "@aragon/os/contracts/lib/math/SafeMath64.sol";
 
 import "@aragon/apps-vault/contracts/IVaultConnector.sol";
 
-import "@aragon/os/contracts/lib/misc/Migrations.sol";
 
-
-contract Finance is AragonApp {
+contract Finance is AragonApp, EtherTokenConstant {
```
bingen: `AragonApp` is always `EtherTokenConstant`, so I'll remove it, ok?

apps/survey/contracts/test/mocks/BadToken.sol
```diff
@@ -20,6 +20,6 @@ contract BadToken is MiniMeToken {
     }
 
     function totalSupplyAt(uint _blockNumber) public constant returns(uint) {
-        return 1;
+        return _blockNumber - _blockNumber + 1;
```
izqui: This is way too hacky ;) You can disable the warning by making `_blockNumber` an unnamed parameter:

```
function totalSupplyAt(uint) public constant returns (uint)
```

apps/finance/package.json
```diff
@@ -34,16 +34,18 @@
...
     "solium": "^1.0.4",
-    "truffle": "4.0.5",
+    "truffle": "4.1.14",
     "truffle-hdwallet-provider": "0.0.3",
     "webpack": "3.10.0"
   },
   "dependencies": {
     "@aragon/apps-vault": "^2.0.1",
-    "@aragon/os": "^3.1.10"
+    "@aragon/os": "4.0.0-alpha.2"
```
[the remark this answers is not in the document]
sohkai: See #317. I'd prefer to pin all contract dependencies, having the experience of what happened with the vault 😉. In the future, we should probably use `lerna` directly to upgrade this dependency so all of them are updated at the same time.

apps/token-manager/contracts/TokenManager.sol
```diff
@@ -2,21 +2,19 @@
...
+import "@aragon/os/contracts/lib/token/ERC20.sol";
+import "@aragon/os/contracts/lib/math/SafeMath.sol";
 
-import "@aragon/os/contracts/lib/misc/Migrations.sol";
+import "@aragon-apps/minime/contracts/ITokenController.sol";
+import "@aragon-apps/minime/contracts/MiniMeToken.sol";
 
 
-contract TokenManager is ITokenController, AragonApp, IForwarder {
+contract TokenManager is ITokenController, IForwarder, AragonApp {
```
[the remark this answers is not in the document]
sohkai: Generally I've found it better practice to lead with the abstract classes, to explicitly declare what interfaces the class implements (or should).

shared/minime/truffle.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
```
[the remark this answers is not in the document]
sohkai: Thought about that... but given that this is never meant to be deployed and is only used internally, I thought it was better to make it more explicit and simple.

apps/finance/contracts/Finance.sol
sohkai: Mmm, I've generally been prefering to keep the "hidden" classes in the explicit inheritance tree, especially in the case of things like shell contracts solely meant to share constants and the like:

1. It explicitly tells readers that we're using the functionality, rather than guessing where it might come from
1. In case it gets removed from a upstream class, nothing breaks. In this case, it's not even `AragonApp` that's inheriting it, but `VaultRecoverable`.

Both of the above are intuitions from the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter).

But this is definitely a bit of a gray zone. I think I prefer to keep it here, because, applying LoD, `AragonApp` doesn't make it obvious that it includes `EtherTokenConstant`. As another example, if a contract were to be `Petrifiable` but also use `TimeHelpers`, I'd include both even though `Petrifiable` uses `TimeHelpers` under the hood, as it's not immediately obvious that `Petrifiable` is related.

An example in the other direction is something like `AppStorage`. If you're using an `AragonApp`, I wouldn't include `AppStorage` again if you're explicitly using the `apps` state as the two are closely coupled.

apps/survey/test/survey.js
```diff
@@ -6,25 +6,53 @@ const getContract = name => artifacts.require(name)
...
   const holder31 = accounts[1]
   const holder50 = accounts[2]
   const nonHolder = accounts[4]
   const NULL_ADDRESS = '0x00'
+  const minimumAcceptanceParticipationPct = pct16(20)
+
+  before(async () => {
+    const kernelBase = await getContract('Kernel').new(true) // petrify immediately
+    const aclBase = await getContract('ACL').new()
+    const regFact = await getContract('EVMScriptRegistryFactory').new()
```
sohkai: One small nitpick: I don't think we need `EVMScripts` with Survey, so not using one in the `DAOFactory` will speed up the tests.

apps/finance/contracts/Finance.sol
bingen: A little related annoyance is the problem with C3 linearization. Trying to put it back the way it was made Solidity fail to compile after merging #410. Same as here I guess: https://github.com/aragon/aragon-apps/pull/256#discussion_r211939204

bingen: I guess we could adopt the rule of adding them from “most base-like” to “most derived”, as it's recommended in Solidity documentation: https://solidity.readthedocs.io/en/v0.4.24/contracts.html#multiple-inheritance-and-linearization

sohkai: Yeah... the order will have to be swapped. Admittedly it makes it not very nice :(.

It's hard to really tell what's "most base-like" some times, but with unstructured storage this is less of a big deal than before.

bingen:
The CI is taking forever but I ran tests locally and they worked, so merging it.

PR #425, opened 2018-08-30
Merge Old vault and upgrade to aragonOS 4

bingen:
Fixed first 4 items from TODO list in #422 

- [X] Update and merge #292
- [X] Fix tests for Finance and Survey by using proxied versions (see Token Manager, Voting)
- [X] Fix compile warnings
- [X] Upgrade and fix linter warnings (will be getting `solium-config-aragon` soon on npm...)
- [x] Update and merge #256, #410 
- [x] Add new tests for app lifecycle (e.g. checking that the base is immediately petrified and unusable)


apps/finance/contracts/Finance.sol
```diff
@@ -94,7 +93,7 @@ contract Finance is AragonApp, EtherTokenConstant {
     // is changed before the operation if needed
     modifier transitionsPeriod {
         bool completeTransition = tryTransitionAccountingPeriod(getMaxPeriodTransitions());
-        require(completeTransition);
+        require(completeTransition, "Incomplete transition");
```
izqui: I wouldn't add revert reasons for now, as hardcoding strings in the contract is going to increase the deployment gas costs enormously. I'd say let's not do it for now and think what the best way to do it would be. This is a good start: https://ethereum-magicians.org/t/building-a-taxonomy-of-smart-contract-errors/1170/7

apps/survey/contracts/Survey.sol
```diff
@@ -109,7 +108,7 @@ contract Survey is AragonApp {
         survey.metadata = _metadata;
         survey.snapshotBlock = getBlockNumber() - 1; // avoid double voting in this very block
         survey.votingPower = token.totalSupplyAt(survey.snapshotBlock);
-        require(survey.votingPower > 0);
+        assert(survey.votingPower > 0);
```
izqui: Why the change to `assert`? Should still be `require` as the token supply could be 0 when a new survey is created, no need to waste all the gas for something that is not an internal consistency.

apps/voting/contracts/Voting.sol
```diff
@@ -155,49 +159,64 @@ contract Voting is IForwarder, AragonApp {
...
-        Vote storage vote = votes[_voteId];
+        Vote storage voteTmp = votes[_voteId];
 
-        return _isVoteOpen(vote) && token.balanceOfAt(_voter, vote.snapshotBlock) > 0;
+        return _isVoteOpen(voteTmp) && token.balanceOfAt(_voter, voteTmp.snapshotBlock) > 0;
     }
 
     function canExecute(uint256 _voteId) public view returns (bool) {
-        Vote storage vote = votes[_voteId];
+        Vote storage voteTmp = votes[_voteId];
```
izqui: Is the rename of `vote` required so it doesn't clash with the `vote` function? The `voteTmp` name feels a bit off, specially when used below when we modify the storage by changing values in `voteTmp`.

bingen: (and yes, I agree with your comment, but my rationale was that I was thinking of `voteTmp` as a pointer to an element in the array)

sohkai: Another typically used naming scheme is `vote_` for local variables.

sohkai:
@bingen Wondering if we can clean this branch so it's just the changes for the `Vault` and `Finance` :).

In particular, I think we could cherry pick all the commits into #422, ignore any conflicts in the vault or finance, and then rebase / restart this branch. Let me know if you need any help :).

apps/finance/.solcover.js
```diff
@@ -1,6 +1,8 @@
 module.exports = {
-    norpc: true,
```
sohkai: I think we should keep the `norpc: true` flag, since we start our own rpc instance in the test script.

shared/test-helpers/ganache-cli.sh
```diff
@@ -15,10 +15,10 @@ testrpc_running() {
 
 start_testrpc() {
   if [ "$SOLIDITY_COVERAGE" = true ]; then
-    npx testrpc-sc -i 16 --gasLimit 0xfffffffffff --port "$testrpc_port"  > /dev/null &
```
sohkai: Mmmm, I'd like to keep this the same as in aragonOS; was there a reason for not starting the testrpc instance here? I'm not sure what flags `solidity-coverage` uses when it starts `testrpc-sc`.

bingen: Yes, I was trying to follow these instructions: 
https://github.com/sc-forks/solidity-coverage/issues/131#issuecomment-417155288
So first I want to be able to make it work (no luck so far) and then I'll care about those details ;-) But, yes, I agree it would be better to keep like it was.

apps/finance/test/finance.js
```diff
@@ -91,6 +91,7 @@ contract('Finance App', accounts => {
 
     it('sets the end of time correctly', async () => {
         const { financeApp } = await newProxyFinance()
+        await financeApp.mock_setTimestamp(100) // to make sure it overflows
```
sohkai: I'd expand on this comment, to something like `to make sure it overflows with MAX_UINT64 period length`

apps/survey/contracts/Survey.sol
```diff
@@ -244,6 +256,7 @@ contract Survey is AragonApp {
         return surveys[_surveyId].metadata;
     }
 
+    /* solium-disable function-order */
```
sohkai: We should probably put this at the top, or disable it for a particular line?

This declaration will disable the rule from this point on in the file.

apps/token-manager/contracts/TokenManager.sol
```diff
@@ -234,6 +253,7 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
         return true;
     }
 
+    /* solium-disable function-order */
```
sohkai: See comment in Survey.sol

apps/token-manager/contracts/TokenManager.sol
```diff
@@ -350,10 +370,9 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
 
     /**
     * @notice Called when `_owner` sends ether to the MiniMe Token contract
```
sohkai: Should probably also modify the `@notice` to omit the owner :)

apps/vault/contracts/Vault.sol
```diff
@@ -1,118 +1,114 @@
...
-import "./VaultBase.sol"; // split made to avoid circular import
+import "@aragon/os/contracts/apps/AragonApp.sol";
 
-import "./connectors/ERC20Connector.sol";
-import "./connectors/ETHConnector.sol";
+import "@aragon/os/contracts/lib/token/ERC20.sol";
 
-import "@aragon/os/contracts/lib/misc/Migrations.sol";
 
+contract Vault is AragonApp {
```
sohkai: Maybe add `EtherTokenConstant` in the inheritance list, based on earlier discussion

PR #426, opened 2018-08-31
Fix decimals in radspec strings

sohkai:
**Doesn't change the Finance app's artifact.json**

Uses external calls in radspec to get the decimal value to use.


izqui:
The `deposit()` function in Finance will accept ETH as a token with #451, so its radspec string will also need to be updated to support the case when the deposited token is ETH.

izqui:
I'm thinking that we should introduce the concept of helper functions in radspec for common operations such as token amount formatting, to avoid super long radspec operations that make reading the description quite hard.

For example the radspec string 'Transfer `@formatTokenAmount(_token, _amount)` to Vault' could execute some JS code that can take into account token decimals, check for the ETH constant and also add the symbol of the token.

sohkai:
Yeah, that would be nice. These radspec strings are getting a little bit out of hand. Question is where we define them; I guess it wouldn't hurt to start with our own use cases and branch out from there.

izqui:
Yep, my idea is that radspec would come with its own set of helpers and those could potentially be extended by passing extra helper functions when executing `radspec.evaluate`. I think for now providing a few token formatting and time helpers would go a long way.

chris-remus:
@bingen to pick this up.

izqui:
Made [some changes](https://github.com/aragon/radspec/pull/42/commits/c6cd40356bc1f15c4acdf8c38871de167ea47e56#diff-46ce6852901fef83cf036e191e10a50cR3) to the `transformTime` helper so it always transforms time to the best `toUnit` by default, so we can remove 'humanize'.

PR #428, opened 2018-09-04
Convert dynamic arrays to mapping

### LINKED ISSUE #68: Make app storage layout more graceful to app upgrades
See my explanation: https://youtu.be/sJ7VECqHFAg?t=9m27s

Basically convert arrays to mappings.

bingen:
To make app storage layout more graceful to app upgrades.

Closes #68.


apps/finance/test/finance.js
```diff
@@ -382,9 +383,10 @@ contract('Finance App', accounts => {
 
         it('doesnt record payment for one time past transaction', async () => {
             await finance.newPayment(token1.address, recipient, 1, time, 1, 1, '')
-            return assertRevert(async () => {
-                await finance.getPayment(1)
```
sohkai: Debating whether or not we should keep these getters reverting if someone requests an out-of-bounds index

apps/token-manager/contracts/TokenManager.sol
```diff
@@ -40,7 +40,9 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
         bool revokable;
     }
 
-    mapping (address => TokenVesting[]) internal vestings;
+    // we are mimicing an array inside the outer mapping, we use a mapping instead to make app upgrade more graceful
+    mapping (address => mapping (uint256 => TokenVesting)) internal vestings;
+    mapping (address => uint256) internal vestingsLength;
```
sohkai: I'd rename this variable to `vestingLengths`, as we have multiple lengths in the mapping

apps/finance/test/finance.js
sohkai: Internal debate ended with adding checks for the length in all of the affected getters.

As we were using arrays before, and those throw `invalid opcode` when accessing out-of-bounds indices, we should aim for the same interface. I think it makes sense to for the caller to know if they've tried to access an out of bounds.

apps/finance/contracts/Finance.sol
```diff
@@ -78,9 +78,13 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
 
     Vault public vault;
 
-    Payment[] internal payments; // first index is 1
-    Transaction[] internal transactions; // first index is 1
-    Period[] internal periods; // first index is 0
+    // we are mimicing arrays, we use mappings instead to make app upgrade more graceful
+    mapping (uint256 => Payment) internal payments; // first index is 1
+    uint256 internal paymentsLength;
```
sohkai: I think all of these lengths storage variables should be public

sohkai:
@izqui we were actually having a conversation about this; what do you think about starting the IDs from 0 in all those cases where we're not? It'd simplify checking for off-by-one errors.

sohkai:
Ok, so turns out the finance does need this, for the notion of transactions that are not linked to any recurring payments.

izqui:
We use `transactionId = 0` to denote when no transactions have occurred in a period yet, [see](https://github.com/aragon/aragon-apps/blob/d021091977feeaa5bcc783dee8e45ad6ac885b14/apps/finance/contracts/Finance.sol#L670). Reverting to starting `transactionId` from index 1.

apps/finance/contracts/Finance.sol
```diff
@@ -22,6 +22,7 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
     uint64 constant public MAX_PAYMENTS_PER_TX = 20;
     uint64 constant public MAX_UINT64 = uint64(-1);
     uint256 constant public MAX_UINT = uint256(-1);
+    uint256 constant NO_TRANSACTION = 0;
```
sohkai: Let's make this explicitly internal, and maybe we could also add a constant for `NO_PAYMENT`?

sohkai: Fixed in d39b2da.

PR #429, opened 2018-09-05
Publish internal packages

sohkai:
Moves the internal packages to be scoped as `@aragon/apps-shared-*` so they can be published.

Both are now published: [minime](https://www.npmjs.com/package/@aragon/apps-shared-minime), [migrations](https://www.npmjs.com/package/@aragon/apps-shared-migrations).

### DIFF
--- a/shared/migrations/package.json
+++ b/shared/migrations/package.json
@@ -2,7 +2,6 @@
   "name": "@aragon/apps-shared-migrations",
   "version": "1.0.0",
   "description": "",
-  "private": true,
   "scripts": {
     "test": "echo \"No tests\""
   }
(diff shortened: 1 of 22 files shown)


shared/migrations/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@aragon-apps/migrations",
+  "name": "@aragon/apps-shared-migrations",
   "version": "1.0.0",
   "description": "",
   "private": true,
```
izqui: We should remove `"private": true` now.

shared/minime/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@aragon-apps/minime",
+  "name": "@aragon/apps-shared-minime",
   "version": "1.0.0",
   "license": "GPL-3.0-or-later",
   "private": "true",
```
izqui: We should remove "private": true now.

PR #431, opened 2018-09-05
Upgrade to aragonOS@4.0.0-beta.1

### LINKED ISSUE #385: Vault should revert on deposit if Kernel reference hasn't been set

sohkai:
Fixes #385.

Also fixes an issue with the Token Manager where held tokens (e.g. those issued to the Token Manager) could be mistakenly (or maliciously) recovered into the default vault.

Todo:

- [x] Add tests for sending eth to uninitialized vaults
- [x] Modify kernel recovery test to use a depositable app stub instead
- [x] Figure out why the other test reverts on sending ETH to the vault with a higher gas limit...

### DIFF
--- a/apps/survey/package.json
+++ b/apps/survey/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@aragon/apps-survey",
-  "version": "0.0.3",
+  "version": "0.1.0",
   "main": "index.js",
   "scripts": {
     "postinstall": "cd app && npm install",
--- a/apps/survey/package.json
+++ b/apps/survey/package.json
@@ -48,6 +48,6 @@
   },
   "dependencies": {
     "@aragon/apps-shared-minime": "1.0.0",
-    "@aragon/os": "4.0.0-alpha.2"
+    "@aragon/os": "4.0.0-beta.1"
   }
 }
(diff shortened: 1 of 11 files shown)


apps/survey/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@aragon/apps-survey",
-  "version": "0.0.3",
+  "version": "0.1.0",
```
sohkai: I've just bumped this to `0.1.0` since it's not going to be audited yet.

sohkai:
Bumped the all the package versions to `beta.1`, and fixed the vault dependency. We can release and tag on `beta.1` before giving it to the auditors.

apps/vault/test/vault.js
```diff
@@ -242,7 +225,7 @@ contract('Vault app', (accounts) => {
         await assertRevert(() => vault.transferToVault(ETH))
       })
 
-      it('fails when attempting to recover ETH out of the vault', async () => {
+      it('fails when attempting to recover tokens out of the vault', async () => {
         await token.transfer(vault.address, 10)
         assert.equal((await token.balanceOf(vault.address)), 10, 'vault should have 10 balance')
         await assertRevert(() => vault.transferToVault(ETH))
```
izqui: Should be `vault.transferToVault(token.address)`

PR #433, opened 2018-09-11
Ignore Vim Tempfiles

wjmelements:
#### Changes
This adds vim tempfiles to the gitignore so they do not appear in `git status`.
Reviewers @izqui


sohkai:
@wjmelements These should belong in your global gitignore instead :).

wjmelements:
TIL, thanks!

PR #434, opened 2018-09-11
Move vote creator to event data

wjmelements:
#### Changes
This moves vote creator to event data because it isn't used anywhere.
It isn't especially important who created the vote.
For client interfaces we still want to note the creator, but the appropriate place to do that is event data.
This saves a minor amount of gas.
Reviewers @izqui

### DIFF
--- a/apps/voting/contracts/Voting.sol
+++ b/apps/voting/contracts/Voting.sol
@@ -31,7 +31,6 @@ contract Voting is IForwarder, AragonApp {
     enum VoterState { Absent, Yea, Nay }
 
     struct Vote {
-        address creator;
         uint64 startDate;
         uint256 snapshotBlock;
         uint256 minAcceptQuorumPct;
@@ -46,7 +45,7 @@ contract Voting is IForwarder, AragonApp {
 
     Vote[] internal votes; // first index is 1
 
-    event StartVote(uint256 indexed voteId);
+    event StartVote(uint256 indexed voteId, address indexed creator);
     event CastVote(uint256 indexed voteId, address indexed voter, bool supports, uint256 stake);
     event ExecuteVote(uint256 indexed voteId);
     event ChangeMinQuorum(uint256 minAcceptQuorumPct);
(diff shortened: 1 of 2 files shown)


izqui:
Thanks for bringing this up @wjmelements. I actually don't think we need those on-chain so they can be safely moved into a log, specially creator which isn't being used anywhere.

wjmelements:
I will make similar changes in the survey contract then.

wjmelements:
Further changes are required in `app/src/script.js` to fetch the data from the event logs.

sohkai:
Just realized, the Finance app also has an instance of `creator` and `reference` that we might be able to remove to save gas.

apps/survey/test/survey.js
```diff
@@ -5,6 +5,7 @@ const timeTravel = require('@aragon/test-helpers/timeTravel')(web3)
 const getContract = name => artifacts.require(name)
 const pct16 = x => new web3.BigNumber(x).times(new web3.BigNumber(10).toPower(16))
 const createdSurveyId = receipt => receipt.logs.filter(x => x.event == 'StartSurvey')[0].args.surveyId
```
bingen: Maybe you could get rid of this one and just use the newly created one below

chris-remus:
Pending decision, do this for Finance? @bingen @sohkai @izqui to talk and decide later today.

apps/finance/contracts/Finance.sol
```diff
@@ -674,14 +667,13 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
         transaction.token = _token;
         transaction.entity = _entity;
         transaction.date = getTimestamp64();
-        transaction.reference = _reference;
```
sohkai: @bingen It might be nice to re-order this to reflect the order of the the [`Transaction`](https://github.com/aragon/aragon-apps/blob/03cb159a3f1280c9f7d076433aad8c0b533b6659/apps/finance/contracts/Finance.sol#L67) struct itself. Right now the storage is being set out of order, and I wouldn't count on the compiler to optimize it correctly for us.

sohkai:
Finally merging this breaking change @wjmelements!

wjmelements:
Hurray!
I'm sorry I didn't have time to finish this.

PR #435, opened 2018-09-13
Vault: allow recoverability is it has not been initialized

bingen:


apps/vault/contracts/Vault.sol
```diff
@@ -112,6 +112,6 @@ contract Vault is EtherTokenConstant, AragonApp, DepositableStorage {
     *      maliciously to transfer funds away from the vault
     */
     function allowRecoverability(address) public view returns (bool) {
-        return false;
+        return hasInitialized() ? false : true;
```
sohkai: I was thinking about this, and wondering what would be best for native ETH. If the vault's not initialized, it should be impossible to deposit ETH into the contract anyway, so it might make sense to say that we disallow recovering ETH too.

Otherwise, we could just do a `!hasInitialized()` here.

apps/vault/test/vault.js
```diff
@@ -230,6 +230,39 @@ contract('Vault app', (accounts) => {
         assert.equal((await token.balanceOf(vault.address)), 10, 'vault should have 10 balance')
         await assertRevert(() => vault.transferToVault(ETH))
       })
+
+      it('checks result of allowRecoverability', async () => {
```
sohkai: I would rename this to say "does not allow recoverability"

apps/vault/test/vault.js
```diff
@@ -230,6 +230,39 @@ contract('Vault app', (accounts) => {
         assert.equal((await token.balanceOf(vault.address)), 10, 'vault should have 10 balance')
         await assertRevert(() => vault.transferToVault(ETH))
       })
+
+      it('does not allow recoverability', async () => {
+        assert.isFalse(await vault.allowRecoverability(token.address))
+        assert.isFalse(await vault.allowRecoverability(ETH))
```
sohkai: Would also be great to test the entire flow, trying to actually recover funds from the vault.

bingen: This is done above:
https://github.com/aragon/aragon-apps/pull/435/files#diff-4182395b6bdaf5bbad54a4607aa62a87R208
https://github.com/aragon/aragon-apps/pull/435/files#diff-4182395b6bdaf5bbad54a4607aa62a87R217
I  just wanted to add the explicit call to that function just in case and for comparison with the non-init case.

sohkai:
@bingen I would move 5715578 into its own PR :)

sohkai:
Hmm, this is a good point; I think given that, the recoverability would only make sense if the vault somehow ended up being petrified (e.g. by being customized with a subclass).

Given that this wouldn't be possible normally, since the base would be deployed without a kernel and all proxies can't be petrified without custom code, it might make sense to keep the recoverability off at all times.

PR #439, opened 2018-09-13
Survey: check if option exists when getting option power

sohkai:
Reverts if trying to get an option power for an option that doesn't exist in a survey.


sohkai:

PR #444, opened 2018-09-13
Voting: add flag to newVote() to decide if the vote should also immediately execute if decided by creator's vote

sohkai:


sohkai:
@izqui @bingen The tests weren't passing for this yet :(. The mock changes the `msg.sender` for `newVote()` so it's a bit of a pain to figure out how to best rewrite these tests.

sohkai:

PR #445, opened 2018-09-13
Voting: reorder min quorum sanity checks for clarity

sohkai:


sohkai:

PR #446, opened 2018-09-13
Move protection checks to external interfaces and optimize them

sohkai:
Simplifies the mental model when checking for access control: all publicly accessible functions include protection modifiers, while internal functions are just logic.


apps/voting/contracts/Voting.sol
```diff
@@ -150,13 +154,13 @@ contract Voting is IForwarder, AragonApp {
...
     */
-    function forward(bytes _evmScript) public isInitialized {
+    function forward(bytes _evmScript) public {
         require(canForward(msg.sender, _evmScript));
         _newVote(_evmScript, "", true);
     }
 
     function canForward(address _sender, bytes) public view returns (bool) {
-        return canPerform(_sender, CREATE_VOTES_ROLE, arr());
+        return hasInitialized() && canPerform(_sender, CREATE_VOTES_ROLE, arr());
```
izqui: `canPerform` already does initialization check. I would remove the check from this function and add a comment clarifying it.

PR #449, opened 2018-09-15
Voting: display the balance at the time the vote was created

bpierre:


bpierre:
@sohkai oops I missed that, I’m closing this 👍

Issue #216, opened 2018-03-28
Voting: add manual execute vote action


izqui:
As per http://wiki.aragon.one/dev/apps/voting/ there are 2 ways a vote can be executed:
- During vote time: the vote will be execute **iff** it is absolutely decided. If a vote requires 50% support, it will be executed on the vote that makes it have 50% absolute support taking the token total supply into account. These votes get automatically executed.
- After vote time: the vote can be executed (needs to be triggered manually by any account) if the relative support of the vote (just the holders that voted on time) is at least the support required (e.g. 50%) and at least minimumAcceptanceQuorum of the total token supply voted yes

For the second case we need a button after no more votes can be casted, to execute the vote. This action (`voting.execute(voteId)`) can be done by anyone without permissions if `voting.canExecute(voteId)` returns true.

We should probably also have a badge in the app to indicate whether a vote has been executed or not

sohkai:
@jounih We should also have a visible state for each vote that requires execution. Maybe something like a "Pending execution" warning.

For example, this is a vote requiring manual execution but the UI currently makes it look as if it's already been executed:

![image](https://user-images.githubusercontent.com/4166642/44954295-1c717c80-aea0-11e8-943a-d4d957086993.png)

![image](https://user-images.githubusercontent.com/4166642/44954296-1e3b4000-aea0-11e8-83df-e49e59de47da.png)

jounih:
Discussed this with @sohkai and @chris-remus - we will add another list view for "Pending execution" votes to the top, and a button to execute them in the side panel.

We will also change the status "Approved" to "Approved, executed", and "Approved, pending" or similar

<img width="1392" alt="screen shot 2018-09-05 at 18 35 58" src="https://user-images.githubusercontent.com/10109867/45110575-958c0080-b13a-11e8-9216-88e201e57e15.png">

jounih:
![webapp-1366px - voting 2x](https://user-images.githubusercontent.com/10109867/45358017-00688c00-b5c0-11e8-9673-7ff0113bc495.jpg)

PR #448, opened 2018-09-15
Voting: updates

bpierre:
- [x] Move to parcel (from react-scripts)
- [x] Move to react-spring (from react-motion)
- [x] Card layout (issue: https://github.com/aragon/design/issues/10)
- [x] Display the user vote
- [x] Display the correct balance (PR: https://github.com/aragon/aragon-apps/pull/449)
- [x] Execute a vote (issue: https://github.com/aragon/aragon-apps/issues/216#issuecomment-420243809, PR: https://github.com/aragon/aragon-apps/pull/450)
- [x] Explicitly change a vote (PR: https://github.com/aragon/aragon-apps/pull/450)

PR #450, opened 2018-09-15
Voting panel: execute + change vote

bpierre:
Related issue: https://github.com/aragon/aragon-apps/issues/216

### DIFF
--- a/apps/voting/app/src/components/VotePanelContent.js
+++ b/apps/voting/app/src/components/VotePanelContent.js
@@ -14,7 +14,8 @@ import {
 } from '@aragon/ui'
 import { combineLatest } from '../rxjs'
 import provideNetwork from '../utils/provideNetwork'
-import { VOTE_NAY, VOTE_YEA } from '../vote-types'
+import { VOTE_NAY, VOTE_YEA, VOTE_STATUS_ACCEPTED } from '../vote-types'
+import { getVoteStatus } from '../vote-utils'
 import VoteSummary from './VoteSummary'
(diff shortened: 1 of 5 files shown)


apps/voting/app/src/components/VotePanelContent.js
```diff
@@ -180,30 +205,82 @@ class VotePanelContent extends React.Component {
...
+                </ButtonsContainer>
+                <Info.Action>
+                  <p>
+                    You will cast {userBalance || '…'} votes, since it was your
+                    token balance at the{' '}
+                    <SafeLink
+                      href={`${etherscanBaseUrl}/block/${snapshotBlock}`}
+                      target="_blank"
+                    >
+                      beginning of the vote
```
izqui: I think we should print the block number here or add it on top right next to `quorum`, thoughts?

izqui:
I'm also seeing some state propagation issues as in #448, after changing my vote the card view doesn't update the state.

2018-09-18 at 10 45 33 am

I'm testing using `aragon run` with a local chain.

apps/voting/app/src/components/VotePanelContent.js
bpierre: Yes maybe we could have it at the top? @jounih 

In the meantime, I added it here.

izqui:
@bpierre @jounih Replying to https://github.com/aragon/aragon-apps/pull/450#discussion_r218734782

👍. Should we also add it to the `You casted x votes` copy?

bpierre:
@izqui yes, adding it

PR #451, opened 2018-09-17
Finance: refactor deposit and allow recoverability before initialization

izqui:
- The logic for depositing and creating an incoming transaction record has been converged into an internal `_deposit` function.
- We now allow calling the `deposit` function for depositing `ETH`, so a proper reference can be provided.
- Similar to #435, funds from the Finance app can only be recovered to the default Vault before the app has been initialized. Added tests for recovery, maybe they can be used in #435 as well @bingen. I think the test for recovering ETH may fail when run in coverage mode, we will need to disable it as we do in aragonOS. 
- `depositToVault` has been renamed to `recoverToVault`, which transfers assets in the Finance app to its own vault. Recovering stuck ETH can be done with `recoverToVault` now, rather than by sending at least 1 wei to the fallback function as the fallback previously sent the entire contract balance to the Vault when executed.

I have also been thinking about whether we should enable `depositable` in the Finance app, so ETH transfers can be done just with `finance.transfer` or `finance.send`. I'm currently siding against enabling it, as when users try to send ETH the gas estimation will be just enough to do a deposit in the proxy, resulting in the funds being stuck in the Finance app until `recoverToVault` is called. If someone wants to send ETH via a smart contract, they could either transfer to `finance.vault()` directly or do a proper `finance.deposit(...)` call.


apps/finance/contracts/Finance.sol
```diff
@@ -136,49 +136,21 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
     }
 
     /**
-    * @dev Deposit for approved ERC20 tokens
+    * @dev Deposit for approved ERC20 tokens or ETH
     * @notice Deposit `_amount / 10^18` `_token.symbol(): string`
```
izqui: Needs to be updated, waiting for #426

apps/finance/contracts/Finance.sol
```diff
@@ -507,6 +485,37 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
 
     // internal fns
 
+    function _deposit(address _token, uint256 _amount, string _reference, address _sender, bool _isExternalDeposit) internal {
+        require(_amount > 0);
+        _recordIncomingTransaction(
```
izqui: Maybe we should do the record at the end so if something reverts we don't waste the gas spent in storage here

apps/finance/test/finance.js
```diff
@@ -255,17 +355,6 @@ contract('Finance App', accounts => {
         assert.equal(await finance.currentPeriodId(), 1, 'should have transitioned 1 periods')
     })
 
-    it("escapes hatch, recovers ETH", async () => {
```
izqui: Removed this test as it is testing the same as the `records ETH deposits using fallback` test case

apps/finance/contracts/Finance.sol
sohkai: It might not be safe to do so, to avoid reentrancy issues.

apps/finance/contracts/Finance.sol
```diff
@@ -136,49 +136,21 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
...
+    * @dev Deposit for approved ERC20 tokens or ETH
     * @notice Deposit `_amount / 10^18` `_token.symbol(): string`
     * @param _token Address of deposited token
     * @param _amount Amount of tokens sent
     * @param _reference Reason for payment
     */
-    function deposit(ERC20 _token, uint256 _amount, string _reference) external isInitialized transitionsPeriod {
-        require(_amount > 0);
-        _recordIncomingTransaction(
+    function deposit(ERC20 _token, uint256 _amount, string _reference) external payable isInitialized transitionsPeriod {
```
bingen: Shouldn't `ERC20` here be `address`? As now it can also be `ETH`.

apps/finance/contracts/Finance.sol
```diff
@@ -328,20 +300,17 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
      * @notice Send tokens held in this contract to the Vault
      * @param _token Token whose balance is going to be transferred.
      */
-    function depositToVault(ERC20 _token) public isInitialized {
-        uint256 value = _token.balanceOf(this);
-        require(value > 0);
+    function recoverToVault(ERC20 _token) public isInitialized transitionsPeriod {
```
bingen: Same here: `ERC20` -> `address`?

apps/finance/test/finance.js
```diff
@@ -181,39 +203,117 @@ contract('Finance App', accounts => {
...
+        const receipt = await finance.send(sentWei)
+        const transactionId = receipt.logs.filter(log => log.event == 'NewTransaction')[0].args.transactionId
 
-        const [periodId, amount, paymentId, paymentRepeatNumber, token, entity, incoming, date, ref] = await finance.getTransaction(1)
+        const [periodId, amount, paymentId, paymentRepeatNumber, token, entity, incoming, date, ref] = await finance.getTransaction(transactionId)
 
-        let finalBalance = await token1.balanceOf(vault.address)
-        assert.equal(finalBalance.toString(), initialBalance.plus(5).toString(), 'deposited tokens must be in vault')
-        assert.equal(await token1.balanceOf(finance.address), 0, 'finance shouldn\'t have tokens')
+        assert.equal(await vault.balance(ETH), VAULT_INITIAL_ETH_BALANCE + 10, 'deposited ETH must be in vault')
```
bingen: Not a big deal but I would use here `sentWei` above instead of `10`.

apps/finance/contracts/Finance.sol
izqui: I can't see how we can avoid reentrancy in this case, if the token contract is malicious this function can re-enter regardless of where we record the transaction

PR #454, opened 2018-09-17
Voting: add flag to newVote() to decide if the vote should also immediately execute if decided by creator's vote

bingen:
Replaces #444


sohkai:
Note that the tests for this will **not** be passing yet.

bingen:
Which ones? I tried locally and it worked for me. I just tried Voting app though, as it's the only one changed.

sohkai:
Oh that's funny, I swear it wasn't working for me locally when I was testing the changes. Seems to be working now!

Issue #270, opened 2018-04-19
Minting 1 new token for a new address turns the 2of2 into a 3of3.


john-light:
Multisig, 2of2, Token Manager, Mint 1 Token to new address. Minting 1 new token for a new address turns the 2of2 into a 3of3.

This was unexpected. It wasn't clear what minting a token to a new address would do.

chris-remus:
TBD on additional scope. @izqui will add more perspective here.

sohkai:
Removing from mainnet scope due to requirement on https://github.com/aragon/aragon/issues/363.

PR #458, opened 2018-09-18
Voting: Allow to modify required support

bingen:
Snapshot it for each vote, create role for it and add tests.

Needed for #270

### DIFF
--- a/apps/voting/contracts/Voting.sol
+++ b/apps/voting/contracts/Voting.sol
@@ -18,6 +18,7 @@ contract Voting is IForwarder, AragonApp {
     using SafeMath64 for uint64;
 
     bytes32 public constant CREATE_VOTES_ROLE = keccak256("CREATE_VOTES_ROLE");
+    bytes32 public constant MODIFY_SUPPORT_ROLE = keccak256("MODIFY_SUPPORT_ROLE");
     bytes32 public constant MODIFY_QUORUM_ROLE = keccak256("MODIFY_QUORUM_ROLE");
 
     uint256 public constant PCT_BASE = 10 ** 18; // 0% = 0; 1% = 10^16; 100% = 10^18
@@ -28,6 +29,7 @@ contract Voting is IForwarder, AragonApp {
         address creator;
         uint64 startDate;
         uint256 snapshotBlock;
+        uint256 supportRequiredPct;
         uint256 minAcceptQuorumPct;
         uint256 yea;
         uint256 nay;
@@ -50,6 +52,7 @@ contract Voting is IForwarder, AragonApp {
     event StartVote(uint256 indexed voteId);
     event CastVote(uint256 indexed voteId, address indexed voter, bool supports, uint256 stake);
     event ExecuteVote(uint256 indexed voteId);
+    event ChangeSupportRequired(uint256 supportRequiredPct);
     event ChangeMinQuorum(uint256 minAcceptQuorumPct);
 
     modifier voteExists(uint256 _voteId) {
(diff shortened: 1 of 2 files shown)


apps/voting/contracts/Voting.sol
```diff
@@ -85,6 +88,21 @@ contract Voting is IForwarder, AragonApp {
         voteTime = _voteTime;
     }
 
+    /**
+    * @notice Change required support to `(_supportRequiredPct - _supportRequiredPct % 10^16) / 10^14`%
```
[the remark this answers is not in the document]
bingen: It was supposed to round to 2 decimals, but I think it's wrong, `14` and `16` should be reverted.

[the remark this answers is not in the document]
bingen: For some reason I changed it in commit 6bb186bbba641fea32a9f21d6ff56b627e4dece5 :thinking:

apps/voting/test/voting.js
```diff
@@ -174,12 +196,13 @@ contract('Voting App', accounts => {
...
 
             it('has correct state', async () => {
-                const [isOpen, isExecuted, creator, startDate, snapshotBlock, minQuorum, y, n, totalVoters, execScript] = await voting.getVote(voteId)
+                const [isOpen, isExecuted, creator, startDate, snapshotBlock, supportRequired, minQuorum, y, n, totalVoters, execScript] = await voting.getVote(voteId)
 
                 assert.isTrue(isOpen, 'vote should be open')
                 assert.isFalse(isExecuted, 'vote should not be executed')
                 assert.equal(creator, holder50, 'creator should be correct')
                 assert.equal(snapshotBlock, await getBlockNumber() - 1, 'snapshot block should be correct')
+                assert.deepEqual(supportRequired, neededSupport, 'required support should be app required support')
```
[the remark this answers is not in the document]
bingen: I agree.

apps/voting/contracts/Voting.sol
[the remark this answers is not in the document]
izqui: Another helper function for radspec :)

apps/voting/test/voting.js
```diff
@@ -144,12 +146,32 @@ contract('Voting App', accounts => {
...
 
+        it('can change required support', async () => {
+            const receipt = await voting.changeSupportRequiredPct(neededSupport.add(1))
+            const events = receipt.logs.filter(x => x.event == 'ChangeSupportRequired')
+
+            assert.equal(events.length, 1, 'should have emitted ChangeSupportRequired event')
+            assert.equal((await voting.supportRequiredPct()).toString(), neededSupport.add(1).toString(), 'should have changed required support')
+        })
+
+        it('fails changing required supoprt lower than minimum acceptance quorum', async () => {
```
izqui: typo: supoprt -> support

apps/voting/contracts/Voting.sol
```diff
@@ -251,6 +271,7 @@ contract Voting is IForwarder, AragonApp {
         vote_.metadata = _metadata;
         vote_.snapshotBlock = getBlockNumber() - 1; // avoid double voting in this very block
         vote_.totalVoters = token.totalSupplyAt(vote_.snapshotBlock);
+        vote_.supportRequiredPct = supportRequiredPct;
```
izqui: 🚨 The cached `vote_.supportRequiredPct` is currently not being used to check whether a vote can be executed in `canExecute()`.

https://github.com/aragon/aragon-apps/blob/97ca6b0ca0d07795208b7901903de470fa75035f/apps/voting/contracts/Voting.sol#L210

apps/voting/test/voting.js
```diff
@@ -195,6 +218,22 @@ contract('Voting App', accounts => {
...
+                // With previous required support at 50%, vote should be approved
+                // with new quorum at 70% it shouldn't have, but since min quorum is snapshotted
+                // it will succeed
+
+                await voting.vote(voteId, true, true, { from: holder50 })
+                await voting.vote(voteId, true, true, { from: holder19 })
+                await timeTravel(votingTime + 1)
+
+                const state = await voting.getVote(voteId)
+              assert.equal(state[5].toNumber(), neededSupport.toNumber(), 'required support in vote should stay equal')
```
izqui: Proper indent 🙏

apps/voting/test/voting.js
```diff
@@ -195,6 +218,22 @@ contract('Voting App', accounts => {
...
 
+          it('changing required support does not affect vote required support', async () => {
+                await voting.changeSupportRequiredPct(pct16(70))
+
+                // With previous required support at 50%, vote should be approved
+                // with new quorum at 70% it shouldn't have, but since min quorum is snapshotted
+                // it will succeed
+
+                await voting.vote(voteId, true, true, { from: holder50 })
+                await voting.vote(voteId, true, true, { from: holder19 })
```
izqui: This test is not actually testing the logic we want to test, as both holders are voting YES, the support for the proposal is 100%.

Because we are not properly using the cached `supportRequiredPct` for the `canExecute` calculation. 

We should add a NO vote by `holder31`, which would make the support of the proposal be 69, and would make this test fail until the issue with `canExecute` is fixed

apps/voting/contracts/Voting.sol
bingen: Good catch, thanks!! I see that the test for it below is wrong too.

PR #459, opened 2018-09-18
Voting: make inequality strict

### LINKED ISSUE #427: Voting: making inequality for passing votes strict?
Related to https://github.com/aragon/governance/issues/32#issuecomment-416337704
It makes sense for Democracy DAOs but it doesn't for Multisigs, as @sohkai pointed out, e.g., a 2 out of 4 signatures Multisig would require the inequality not to be strict to easily set up a 50%.
An option would be to add a boolean initialization parameter to indicate how inequalities (for both quorum and support) should behave (something like `bool strictThresholds`), which would be set up to by default to `true` for Democracy DAO kit and to `false` for Multisig.
Opening this issue for discussion.

bingen:
Closes #427.

### DIFF
--- a/apps/voting/test/voting.js
+++ b/apps/voting/test/voting.js
@@ -66,10 +66,10 @@ contract('Voting App', accounts => {
         await acl.createPermission(ANY_ADDR, voting.address, MODIFY_QUORUM_ROLE, root, { from: root })
     })
 
-    context('normal token supply', () => {
-        const holder19 = accounts[0]
-        const holder31 = accounts[1]
-        const holder50 = accounts[2]
+    context('normal token supply, common tests', () => {
+        const holder20 = accounts[0]
+        const holder29 = accounts[1]
+        const holder51 = accounts[2]
         const nonHolder = accounts[4]
 
         const neededSupport = pct16(50)
@@ -77,10 +77,6 @@ contract('Voting App', accounts => {
 
         beforeEach(async () => {
             token = await MiniMeToken.new(NULL_ADDRESS, NULL_ADDRESS, 0, 'n', 0, 'n', true) // empty parameters minime
-
-            await token.generateTokens(holder19, 19)
-            await token.generateTokens(holder31, 31)
-            await token.generateTokens(holder50, 50)
 
             await voting.initialize(token.address, neededSupport, minimumAcceptanceQuorum, votingTime)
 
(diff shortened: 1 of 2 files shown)


apps/voting/test/voting.js
```diff
@@ -174,189 +139,234 @@ contract('Voting App', accounts => {
...
+            const holder51 = accounts[2]
+            const nonHolder = accounts[4]
+
+            const neededSupport = pct16(50)
+            const minimumAcceptanceQuorum = pct16(20)
 
             beforeEach(async () => {
+                token = await MiniMeToken.new(NULL_ADDRESS, NULL_ADDRESS, 0, 'n', decimals, 'n', true) // empty parameters minime
+
+                await token.generateTokens(holder20, 20)
```
izqui: Assigned balances should be multiplied by `10 ^ decimals`, otherwise we are just testing the same every time. You can use logic similar to the one in the `pct16` helper

apps/voting/test/voting.js
```diff
@@ -174,189 +139,234 @@ contract('Voting App', accounts => {
...
             return assertRevert(async () => {
                 await voting.changeMinAcceptQuorumPct(neededSupport.plus(1))
             })
         })
 
-        context('creating vote', () => {
-            let script, voteId
+    })
+
+    for (const decimals of [0, 2, 10, 18, 36]) {
```
izqui: Nice one. We probably don't need to test all these cases though, I would test a maximum of 3, otherwise it is going to take for ever to run voting tests

apps/voting/test/voting.js
```diff
@@ -174,189 +139,234 @@ contract('Voting App', accounts => {
...
+                    assert.equal(await voting.getVoteMetadata(voteId), 'metadata', 'should have returned correct metadata')
+                    assert.equal(await voting.getVoterState(voteId, nonHolder), VOTER_STATE.ABSENT, 'nonHolder should not have voted')
+                })
 
-                const state = await voting.getVote(voteId)
-                assert.equal(state[5].toNumber(), neededSupport.toNumber(), 'required support in vote should stay equal')
-                await voting.executeVote(voteId) // exec doesn't fail
-            })
+                it('fails getting a vote out of bounds', async () => {
+                    assertRevert(async () => {
```
izqui: await or return

apps/voting/test/voting.js
```diff
@@ -174,189 +139,234 @@ contract('Voting App', accounts => {
...
+                    })
+                })
 
-            it('vote can be not executed automatically if decided', async () => {
-                await voting.vote(voteId, true, false, { from: holder50 }) // doesnt cause execution
-                await voting.executeVote(voteId)
-                assert.equal(await executionTarget.counter(), 2, 'should have executed result')
-            })
+                it('cannot execute vote if not support met', async () => {
+                    await voting.vote(voteId, false, true, { from: holder29 })
```
izqui: Would be interesting to test when a vote gets 50% support but doesn't pass because of the strict check.

bingen:
Some of them are tested at the end, where it says `isValuePct unit test`. Let me know which ones you want to add and I'll do it.

apps/voting/test/voting.js
```diff
@@ -174,189 +140,234 @@ contract('Voting App', accounts => {
...
                 await voting.changeMinAcceptQuorumPct(neededSupport.plus(1))
             })
         })
 
-        context('creating vote', () => {
-            let script, voteId
+    })
+
+    for (const decimals of [0, 2, 18]) {
+        context('normal token supply, ' + decimals + ' decimals', () => {
```
sohkai: We can use the [backtick syntax](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals) here :)

PR #471, opened 2018-10-08
Voting fixes

sohkai:
Small fixes related to #448.


apps/voting/app/src/components/VoteStatus.js
```diff
@@ -14,25 +14,25 @@ const ATTRIBUTES = {
     label: 'Ongoing',
     Icon: IconTime,
     color: theme.textTertiary,
-    bord: false,
+    bold: false,
```
bpierre: 👀

Issue #461, opened 2018-09-18
Optimize contract state by using uint64


sohkai:
There are a number of storage variables (or in structs) in the Finance, Survey, Token Manager, and Voting that could be reduced to `uint64`, e.g. block number. Doing so would usually save at least one SSTORE if the state was properly organized.

IvanTheGreatDev:
I'd like to tackle this!

PR #473, opened 2018-10-08
Packs certain uint256s into uint64 to save gas

IvanTheGreatDev:
#461
This reduces uint256s to uint64s where it saves gas due to the tight packing feature of Solidity.

### DIFF
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -26,11 +26,11 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
     bytes32 public constant DISABLE_PAYMENTS_ROLE = keccak256("DISABLE_PAYMENTS_ROLE");
 
     uint64 public constant MAX_PAYMENTS_PER_TX = 20;
+    uint64 internal constant NO_PAYMENT = 0;
+    uint64 internal constant NO_TRANSACTION = 0;
+    uint64 internal constant MAX_UINT64 = uint64(-1);
 
     uint256 internal constant MAX_UINT = uint256(-1);
-    uint64 internal constant MAX_UINT64 = uint64(-1);
-    uint256 internal constant NO_PAYMENT = 0;
-    uint256 internal constant NO_TRANSACTION = 0;
 
     // Order optimized for storage
     struct Payment {
--- a/apps/survey/contracts/Survey.sol
+++ b/apps/survey/contracts/Survey.sol
@@ -19,8 +19,8 @@ contract Survey is AragonApp {
     bytes32 public constant CREATE_SURVEYS_ROLE = keccak256("CREATE_SURVEYS_ROLE");
     bytes32 public constant MODIFY_PARTICIPATION_ROLE = keccak256("MODIFY_PARTICIPATION_ROLE");
 
-    uint256 public constant PCT_BASE = 10 ** 18; // 0% = 0; 1% = 10^16; 100% = 10^18
-    uint256 public constant ABSTAIN_VOTE = 0;
+    uint64 public constant PCT_BASE = 10 ** 18; // 0% = 0; 1% = 10^16; 100% = 10^18
+    uint64 public constant ABSTAIN_VOTE = 0;
 
     struct OptionCast {
         uint256 optionId;
@@ -53,9 +53,9 @@ contract Survey is AragonApp {
         mapping (address => MultiOptionVote) votes;     // voter -> options voted, with its stakes
     }
 
+    uint64 public surveyTime;
     MiniMeToken public token;
     uint256 public minParticipationPct;
-    uint64 public surveyTime;
 
     // We are mimicing an array, we use a mapping instead to make app upgrade more graceful
     mapping (uint256 => SurveyStruct) internal surveys;
--- a/apps/voting/contracts/Voting.sol
+++ b/apps/voting/contracts/Voting.sol
@@ -21,29 +21,29 @@ contract Voting is IForwarder, AragonApp {
     bytes32 public constant MODIFY_SUPPORT_ROLE = keccak256("MODIFY_SUPPORT_ROLE");
     bytes32 public constant MODIFY_QUORUM_ROLE = keccak256("MODIFY_QUORUM_ROLE");
 
-    uint256 public constant PCT_BASE = 10 ** 18; // 0% = 0; 1% = 10^16; 100% = 10^18
+    uint64 public constant PCT_BASE = 10 ** 18; // 0% = 0; 1% = 10^16; 100% = 10^18
 
     enum VoterState { Absent, Yea, Nay }
 
     struct Vote {
         address creator;
+        bool executed;
         uint64 startDate;
-        uint256 snapshotBlock;
-        uint256 supportRequiredPct;
-        uint256 minAcceptQuorumPct;
+        uint64 snapshotBlock;
+        uint64 supportRequiredPct;
+        uint64 minAcceptQuorumPct;
         uint256 yea;
         uint256 nay;
         uint256 totalVoters;
         string metadata;
         bytes executionScript;
-        bool executed;
         mapping (address => VoterState) voters;
     }
 
-    MiniMeToken public token;
-    uint256 public supportRequiredPct;
-    uint256 public minAcceptQuorumPct;
+    uint64 public supportRequiredPct;
+    uint64 public minAcceptQuorumPct;
     uint64 public voteTime;
+    MiniMeToken public token;
 
     // We are mimicing an array, we use a mapping instead to make app upgrade more graceful
     mapping (uint256 => Vote) internal votes;
@@ -69,8 +69,8 @@ contract Voting is IForwarder, AragonApp {
     */
     function initialize(
         MiniMeToken _token,
-        uint256 _supportRequiredPct,
-        uint256 _minAcceptQuorumPct,
+        uint64 _supportRequiredPct,
+        uint64 _minAcceptQuorumPct,
         uint64 _voteTime
     )
         external
@@ -91,9 +91,9 @@ contract Voting is IForwarder, AragonApp {
     * @notice Change required support to `(_supportRequiredPct - _supportRequiredPct % 10^16) / 10^14`%
     * @param _supportRequiredPct New required support
     */
-    function changeSupportRequiredPct(uint256 _supportRequiredPct)
+    function changeSupportRequiredPct(uint64 _supportRequiredPct)
         external
-        authP(MODIFY_SUPPORT_ROLE, arr(_supportRequiredPct, supportRequiredPct))
+        authP(MODIFY_SUPPORT_ROLE, arr(uint256(_supportRequiredPct), uint256(supportRequiredPct)))
     {
         require(minAcceptQuorumPct <= _supportRequiredPct);
         require(_supportRequiredPct < PCT_BASE);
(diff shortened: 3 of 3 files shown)


IvanTheGreatDev:
I'm not sure how coverage is decreased, can someone assist? @sohkai sorry to bother

IvanTheGreatDev:
40 was used to fit the bool within the address,
address - 160 bits
timestamp - 48 bits
blockNumber - 40 bits
boolean - 8 bits

Would you like me to find a solution without the uint40?

IvanTheGreatDev:
Oh wait, apologies, `uint40` was a change I initially made but later reverted. I'll get that fixed now.

apps/voting/contracts/Voting.sol
```diff
@@ -21,29 +21,29 @@ contract Voting is IForwarder, AragonApp {
...
     struct Vote {
         address creator;
+        bool executed;
         uint64 startDate;
-        uint256 snapshotBlock;
-        uint256 supportRequiredPct;
-        uint256 minAcceptQuorumPct;
+        uint64 snapshotBlock;
+        uint64 supportRequiredPct;
+        uint64 minAcceptQuorumPct;
```
izqui: Worried about these being uint64 given that the max uint64 is `~1.8 * 10^19`and the max values they can have is `10^18`. We should at least explicitly cast `uint64` to `uint256` whenever used for math operations in `canExecute` and `isValuePct`.

Also extending these tests [here](https://github.com/aragon/aragon-apps/blob/ba4538d1e49cc36d411bc7bbd103c3348ed8c95d/apps/voting/test/voting.js#L151) so we also verify the math still works for bigger token amounts.

apps/finance/contracts/Finance.sol
```diff
@@ -26,11 +26,11 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
     bytes32 public constant DISABLE_PAYMENTS_ROLE = keccak256("DISABLE_PAYMENTS_ROLE");
 
     uint64 public constant MAX_PAYMENTS_PER_TX = 20;
+    uint64 internal constant NO_PAYMENT = 0;
```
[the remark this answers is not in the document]
IvanTheGreatDev: Good point, slipped my mind! I'll get on that!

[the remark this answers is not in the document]
IvanTheGreatDev: Although, I just realized, these are internal fields, does that still apply? 

I did some testing with 
``` 
pragma solidity ^0.4.0; 

contract Test { 
    uint64 constant internal thing = 43; 
    uint64 constant internal thing2 = 43;
    uint64 constant internal thing3 = 43; 
    uint64 constant internal thing4 = 43; 
} 
``` 
So it turns out changing the constants uint type has an extremely minimal effect on the gas, so because of this it's fine to make them `uint256`. 
However, the real culprit is the `public` constant, why does this need to be public? I don't see it being used anywhere. It's inclusion into the contract alone costs about 50K extra gas on deploy.

apps/voting/contracts/Voting.sol
IvanTheGreatDev: Sure thing, I went through and made sure they were explicitly converted before any math was performed.

I'm having trouble understanding how I would change those tests, they seem to test for 18 decimals already it seems. What should I change?

(sorry for the duplicate messages, I clicked the wrong comment button)

izqui: I would just add an extra decimal amount to test, so it is `[0, 2, 18, 26]` just to be extra careful

apps/finance/contracts/Finance.sol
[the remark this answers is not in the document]
izqui: With internal fields the impact is not that bad, I was actually testing the cost of using the getter generated by the `public constant`. I think `MAX_PAYMENTS_PER_TX` could be made internal as it is not used by the tests.

@sohkai what do you think? You made the majority of constants internal in #453 but not this one, was there a reason?

apps/voting/contracts/Voting.sol
IvanTheGreatDev: KK 👍

apps/finance/contracts/Finance.sol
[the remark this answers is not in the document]
sohkai: I thought it'd be useful information to be able to query for, since it's a configuration parameter that may change in the future.

I'm not opposed to making it internal though.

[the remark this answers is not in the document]
IvanTheGreatDev: If it were to be changed later we should remove `constant` and make a setter function for the field then no?

apps/voting/contracts/Voting.sol
```diff
@@ -199,7 +199,7 @@ contract Voting is IForwarder, AragonApp {
         }
 
         // Voting is already decided
-        if (_isValuePct(vote_.yea, vote_.totalVoters, vote_.supportRequiredPct)) {
+        if (_isValuePct(vote_.yea, vote_.totalVoters, uint256(vote_.supportRequiredPct))) {
```
sohkai: I know @izqui commented on this in https://github.com/aragon/aragon-apps/pull/473/files#r224013414, but across function calls the cast is redundant as the solidity compiler automatically casts it for us.

We never actually do any math involving `minAcceptQuorumPct` and `supportRequiredPct` outside of a comparison (which would also automatically be casted for us), and it should be a red flag if we ever did.

I wouldn't mind removing the explicit upwards casts @izqui.

apps/finance/contracts/Finance.sol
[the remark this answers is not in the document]
sohkai: Ah, I meant across different versions of the app. Maybe later on we decide in a new version of the contract to use `50` or something. But realistically there's very little a contract or client could do even knowing about this parameter, and it's something really easy to expose later.

apps/survey/contracts/Survey.sol
```diff
@@ -41,9 +41,9 @@ contract Survey is AragonApp {
     struct SurveyStruct {
         address creator;
         uint64 startDate;
-        uint256 options;
-        uint256 snapshotBlock;
-        uint256 minParticipationPct;
+        uint64 snapshotBlock;
+        uint64 minParticipationPct;
+        uint64 options;
```
sohkai: I think we should keep `options` to be `uint256` as `optionPower` still uses `uint256` for its mapping. In general, I think any "length"-type parameter should be kept to `uint256`.

apps/survey/contracts/Survey.sol
```diff
@@ -64,9 +64,9 @@ contract Survey is AragonApp {
     event StartSurvey(uint256 indexed surveyId);
     event CastVote(uint256 indexed surveyId, address indexed voter, uint256 option, uint256 stake, uint256 optionPower);
     event ResetVote(uint256 indexed surveyId, address indexed voter, uint256 option, uint256 previousStake, uint256 optionPower);
-    event ChangeMinParticipation(uint256 minParticipationPct);
+    event ChangeMinParticipation(uint64 minParticipationPct);
 
-    modifier acceptableMinParticipationPct(uint256 _minParticipationPct) {
+    modifier acceptableMinParticipationPct(uint64 _minParticipationPct) {
         require(_minParticipationPct > 0 && _minParticipationPct <= PCT_BASE);
```
sohkai: We should also change `PCT_BASE` to be `uint64`, like in Voting.

apps/survey/contracts/Survey.sol
```diff
@@ -253,11 +253,11 @@ contract Survey is AragonApp {
...
             uint64 _startDate,
-            uint256 _snapshotBlock,
-            uint256 _minParticipationPct,
+            uint64 _snapshotBlock,
+            uint64 _minParticipationPct,
+            uint64 _options,
             uint256 _votingPower,
-            uint256 _participation,
-            uint256 _options
+            uint256 _participation
```
sohkai: Let's revert the return order; they're already not in sync with the storage struct anyway, and changing it means we'd have to change it in the frontend as well.

sohkai:
Just to resurface https://github.com/aragon/aragon-apps/pull/473#discussion_r224439935 (in case we need to discuss):

Across function calls the upwards casts are redundant as the solidity compiler automatically adds it for us.

We never actually do any math involving `minAcceptQuorumPct` and `supportRequiredPct` outside of a comparison (which would also automatically be casted for us), and it should be a red flag if we ever did.

I wouldn't mind removing the explicit upwards casts @izqui.

apps/survey/contracts/Survey.sol
IvanTheGreatDev: Apologies, not sure why I made that change.

apps/survey/contracts/Survey.sol
IvanTheGreatDev: Fair point, I'll revert.

IvanTheGreatDev:
Let me know if there are any other changes needed! 😄

apps/survey/test/survey.js
```diff
@@ -1,347 +1,557 @@
...
+  let APP_MANAGER_ROLE, CREATE_SURVEYS_ROLE, MODIFY_PARTICIPATION_ROLE;
+
+  const surveyTime = 1000;
+  const root = accounts[0];
+  const holder19 = accounts[0];
+  const holder31 = accounts[1];
+  const holder50 = accounts[2];
+  const nonHolder = accounts[4];
+  const NULL_ADDRESS = "0x00";
+  const minimumAcceptanceParticipationPct = pct16(20);
```
bingen: Why are those all these changes? I don't have a strong opinion about format issues, but we don't usually use trailing semicolon for JS.

IvanTheGreatDev: Ah crap that must be my linter, I'll revert them.

apps/finance/contracts/Finance.sol
```diff
@@ -25,12 +25,11 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
     bytes32 public constant EXECUTE_PAYMENTS_ROLE = keccak256("EXECUTE_PAYMENTS_ROLE");
     bytes32 public constant DISABLE_PAYMENTS_ROLE = keccak256("DISABLE_PAYMENTS_ROLE");
 
-    uint64 public constant MAX_PAYMENTS_PER_TX = 20;
-
-    uint256 internal constant MAX_UINT = uint256(-1);
-    uint64 internal constant MAX_UINT64 = uint64(-1);
     uint256 internal constant NO_PAYMENT = 0;
     uint256 internal constant NO_TRANSACTION = 0;
+    uint256 internal constant MAX_PAYMENTS_PER_TX = 20;
```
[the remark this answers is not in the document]
IvanTheGreatDev: It doesn't really matter and comparisons are automatically casted down. Should we stick with uint256 for internal constants?

apps/survey/contracts/Survey.sol
```diff
@@ -19,8 +19,8 @@ contract Survey is AragonApp {
     bytes32 public constant CREATE_SURVEYS_ROLE = keccak256("CREATE_SURVEYS_ROLE");
     bytes32 public constant MODIFY_PARTICIPATION_ROLE = keccak256("MODIFY_PARTICIPATION_ROLE");
 
-    uint256 public constant PCT_BASE = 10 ** 18; // 0% = 0; 1% = 10^16; 100% = 10^18
-    uint256 public constant ABSTAIN_VOTE = 0;
+    uint64 public constant PCT_BASE = 10 ** 18; // 0% = 0; 1% = 10^16; 100% = 10^18
+    uint64 public constant ABSTAIN_VOTE = 0;
```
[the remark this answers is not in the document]
IvanTheGreatDev: Certainly does but these are being checked inside of the test file I'm pretty sure, hence why I kept them public.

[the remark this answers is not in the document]
IvanTheGreatDev: Although it seems it's cheaper to keep them as uint256 so I'll revert that.

IvanTheGreatDev:
Alright, issues resolved 👍

chris-remus:
@izqui to review once more, expecting to merge after that.

apps/survey/contracts/Survey.sol
```diff
@@ -127,7 +127,7 @@ contract Survey is AragonApp {
     * @notice Create a new non-binding survey about "`_metadata`"
     * @param _metadata Survey metadata
     * @param _options Number of options voters can decide between
-    * @return surveyId id for newly created survey
+    * x@return surveyId id for newly created survey
```
sohkai: Small typo here, ~~x~~

izqui:
```suggestion
    * @return surveyId id for newly created survey
```

izqui:
Thanks for your patience, @IvanTheGreatDev! Finally merged 🎉 :)

Issue #374, opened 2018-06-13
Finance: rename setPaymentDisabled


sohkai:
`setPaymentDisabled()` can also enable the payment, which is confusing

izqui:
Yes @IvanTheGreatDev. Closed by #474!

PR #474, opened 2018-10-09
Rename setPaymentDisabled to setPaymentStatus

IvanTheGreatDev:
Also changes constant to view app-wide and renamed app-wide from disabled to inactive for the payment status.

#374

### DIFF
--- a/apps/finance/artifact.json
+++ b/apps/finance/artifact.json
@@ -206,7 +206,7 @@
 					"type": "string"
 				},
 				{
-					"name": "disabled",
+					"name": "inactive",
 					"type": "bool"
 				},
 				{
@@ -683,11 +683,11 @@
 					"type": "uint256"
 				},
 				{
-					"name": "_disabled",
+					"name": "_active",
 					"type": "bool"
 				}
 			],
-			"name": "setPaymentDisabled",
+			"name": "setPaymentStatus",
 			"outputs": [],
 			"payable": false,
 			"stateMutability": "nonpayable",
@@ -890,7 +890,7 @@
 				},
 				{
 					"indexed": false,
-					"name": "disabled",
+					"name": "inactive",
 					"type": "bool"
 				}
 			],
@@ -984,11 +984,11 @@
 			"notice": "Execute pending payment #`_paymentId`"
 		},
 		{
-			"sig": "setPaymentDisabled(uint256,bool)",
+			"sig": "setPaymentStatus(uint256,bool)",
 			"roles": [
 				"DISABLE_PAYMENTS_ROLE"
 			],
-			"notice": "`_disabled ? 'Disable' : 'Enable'` payment `_paymentId`"
+			"notice": "`_active ? 'Active' : 'Inactive'` payment `_paymentId`"
 		},
 		{
 			"sig": "depositToVault(address)",
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -37,7 +37,7 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
         address token;
         address receiver;
         address createdBy;
-        bool disabled;
+        bool inactive;
         uint256 amount;
         uint64 initialPaymentTime;
         uint64 interval;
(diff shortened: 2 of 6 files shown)


apps/finance/artifact.json
```diff
@@ -984,11 +984,11 @@
 			"notice": "Execute pending payment #`_paymentId`"
 		},
 		{
-			"sig": "setPaymentDisabled(uint256,bool)",
+			"sig": "setPaymentStatus(uint256,bool)",
 			"roles": [
 				"DISABLE_PAYMENTS_ROLE"
 			],
-			"notice": "`_disabled ? 'Disable' : 'Enable'` payment `_paymentId`"
+			"notice": "`_active ? 'Active' : 'Inactive'` payment `_paymentId`"
```
sohkai: @izqui I'm thinking we should just remove this file from the repo for now, and add it back when we're getting closer to deployment. We'll have to regen it anyway at that point.

apps/finance/contracts/Finance.sol
```diff
@@ -319,17 +319,17 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
...
-    * @notice `_disabled ? 'Disable' : 'Enable'` payment `_paymentId`
+    * @notice `_active ? 'Active' : 'Inactive'` payment `_paymentId`
     * @param _paymentId Identifier for payment
-    * @param _disabled Whether it will be disabled or enabled
+    * @param _active Whether it will be active or inactive
     */
-    function setPaymentDisabled(uint256 _paymentId, bool _disabled)
+    function setPaymentStatus(uint256 _paymentId, bool _active)
         external
         authP(DISABLE_PAYMENTS_ROLE, arr(_paymentId))
```
sohkai: We should also rename this role, so it's less confusing. It can actually enable or disable existing payments, so I'd rename it to `MANAGE_PAYMENTS_ROLE`.

shared/minime/contracts/MiniMeToken.sol
```diff
@@ -192,7 +192,7 @@ contract MiniMeToken is Controlled {
         require((_to != 0) && (_to != address(this)));
         // If the amount being transfered is more than the balance of the
         //  account the transfer returns false
-        var previousBalanceFrom = balanceOfAt(_from, block.number);
+        uint256 previousBalanceFrom = balanceOfAt(_from, block.number);
```
sohkai: Could you revert the changes to the `MiniMeToken` here?

I'm not opposed to fixing the compiler warnings here, but we'd also like it to stay as closely as possible to the original version and we should have a separate discussion if we should just [update to `MMT_0.2`](https://github.com/giveth/minime/blob/master/contracts/MiniMeToken.sol).

apps/finance/test/finance.js
```diff
@@ -667,8 +667,8 @@ contract('Finance App', accounts => {
                 })
             })
 
-            it('fails executing disabled payment', async () => {
-                await finance.setPaymentDisabled(1, true)
+            it('fails executing inactive payment', async () => {
+                await finance.setPaymentStatus(1, false)
```
sohkai: We should also add a test to re-enable a payment and see that it can be executed.

shared/minime/contracts/MiniMeToken.sol
IvanTheGreatDev: Sure thing!

apps/finance/test/finance.js
IvanTheGreatDev: Good idea, I'll add that!

apps/finance/artifact.json
IvanTheGreatDev: Hmm, I don't know if there's a difference between the two. I kind of feel like enable/disable is better.

IvanTheGreatDev: I'd say active/inactive makes more sense if we choose `status` for the name of the function, because saying a payments status is `disable` is weird, but `inactive` sounds totally fine.

sohkai: Agreed, thought about it some more and it makes more sense if you can re-enable recurring payments ("active recurring payment" sounds a lot better than "enabled recurring payment").

sohkai:
Just to re-surface this from https://github.com/aragon/aragon-apps/pull/474#discussion_r224405146:

> Although @izqui we should probably think of a way to completely disable a payment at some point (or remove the ability to re-enable payments).

IvanTheGreatDev:
Should be good to go now! 😄

IvanTheGreatDev:
What is this PR waiting on?

izqui:
Just added 51539caa7c1c4ee07584452ae51191b010f018a3 passing whether we are setting the payment as active as a parameter to the role

PR #477, opened 2018-10-10
Fix arapp roles

sohkai:
Oops 😅.

Leads me to really think we should add some tooling in aragon-cli to parse contract comments so we can make this more automated.


izqui:
Agreed, or at least detect when the length of params to the role doesn't match with the arapp.

PR #479, opened 2018-10-11
Payroll App - Milestone 1

sistemico:
We proposed to separate the development of Payroll app in milestones defined by working features comprised of both UI and Contracts. This pull request is the initial milestone including project setup, migration to aragonOS v4 and the functionality needed to add new employees.

## Deliverables

- [x] Migration to Aragon CLI v5.0.0: the current partial implementation of this app includes old scripts and tools that since have been integrated into latest version of aragon-cli and simplify development.

- [x]  App layout and some UI components: This includes the initial UI app setup and basic layout with tabs for “My payroll” and “Team payroll”.
  - [x] Separate package.json for UI
  - [x] Migrate build to Parcel 1.10.x and Babel 7
  - [x] Reusable components (could be moved to Aragon UI as a contribution in the future)

- [x] Migration of contracts to aragonOS v4

- [x] Payroll Kit

- [x] Feature: Add employee

  - An initial implementation of this features as described and designed in https://github.com/aragon/design/issues/2 with the limitation that it will not be integrated with an Identity app which means that some fields like account address, name and role will use mock data.

  - This feature will include a working UI following this mockup to add employees integrated with the corresponding contract.

## Related work
This branch also includes the following PRs:
- https://github.com/aragon/aragon-apps/pull/361
- https://github.com/aragon/aragon-apps/pull/383
- https://github.com/aragon/aragon-apps/pull/394


future-apps/payroll/package.json
```diff
@@ -1,39 +1,45 @@
...
-    "@aragon/os": "^3.0.6"
+    "@aragon/apps-finance": "^2.0.0-beta.2",
+    "@aragon/os": "^4.0.0-beta.3",
+    "@aragon/ppf-contracts": "^1.0.2"
+  },
+  "devDependencies": {
+    "@aragon/apps-shared-migrations": "^1.0.0",
+    "@aragon/apps-shared-minime": "^1.0.0",
+    "@aragon/apps-vault": "^3.0.0-beta.2",
+    "@aragon/cli": "git+https://github.com/aragon/aragon-cli.git#06e2f1cee4384549441d7ec8807b7635dbd64b69",
```
sistemico: We had to use the latest version from repo because the version published in npm is broken.

sistemico: Aragon CLI was updated to the just-released version (v5.1.0)

sistemico:
Updated `@aragon/cli` and `@aragon/client` package to the final versions released yesterday. Those were just package versioning adjustments, no functionality was altered. This pull request continues to be ready for review.

sistemico:
@sohkai Why are `package-lock.json` files not versioned in`aragon-apps`? Is there any particular reason? Should we exclude those files from this PR?

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -1,36 +1,48 @@
-pragma solidity 0.4.18;
+pragma solidity 0.4.24;
 
 import "@aragon/os/contracts/apps/AragonApp.sol";
-import "@aragon/os/contracts/common/Initializable.sol";
-import "@aragon/os/contracts/common/IForwarder.sol";
+// import "@aragon/os/contracts/common/IForwarder.sol";
```
Quazia: I understand the need to have this workaround now for coverage but make sure to note and re-implement this ASAP.

Quazia: I understand the need to have this workaround now for coverage but make sure to note and re-implement this ASAP.

future-apps/payroll/app/src/panels/AddEmployee.js
```diff
@@ -0,0 +1,206 @@
...
+            <Input.Entity
+              ref={el => this.entitySearch = el}
+              key={entity && entity.domain}
+              value={entity && entity.domain}
+              onChange={this.handleEntityChange}
+            />
+          </Field>
+
+          <Field label='Salary'>
+            <Input.Currency
```
Quazia: I'm not seeing any [icons](https://ui.aragon.org/icons/) currently for this, but in the [design ](https://github.com/aragon/design/issues/2) there's a '$' icon in this field.

future-apps/payroll/app/src/panels/AddEmployee.js
```diff
@@ -0,0 +1,206 @@
...
+              innerRef={el => this.salaryInput = el}
+              value={salary || ''}
+              onChange={this.handleSalaryChange}
+            />
+          </Field>
+
+          <Field label='Start Date'>
+            <Input.Date
```
Quazia: Same as with `Icon.Currency` on with calendar icon. I'm not sure what the best approach here is, I guess either:

1. Make a calendar and $ icon for these fields or,
2. Use a placeholder for now and wait until icons are added to the aragon-ui repo

future-apps/payroll/app/src/panels/AddEmployee.js
sistemico: That feature will be implemented in next milestone

future-apps/payroll/app/src/panels/AddEmployee.js
sistemico: That feature will be implemented in next milestone

Quazia:
All pending issues are being addressed in the next milestone. @sohkai can I merge this as the team is currently close to submitting a PR that builds on changes in this PR.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -112,119 +147,158 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
      */
     function setEmployeeSalary(
         uint128 employeeId,
-        uint256 yearlyDenominationSalary
+        uint256 denominationSalary
     )
+        employeeExists(employeeId)
         external
-        authP(ADD_EMPLOYEE_ROLE, arr(yearlyDenominationSalary, 0))
+        authP(SET_EMPLOYEE_SALARY_ROLE, arr(uint256(employeeId), denominationSalary))
```
bingen: I don't think we need to cast `employeeId` here, but in case we do, I would also cast `uint64` dates in `arr` around.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -250,121 +324,159 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
         // check it's non-null address
         require(newAddress != address(0));
-        // check that employee exists (and matches)
+
         uint128 employeeId = employeeIds[msg.sender];
         Employee storage employee = employees[employeeId];
-        // check that employee exists (and matches)
-        require(employee.accountAddress == msg.sender);
 
+        emit ChangeAddressByEmployee(employeeId, employee.accountAddress, newAddress);
```
bingen: I guess I would move this down, to follow the same pattern as other functions.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -250,121 +324,159 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
+     * @return Employee's accrued value
      * @return Employee's name
-     * @return Employee's last call to payment distribution date
      * @return Employee's last payment received date
+     * @return Employee's termination date (max uint64 if none)
+     * @return Bool indicating if employee is terminated
      */
     function getEmployee(uint128 employeeId)
-        external
+        public
```
bingen: I don't remember why this change. `getEmployeeByAddress` is kept as `external`.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -386,10 +498,11 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
 
     function _addEmployee(
         address accountAddress,
-        uint256 initialYearlyDenominationSalary,
+        uint256 initialDenominationSalary,
         string name,
-        uint256 startDate
+        uint64 startDate
     )
+        isInitialized
```
bingen: This is not needed, as external methods calling this one are already protected.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -398,42 +511,82 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
     }
 
+    function _addAccruedValue(uint128 employeeId, uint256 amount) internal {
+        employees[employeeId].accruedValue = employees[employeeId].accruedValue.add(amount);
+
+        emit AddEmployeeAccruedValue(employeeId, amount);
+    }
+
+    function _setPriceFeed(IFeed _feed) internal {
+        require(_feed != address(0));
```
bingen: We could use `isContract` here.

future-apps/payroll/contracts/PayrollKit.sol
```diff
@@ -0,0 +1,144 @@
...
+
+
+contract PPFMock is IFeed {
+  function get(address base, address quote) external view returns (uint128 xrt, uint64 when) {
+      xrt = 1;
+      when = uint64(now);
+  }
+}
+
+contract KitBase is APMNamehash, EVMScriptRegistryConstants {
```
bingen: I would reuse `KitBase` from dao-kits repo.

future-apps/payroll/contracts/PayrollKit.sol
```diff
@@ -0,0 +1,144 @@
...
+        acl.grantPermission(root, app, permission);
+        acl.revokePermission(this, app, permission);
+        acl.setPermissionManager(root, app, permission);
+    }
+}
+
+contract PayrollKit is KitBase {
+    MiniMeTokenFactory tokenFactory;
+
+    uint64 financePeriodDuration = 31557600;
```
bingen: A comment "translating" this value to a more human readable one would be nice.

future-apps/payroll/contracts/PayrollKit.sol
```diff
@@ -0,0 +1,144 @@
...
+    function newInstance() public returns (Kernel dao, Payroll payroll) {
+        address root = msg.sender;
+        address employer = msg.sender;
+
+        dao = fac.newDAO(this);
+        ACL acl = ACL(dao.acl());
+
+        PPFMock priceFeed = new PPFMock();
+
+        MiniMeToken denominationToken = tokenFactory.createCloneToken(MiniMeToken(0), 0, "US Dollar", 18, "USD", true);
```
bingen: I would allow to pass these params ("USD", etc) dynamically. Or at least define them as constants at the beginning of the contract.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -250,121 +324,159 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
+        emit DetermineAllocation(employeeIds[msg.sender], msg.sender);
     }
 
     /**
      * @dev To withdraw payment by employee (the caller). The amount owed since last call will be transferred.
      * @notice To withdraw payment by employee (the caller). The amount owed since last call will be transferred.
      */
-    function payday() external {
+    function payday() isInitialized employeeMatches external {
         Employee storage employee = employees[employeeIds[msg.sender]];
```
bingen: This is not used.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -250,121 +324,159 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
+
+        emit DetermineAllocation(employeeIds[msg.sender], msg.sender);
     }
 
     /**
      * @dev To withdraw payment by employee (the caller). The amount owed since last call will be transferred.
      * @notice To withdraw payment by employee (the caller). The amount owed since last call will be transferred.
      */
-    function payday() external {
+    function payday() isInitialized employeeMatches external {
```
bingen: I think `isInitialized` is not needed here, as if `employeeMatches` is passing it means there's an element in `employees` with `accountAddress` non-zero, with can only be done after adding an employee and therefore having had the contract initialized. I would remove it and add a comment explaining it. Same applies to other functions with both modifiers.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -443,24 +596,62 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
+        return employees[employeeId].denominationTokenSalary.mul(time);
     }
 
     /**
-     * @dev Check that a token has the exchange rate already set
-     *      Internal function, needed to ensure that we have the rate before making a payment.
+     * @dev Check that a token gets a correct exchange rate.
+     *      Internal function, needed to ensure that we have a recent rate before making a payment.
      * @param token The token address
      * @return True if we have the exchange rate, false otherwise
```
sohkai: The `@return` is wrong: if we have the exchange rate, we return `ONE`, and `0` otherwise.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -443,24 +596,62 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
+            return ONE;
         }
-        return true;
-    }
 
-    function getTimestamp() internal view returns (uint256) { return now; }
+        (xrt, when) = feed.get(denominationToken, token);
 
+        // check it's recent enough
+        if (when < getTimestamp64().sub(rateExpiryTime)) {
```
sohkai: It might just be me, but this calculation seems more clear as `getTimestamp64().sub(when) >= rateExpiryTime`.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -443,24 +596,62 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
-     *      Internal function, needed to ensure that we have the rate before making a payment.
+     * @dev Check that a token gets a correct exchange rate.
+     *      Internal function, needed to ensure that we have a recent rate before making a payment.
      * @param token The token address
      * @return True if we have the exchange rate, false otherwise
      */
-    function checkExchangeRate(address token) internal view returns (bool) {
-        if (exchangeRates[token] == 0) {
-            return false;
+    function getExchangeRate(address token) internal view returns (uint128) {
```
sohkai: Should include the `_` prefix since it's `internal`.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -398,42 +511,82 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
      * @return True if something has been paid
      */
-    function _payTokens(uint128 employeeId, uint256 time) internal returns (bool somethingPaid) {
+    function _payTokens(uint128 employeeId) internal returns (bool somethingPaid) {
         Employee storage employee = employees[employeeId];
+
+        // get the min of current date and termination date
+        uint64 timestamp = getTimestamp64();
+        uint64 toDate;
+        if (employee.terminated && employee.endDate < timestamp) {
```
sohkai: I think this can just be `employee.endDate < timestamp`, since active employees have their `endDates` as `MAX_UINT64`.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -443,24 +596,62 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
                 1,
                 ""
```
sohkai: @bingen I wonder if we should add a generic reference, like "Payroll"?

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -443,24 +596,62 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
             );
-            SendPayroll(employee.accountAddress, token, tokenAmount);
+            emit SendPayroll(employee.accountAddress, token, tokenAmount);
             somethingPaid = true;
         }
+
+        // try to remove employee
```
sohkai: @bingen was there a reason to remove terminated employees, rather than leaving them? Gas refund?

future-apps/payroll/package.json
```diff
@@ -1,39 +1,45 @@
...
+    "start": "aragon run --network rpc --kit PayrollKit --kit-init @ARAGON_ENS",
+    "test": "TRUFFLE_TEST=true ./node_modules/@aragon/os/scripts/ganache-cli.sh",
+    "test:gas": "GAS_REPORTER=true npm test"
   },
   "dependencies": {
-    "@aragon/apps-finance": "^1.0.0",
-    "@aragon/os": "^3.0.6"
+    "@aragon/apps-finance": "^2.0.0-beta.2",
+    "@aragon/os": "^4.0.0-beta.3",
+    "@aragon/ppf-contracts": "^1.0.2"
```
sohkai: We should update these to their latest, non-beta versions, and pin them. Doing so may cause the tests to fail, but it's very like due to the Finance app having a role's name changed (`DISABLE` => `MANAGE`) and `kernel.newAppInstance()` defaulting to the 4-argument overload.

future-apps/payroll/contracts/Payroll.sol
lmcorbalan: Would you do the same with `isTokenAllowed`?

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -235,7 +309,7 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
             but in case it happens, this function allows to recover them.
      * @notice Allows to send tokens from this contract to Finance, to avoid locked tokens in contract forever
      */
-    function depositToFinance(address token) external {
+    function depositToFinance(address token) isInitialized external {
```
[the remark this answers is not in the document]
lmcorbalan: @sohkai @bingen 
Should we remove `depositToFinance` or change it to?: 😅 
```
function depositToFinance(address _token) public isInitialized {
    ERC20 tokenContract = ERC20(_token);
    uint256 amount = _token == ETH ? address(this).balance : tokenContract.balanceOf(this);
    require(amount > 0, ERROR_RECOVER_AMOUNT_ZERO);
    
    tokenContract.approve(address(finance), value);
    finance.deposit(tokenContract,  amount,  "Adding Funds");
}
```

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -1,36 +1,48 @@
...
         address accountAddress; // unique, but can be changed over time
         mapping(address => uint8) allocation;
-        uint256 denominationTokenSalary; // per second
-        uint256 lastPayroll;
+        uint256 denominationTokenSalary; // per second in denomination Token
+        uint256 accruedValue;
+        uint64 lastPayroll;
+        uint64 endDate;
+        bool terminated;
         string name;
```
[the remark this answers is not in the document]
lmcorbalan: @sohkai @bingen what do you think about changing name to ens when adding an employee and in the event?
We think it'll be more useful than name, like we could use it for retrieving employee info from identity app

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -112,119 +147,158 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
-     * @dev Add employee to Payroll
-     * @notice Add employee to Payroll. See addEmployeeWithNameAndStartDate
+     * @dev Add employee to Payroll. See addEmployeeWithNameAndStartDate
+     * @notice Add employee `name` with address `accountAddress` to Payroll with a salary of `initialDenominationSalary` per second.
      * @param accountAddress Employee's address to receive Payroll
-     * @param initialYearlyDenominationSalary Employee's salary
+     * @param initialDenominationSalary Employee's salary, per second in denomination Token
      * @param name Employee's name
      */
     function addEmployeeWithName(
```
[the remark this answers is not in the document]
bingen: But what if an employee doesn't have an ENS name?

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -443,24 +596,62 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
+
+        employees[employeeId].terminated = true;
+        employees[employeeId].endDate = endDate;
+
+        emit TerminateEmployee(employeeId, employees[employeeId].accountAddress, endDate);
+    }
+
+    function _getOwedSalary(uint128 employeeId, uint64 date) internal view returns (uint256) {
+        // get time that has gone by (seconds)
+        uint64 time = date.sub(employees[employeeId].lastPayroll);
```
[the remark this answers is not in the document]
bingen: Good catch!

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -112,119 +147,158 @@ contract Payroll is AragonApp { //, IForwarder { // makes coverage crash (remove
...
+     */
+    function addAccruedValue(
+        uint128 employeeId,
+        uint256 amount
+    )
+        employeeExists(employeeId)
+        external
+        authP(ADD_ACCRUED_VALUE_ROLE, arr(uint256(employeeId), amount))
+    {
+        require(amount <= MAX_ACCRUED_VALUE);
```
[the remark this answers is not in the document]
bingen: I don't think so: if I recall correctly, the point of this check was to avoid the employer "blocking" the employee setting an accrued value so high that would make payroll overflow. But if an employee is owed more than `2^128`, it should be allowed to be paid.

future-apps/payroll/contracts/Payroll.sol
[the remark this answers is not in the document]
sohkai: @lmcorbalan I think resolving ENS should be a front-end aspect by pulling from the recipient address; the name provided here is really metadata for the employee (I think of it as their "de facto" name in the context of the organization, since an employee can change their recipient address).

future-apps/payroll/contracts/Payroll.sol
sohkai: Hmm... re-enabling could be interesting. The only complication I could see with it is dealing with accrued balances during re-enabling.

future-apps/payroll/app/src/components/Tab/index.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
```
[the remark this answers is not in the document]
lmcorbalan: We'll address in with https://github.com/protofire/aragon-apps/issues/62

PR #480, opened 2018-10-12
Token manager updates

bpierre:


apps/token-manager/app/src/App.js
```diff
@@ -41,31 +42,45 @@ class App extends React.Component {
...
-    const toMint = new BN(`${amountConverted}`, 10)
-    app.mint(recipient, toMint.toString())
+    const amountRounded = Math.floor(parseFloat(amount) * tokenDecimalsBase)
+    const amountValue = new BN(`${amountRounded}`, 10).toString()
+
+    if (mode === 'assign') {
+      app.mint(holder, amountValue)
+    }
+    if (mode === 'remove') {
+      app.burn(holder, amountValue)
```
bpierre: @sohkai do you know why this doesn’t work? `BURN_ROLE` doesn’t appear in the `artifact.json`.

sohkai: We completely forgot it before on 0.5. The current Rinkeby instances would need a redeployment with a re-gened `artifact.json` now that https://github.com/aragon/aragon-apps/pull/477 is merged.

Let's just make sure this works on the new DAOs :).

PR #483, opened 2018-10-13
Voting: refactor state and better enforce bignumber purity

sohkai:
Depends on #483.

Refactors the internal state model of the app.

Main changes:

- Moved most of the vote data structure manipulation into the `observe` block, as they don't need to be done on each render, and this helps ensure type purity as it's the "entry point" for data
- Moved everything related to a vote's contract state into `vote.data`
- Added a settings context provider (old API) for cases where we need some of the voting app's settings
- Except for time-based values, avoid passing back numbers from the script (prefer strings)
- Add `numData` fields to both the settings and each vote, for calculations that don't require accurate operations
- Use exact calculations when they're safer (e.g. when mimicking [`canExecute()`](https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol#L194)

### DIFF
--- a/apps/voting/app/src/App.js
+++ b/apps/voting/app/src/App.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
-import { isBefore } from 'date-fns/esm'
 import { AragonApp, AppBar, Button, SidePanel, observe } from '@aragon/ui'
+import BN from 'bn.js'
 import EmptyState from './screens/EmptyState'
 import Votes from './screens/Votes'
 import tokenAbi from './abi/token-balanceOfAt.json'
@@ -9,12 +9,12 @@ import VotePanelContent from './components/VotePanelContent'
 import NewVotePanelContent from './components/NewVotePanelContent'
 import AppLayout from './components/AppLayout'
 import { networkContextType } from './utils/provideNetwork'
-import { safeDiv } from './math-utils'
+import { settingsContextType } from './utils/provideSettings'
 import { hasLoadedVoteSettings } from './vote-settings'
 import { VOTE_YEA } from './vote-types'
 import {
   EMPTY_CALLSCRIPT,
-  getQuorumProgress,
+  isVoteOpen,
   voteTypeFromContractEnum,
 } from './vote-utils'
 
@@ -27,20 +27,23 @@ class App extends React.Component {
       etherscanBaseUrl: 'https://rinkeby.etherscan.io',
       name: 'rinkeby',
     },
-    pctBase: -1,
-    tokenAddress: null,
-    tokenDecimals: 18,
+    settingsLoaded: false,
     tokenSymbol: '',
-    supportRequiredPct: -1,
     userAccount: '',
     votes: [],
-    voteTime: -1,
   }
   static childContextTypes = {
     network: networkContextType,
+    settings: settingsContextType,
   }
   getChildContext() {
-    return { network: this.props.network }
+    return {
+      network: this.props.network,
+      settings: {
+        pctBase: this.props.pctBase,
+        voteTime: this.props.voteTime,
+      },
+    }
   }
 
   constructor(props) {
@@ -49,7 +52,6 @@ class App extends React.Component {
     this.state = {
       createVoteVisible: false,
       currentVoteId: -1,
-      settingsLoaded: false,
       tokenContract: this.getTokenContract(props.tokenAddress),
       voteVisible: false,
       voteSidebarOpened: false,
@@ -57,15 +59,6 @@ class App extends React.Component {
     }
   }
   componentWillReceiveProps(nextProps) {
-    const { settingsLoaded } = this.state
-
-    // Is this the first time we've loaded the settings?
-    if (!settingsLoaded && hasLoadedVoteSettings(nextProps)) {
-      this.setState({
-        settingsLoaded: true,
-      })
-    }
-
     // Refresh the token contract if its address changes
     if (nextProps.tokenAddress !== this.props.tokenAddress) {
       this.setState({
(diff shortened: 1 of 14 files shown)


apps/voting/app/src/App.js
```diff
@@ -146,45 +139,37 @@ class App extends React.Component {
...
-      settingsLoaded,
       tokenContract,
       voteSidebarOpened,
       voteVisible,
       userAccountVotes,
     } = this.state
 
-    const displayVotes = settingsLoaded && votes.length > 0
-    const supportRequired = settingsLoaded ? supportRequiredPct / pctBase : -1
+    const now = new Date()
```
sohkai: Something to solve later: I was getting spooked we weren't re-rendering enough, to check for cases where a vote goes from open to closed when the app is open.

apps/voting/app/src/App.js
```diff
@@ -260,7 +243,65 @@ class App extends React.Component {
...
+                data: {
+                  ...data,
+                  endDate: new Date(data.startDate + voteTime),
+                  minAcceptQuorum: new BN(data.minAcceptQuorum),
+                  nay: new BN(data.nay),
+                  totalVoters: new BN(data.totalVoters),
+                  yea: new BN(data.yea),
+                  supportRequiredPct: new BN(supportRequiredPct),
+                },
+                numData: {
```
sohkai: Not the nicest way of doing it... but saves us from having to convert back and forth since the majority of the calculations don't need the accuracy of BN.js.

apps/voting/app/src/components/SummaryRows.js
```diff
@@ -11,7 +12,7 @@ const SummaryRows = ({ yea, nay, symbol }) => (
         <div>{yea.pct}%</div>
       </RowStart>
       <Amount>
-        {yea.amount} {symbol}
+        {formatNumber(yea.amount, 5)} {symbol}
```
sohkai: Right now this always forces 5 decimal points. Given that most people will be using whole units for tokens, at least initially, this could look odd (that all the decimals are always `.00000`):



------------

Wonder if we should not display the decimals if it's a whole number?

apps/voting/app/src/vote-settings.js
```diff
@@ -1,8 +1,8 @@
 const voteSettings = [
   ['token', 'tokenAddress'],
   ['voteTime', 'voteTime', 'time'],
-  ['PCT_BASE', 'pctBase', 'number'],
-  ['supportRequiredPct', 'supportRequiredPct', 'number'],
+  ['PCT_BASE', 'pctBase', 'bignumber'],
+  ['supportRequiredPct', 'supportRequiredPct', 'bignumber'],
```
sohkai: We don't do any parsing of `'number'` declarations anymore, but I thought it'd be fitting to assign `'bignumber'` for numbers that are expected to be passed back as a string.

apps/voting/app/src/App.js
```diff
@@ -207,7 +192,7 @@ class App extends React.Component {
           </AppLayout.Header>
           <AppLayout.ScrollWrapper>
             <AppLayout.Content>
-              {displayVotes ? (
+              {appReady ? (
```
sohkai: I was thinking it would also be nice to disable the top bar and empty state card's action if the app's not ready yet (requires https://github.com/aragon/aragon-ui/pull/219).

apps/voting/app/src/App.js
```diff
@@ -261,6 +243,66 @@ class App extends React.Component {
...
+                  data: {
+                    ...data,
+                    endDate: new Date(data.startDate + voteTime),
+                    minAcceptQuorum: new BN(data.minAcceptQuorum),
+                    nay: new BN(data.nay),
+                    totalVoters: new BN(data.totalVoters),
+                    yea: new BN(data.yea),
+                    supportRequiredPct: new BN(supportRequiredPct),
+                  },
+                  numData: {
```
sohkai: Not the nicest... but saves us from having to convert back and forth since the majority of the calculations don't need the accuracy of BN.js.

apps/voting/app/src/App.js
```diff
@@ -207,7 +192,7 @@ class App extends React.Component {
           </AppLayout.Header>
           <AppLayout.ScrollWrapper>
             <AppLayout.Content>
-              {displayVotes ? (
+              {appStateReady && votes.length > 0 ? (
```
sohkai: I was thinking it would also be nice to disable the top bar and empty state card's action if the app's not ready yet (requires https://github.com/aragon/aragon-ui/pull/219).

apps/voting/app/src/components/SummaryRows.js
bpierre: Yes maybe we could check that both are whole numbers and remove the decimals if that’s the case?

apps/voting/app/src/vote-settings.js
bpierre: I’m not sure it’s related, but I get some values like `pctBase` and `supportRequiredPct` as `Number` on a testing DAO (`p.aragonid.eth`), making BN.js throw an error when being instantiated.

We could make it work by converting the values:

```diff
diff --git a/apps/voting/app/src/App.js b/apps/voting/app/src/App.js
index bd2df99..0a56c58 100644
--- a/apps/voting/app/src/App.js
+++ b/apps/voting/app/src/App.js
@@ -264,8 +264,8 @@ export default observe(
         ...state,
 
         appStateReady,
-        pctBase: new BN(pctBase),
-        supportRequiredPct: new BN(supportRequiredPct),
+        pctBase: new BN(String(pctBase)),
+        supportRequiredPct: new BN(String(supportRequiredPct)),
         tokenDecimals: new BN(tokenDecimals),
 
         numData: {
@@ -284,11 +284,11 @@ export default observe(
                   data: {
                     ...data,
                     endDate: new Date(data.startDate + voteTime),
-                    minAcceptQuorum: new BN(data.minAcceptQuorum),
+                    minAcceptQuorum: new BN(String(data.minAcceptQuorum)),
                     nay: new BN(data.nay),
-                    totalVoters: new BN(data.totalVoters),
+                    totalVoters: new BN(String(data.totalVoters)),
                     yea: new BN(data.yea),
-                    supportRequiredPct: new BN(supportRequiredPct),
+                    supportRequiredPct: new BN(String(supportRequiredPct)),
                   },
                   numData: {
                     minAcceptQuorum:

```

But I don’t understand why? It comes as strings with other DAOs.

sohkai: @bpierre It's likely the cache, if you hadn't purged local storage before testing.

bpierre: It was the cache 😅 thanks!

apps/voting/app/src/components/SummaryRows.js
bpierre: We should also probably adapt it based `tokenDecimals` if less than 5.

sohkai:
@bpierre Added a few commits:

- 5e0da7b: Don't break the script completely if radspec evaluation breaks in aragon.js
- 3e08b93: Fixes the `getVoteSuccess()` calculation I tried to fix earlier
- 932f00e: Trims the trailing 0s off of formatted token balances

PR #485, opened 2018-10-14
Token Manager: small updates

sohkai:
Builds on #480.

A number of small things, changing some linting behaviour and mostly cosmetic code changes to align the Token Manager with the refactored Voting app in #483.

Easiest to review commit-by-commit.

PR #486, opened 2018-10-14
Token Manager: fix input fields' step behavour and accuracy

sohkai:
Builds on #480 and #485.

Introduces `fromDecimals()` and `toDecimals()` utilities to translate between a string number and a token decimal base (very, very similar to `toWei()` and `fromWei()`).

Fixes the input field not having a step, and the accuracy of the conversion (`parseFloat()` would be inaccurate on large decimal bases).

### DIFF
--- a/apps/token-manager/app/src/App.js
+++ b/apps/token-manager/app/src/App.js
@@ -32,16 +32,14 @@ class App extends React.Component {
     assignTokensConfig: initialAssignTokensConfig,
     sidepanelOpened: false,
   }
-  handleUpdateTokens = ({ mode, amount, holder }) => {
-    const { app, tokenDecimalsBase } = this.props
-    const amountRounded = Math.floor(parseFloat(amount) * tokenDecimalsBase)
-    const amountValue = new BN(`${amountRounded}`, 10).toString()
+  handleUpdateTokens = ({ amount, holder, mode }) => {
+    const { app } = this.props
 
     if (mode === 'assign') {
-      app.mint(holder, amountValue)
+      app.mint(holder, amount)
     }
     if (mode === 'remove') {
-      app.burn(holder, amountValue)
+      app.burn(holder, amount)
     }
 
     this.handleSidepanelClose()
@@ -73,6 +71,7 @@ class App extends React.Component {
     const {
       appStateReady,
       holders,
+      numData,
       tokenAddress,
       tokenSymbol,
       tokenName,
@@ -137,7 +136,7 @@ class App extends React.Component {
           {appStateReady && (
             <AssignVotePanelContent
               opened={sidepanelOpened}
-              tokenDecimalsBase={tokenDecimalsBase}
+              tokenDecimals={numData.tokenDecimals}
               onUpdateTokens={this.handleUpdateTokens}
               {...assignTokensConfig}
             />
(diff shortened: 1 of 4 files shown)


apps/token-manager/app/src/components/Panels/AssignVotePanelContent.js
```diff
@@ -1,8 +1,11 @@
 import React from 'react'
 import styled from 'styled-components'
 import { Button, Field, IconCross, Text, TextInput } from '@aragon/ui'
-import BN from 'bn.js'
 import { addressPattern, isAddress } from '../../web3-utils'
+import { fromDecimals, toDecimals } from '../../utils'
+
+// Any more and the number input field starts to put numbers in scientific notation
+const MAX_INPUT_DECIMAL_BASE = 6
```
sohkai: Any more and we get this:

apps/token-manager/app/src/components/Panels/AssignVotePanelContent.js
```diff
@@ -77,10 +80,11 @@ class AssignVotePanelContent extends React.Component {
   }
   render() {
     const { amount, holder } = this.state
-    const { mode, tokenDecimalsBase } = this.props
-    const tokenBase = tokenDecimalsBase
-      ? new BN(1).div(tokenDecimalsBase).toNumber()
```
sohkai: This was always `0` before 😅

apps/token-manager/app/src/App.js
```diff
@@ -169,15 +168,20 @@ export default observe(
...
         }
       }
-      const { tokenSupply, holders, tokenDecimals } = state
+      const { holders, tokenDecimals, tokenSupply } = state
       const tokenDecimalsBase = new BN(10).pow(new BN(tokenDecimals))
       return {
         ...state,
         appStateReady,
         tokenDecimalsBase,
+        numData: {
```
bpierre: Maybe we should add a comment here, to say that these numbers are not safe integers?

apps/token-manager/app/src/components/Panels/AssignVotePanelContent.js
bpierre: Do you know if it’s anywhere in the HTML spec, or if it depends on the browser implementation?

We should probably have our own validator in the future, to allow the exact decimal base we want (when we will start working on reusable form validation and errors).

apps/token-manager/app/src/utils.js
```diff
@@ -1,5 +1,79 @@
 import BN from 'bn.js'
 
+/**
+ * Get the whole and decimal parts from a number.
+ * Trims leading and trailing zeroes.
+ *
+ * @param {string} num the number
+ * @returns {Array<string>} array with the [<whole>, <decimal>] parts of the number
+ */
+function splitDecimalNumber(num) {
```
bpierre: We should start moving these math and web3 utils somewhere else. Having them in Aragon UI could work, but they might be useful in other places too (e.g. CLI) and it would be weird to import Aragon UI only for that. Should we create a new package, like `aragon-utils` or similar? @izqui @sohkai

sohkai: I was thinking for a while of moving them all to a shared lerna package under aragon-apps, since it's mostly used here.

Let's do a good job of that later, so we can add tests and etc. :).

apps/token-manager/app/src/components/Panels/AssignVotePanelContent.js
sohkai: From https://www.w3.org/TR/html51/sec-forms.html#number-state-typenumber:

> If the user agent provides a user interface for selecting a number, then the value must be set to the best representation of the number representing the user’s selection as a floating-point number.

> The algorithm to convert a number to a string, given a number input, is as follows: Return a valid floating-point number that represents input.

> The best representation of the number n as a floating-point number is the string obtained from running ToString(n). The abstract operation ToString is not uniquely determined. When there are multiple possible strings that could be obtained from ToString for a particular value, the user agent must always return the same string for that value (though it may differ from the value used by other user agents).

Seems like it's pretty up in the air.

sohkai: @bpierre Can confirm, firefox implements this differently. They don't quite support all the way up to 18 decimals, but pretty close.

PR #492, opened 2018-10-16
Vault: remove ability to call contracts with data

### LINKED ISSUE #488: Vault: remove ability to call contracts with data
`Vault.transfer()` should just use the gas-limited `.transfer()` for ETH, to avoid potential security threats.

This will also allow us to remove a bunch of code that was put in place to support the `data` parameter in both `deposit()` and `transfer()`.

bingen:
Closes #488.

### DIFF
--- a/apps/vault/test/vault.js
+++ b/apps/vault/test/vault.js
@@ -23,7 +23,6 @@ contract('Vault app', (accounts) => {
   let ETH, ANY_ENTITY, APP_MANAGER_ROLE, TRANSFER_ROLE
 
   const root = accounts[0]
-  const NO_DATA = '0x'
 
   before(async () => {
     const kernelBase = await Kernel.new(true) // petrify immediately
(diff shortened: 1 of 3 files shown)


apps/vault/test/vault.js
```diff
@@ -96,19 +95,11 @@ contract('Vault app', (accounts) => {
...
       assert.equal((await getBalance(testAccount)).toString(), initialBalance.add(transferValue).toString(), "should have sent eth")
       assert.equal((await getBalance(vault.address)).toString(), depositValue - transferValue, "should have remaining balance")
-
-      // Transfer without data param
-      /* Waiting for truffle to get overloading...
-      await vault.transfer(ETH, testAccount, transferValue)
-
-      assert.equal((await getBalance(testAccount)).toString(), initialBalance.add(transferValue * 2).toString(), "should have sent eth")
-      assert.equal((await getBalance(vault.address)).toString(), depositValue - transferValue * 2, "should have remaining balance")
-      */
```
sohkai: We could add a test with a mock that does too much in its fallback function (e.g. setting some state) to make sure we're only transferring 2.3k gas with `transfer()`.

PR #497, opened 2018-10-17
Fix date format

sohkai:
Fixes the datetime formatting and pins `date-fns` to `@2.0.0-beta.22` so we don't get more unexpected surprises :).

See https://gist.github.com/kossnocorp/a307a464760b405bb78ef5020a4ab136#changed-8 for the new format tokens and https://date-fns.org/v2.0.0-alpha.22/docs/Unicode-Tokens for popularly confused unicode tokens.

### DIFF
--- a/apps/finance/app/src/components/TransferRow.js
+++ b/apps/finance/app/src/components/TransferRow.js
@@ -8,7 +8,6 @@ import {
   ContextMenu,
   ContextMenuItem,
   SafeLink,
-  formatHtmlDatetime,
   theme,
 } from '@aragon/ui'
 import provideNetwork from '../lib/provideNetwork'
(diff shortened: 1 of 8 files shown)


apps/finance/app/src/components/TransferRow.js
```diff
@@ -61,12 +60,12 @@ class TransferRow extends React.Component {
       true,
       { rounding: 5 }
     )
-    const formattedDate = formatHtmlDatetime(date)
+    const formattedDate = format(date, "yyyy-MM-dd'T'HH:mm:ss.SSSxxx")
```
sohkai: Note that these can all be replaced by [`formatHtmlDatetime()` once that's updated](https://github.com/aragon/aragon-ui/pull/223).

Issue #265, opened 2018-04-19
Assign Token action is not creating a transaction


john-light:
Window closes automatically and does not initiate signing.

Firefox - PC

Chrome - Mac

izqui:
Do you recall if this was happening in a multisig org? I think it may have been caused by trying to assign a token to a holder that already had one (and there is a limit of just one token per holder in multisig DAOs).

If that was the case, it will be solved with #291, which will provide a warning or disallow the assign action if we know it is going to fail for this reason.

john-light:
I do not recall and I apologize for not documenting that detail when the issue was first reported. If I had to guess, I would say yes. Go ahead and link #291 to close this issue and if it gets reported again afterwards then we can reopen or create a new issue.

Issue #291, opened 2018-04-27
Token Manager: limit minted tokens in multisig version to just 1


sohkai:
Non-obvious limitation with how we set up the contracts in our multisig templates right now.

sohkai:
It could also detect if someone was already a holder and reject assigning tokens to them in the UI.

bpierre:
Fixed by https://github.com/aragon/aragon-apps/pull/504

PR #504, opened 2018-10-18
Prevent impossible assign

bpierre:
Fixes https://github.com/aragon/aragon-apps/issues/265, https://github.com/aragon/aragon-apps/issues/291.

Prevents impossible assignments to happen when the balance of an account has reached the maximum amount allowed.

- Disable the context menu entry if the maximum is reached.
- Disable the number field too (in case the panel is opened while an update happens).
- Move the “same balance for every holders” check in a separate function.


apps/token-manager/app/src/App.js
```diff
@@ -181,13 +216,11 @@ export default observe(
...
-        holders: holders
-          ? holders
-              .map(holder => ({ ...holder, balance: new BN(holder.balance) }))
-              .sort((a, b) => b.balance.cmp(a.balance))
-          : [],
+        holders: updatedHolders,
         tokenDecimals: new BN(tokenDecimals),
         tokenSupply: new BN(tokenSupply),
+        maxAccountTokens: new BN(maxAccountTokens),
+        groupMode: hasEveryHolderSameBalance(updatedHolders, tokenDecimalsBase),
```
bpierre: @sohkai @izqui What do you think about replacing this by `maxAccountTokens === '1'`?

apps/token-manager/app/src/script.js
```diff
@@ -166,6 +167,17 @@ function updateHolders(holders, changed) {
   }
 }
 
+function loadMaxAccountTokens() {
+  return new Promise((resolve, reject) =>
+    app
+      .call('maxAccountTokens')
+      .first()
+      .subscribe(value => {
```
sohkai: Nitpick: could just be `.subscribe(resolve, reject)`

apps/token-manager/app/src/components/HolderRow.js
```diff
@@ -33,9 +33,14 @@ class HolderRow extends React.Component {
       address,
       balance,
       groupMode,
-      tokenDecimalsBase,
       isCurrentUser,
+      maxAccountTokens,
+      tokenDecimalsBase,
     } = this.props
+
+    const oneToken = formatBalance(balance, tokenDecimalsBase) === '1'
```
sohkai: Another way to check this would be `balance.eq(tokenDecimalsBase)`, which might be nicer since there's less conversions involved?

apps/token-manager/app/src/App.js
sohkai: Yeah, I think it could just be `tokenTransfersEnabled && maxAccountTokens === '1'` (or the equivalent check of `maxAccountTokens.eq(tokenDecimalsBase)`).

That way the UI also wouldn't change when new democracy orgs mint new tokens.

apps/token-manager/app/src/App.js
```diff
@@ -27,11 +29,17 @@ class App extends React.Component {
...
     userAccount: '',
+    groupMode: false,
   }
   state = {
     assignTokensConfig: initialAssignTokensConfig,
     sidepanelOpened: false,
   }
+  getHolderBalance(address) {
+    const { holders } = this.props
+    const holder = holders.find(holder => holder.address === address)
```
sohkai: This should use `addressesEqual()`, otherwise the comparison will fail between checksummed and non-checksummed (resulting in odd amounts, like `2` when they should be `0`).

apps/token-manager/app/src/components/Panels/AssignVotePanelContent.js
```diff
@@ -115,6 +115,8 @@ class AssignVotePanelContent extends React.Component {
               value={amount}
               onChange={this.handleAmountChange}
               min={minTokenStep}
+              {...(maxAmount === '-1' ? {} : { max: maxAmount })}
```
sohkai: As a guard against weird calculated states, it might be worth it to also pass in `maxAccountTokens` here or use it earlier when we set the config, to make sure it's `Math.max(maxAmount, maxAccountTokens)`.

sohkai: Actually, I'm thinking we should move the `maxAmount` calculation completely into the panel, and just give the panel the holder's balance and the Token Manager's `maxAccountTokens`.

Right now, if you open the assign token panel without using the context menu, the max amount doesn't get computed correctly (`getHolderBalance()` always returns `-1`).

apps/token-manager/app/src/web3-utils.js
```diff
@@ -1,10 +1,10 @@
-import { toChecksumAddress } from 'web3-utils'
+import { toChecksumAddress, isAddress } from 'web3-utils'
 
 // Check address equality without checksums
```
sohkai: Let's also change this comment to be **with** checksums :).

apps/token-manager/app/src/web3-utils.js
```diff
@@ -1,10 +1,10 @@
...
 export function addressesEqual(first, second) {
-  first = first && toChecksumAddress(first)
-  second = second && toChecksumAddress(second)
-  return first === second
+  return isAddress(first) && isAddress(second)
+    ? toChecksumAddress(first) === toChecksumAddress(second)
```
sohkai: We should change the above comment to be just `Check address equality` (since `isAddress()` checks for the checksums), and then just compare them lowercase (checksums already passed with `isAddress()`, and converting to lowercase is faster than to the checksummed version).

apps/token-manager/app/src/App.js
```diff
@@ -200,14 +199,16 @@ export default observe(
...
-        ? holders
-            .map(holder => ({ ...holder, balance: new BN(holder.balance) }))
-            .sort((a, b) => b.balance.cmp(a.balance))
-        : []
+      const {
+        holders,
+        maxAccountTokens,
+        tokenDecimals,
+        tokenSupply,
+        tokenTransfersEnabled,
```
sohkai: I think `tokenTransfersEnabled` should be from the props?

apps/token-manager/app/src/components/Panels/AssignVotePanelContent.js
```diff
@@ -2,89 +2,132 @@ import React from 'react'
...
     }
 
     // Finished closing the panel, its state can be reset
     if (!opened && this.props.opened) {
       this.setState({ ...initialState })
     }
   }
-  handleAmountChange = event => {
-    this.setState({ amount: event.target.value })
+  filteredHolderAddress() {
```
sohkai: Should these new functions all use the assignment notation since they access `this`?

bpierre: I usually only do it when the method need to be called from somewhere else, otherwise I try to keep it as a method. That way it’s only declared once on the prototype, rather than being attached to every instance 🏋️‍♂️

apps/token-manager/app/src/components/Panels/AssignVotePanelContent.js
```diff
@@ -2,89 +2,132 @@ import React from 'react'
...
-  handleHolderChange = event => {
-    this.setState({
-      holder: {
+  filteredAmount() {
+    const { tokenDecimals } = this.props
+    const { amountField } = this.state
+    return toDecimals(amountField.value.trim(), tokenDecimals)
+  }
+  updateHolderAddress(mode, value) {
+    const { holderField } = this.state
```
sohkai: Think we can remove this :)

apps/token-manager/app/src/components/Panels/AssignVotePanelContent.js
```diff
@@ -2,89 +2,132 @@ import React from 'react'
...
+    const {
+      mode,
+      tokenDecimals,
+      tokenDecimalsBase,
+      getHolderBalance,
+    } = this.props
+
+    const holderBalance = getHolderBalance(this.filteredHolderAddress())
+
+    const holderBalanceFormatted = formatBalance(
```
sohkai: Don't think this is being used anymore

apps/token-manager/app/src/components/Panels/AssignVotePanelContent.js
```diff
@@ -2,89 +2,132 @@ import React from 'react'
...
+        ...holderField,
+        value,
         error: false,
-        value: event.target.value,
       },
+      amountField: {
+        ...amountField,
+        max: formatBalance(
+          mode === 'assign'
+            ? maxAccountTokens.sub(holderBalance)
```
sohkai: I think this would still be `2` in multisig scenarios when the given holder doesn't have any tokens. `getHolderBalance()` would return `-1` and `maxAccountTokens` would be `1`.

bpierre: I move to use `0` instead of `-1`, which is the correct value anyway 😄

apps/token-manager/app/src/components/Panels/AssignVotePanelContent.js
```diff
@@ -106,24 +126,32 @@ class AssignVotePanelContent extends React.Component {
...
               step={minTokenStep}
               required
               wide
             />
           </Field>
-          <Button mode="strong" type="submit" wide>
+          <Button
+            mode="strong"
+            type="submit"
+            disabled={amountField.max === '0'}
```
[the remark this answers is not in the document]
bpierre: > I promise this is the last comment!

Don’t apologize for making things better! 😄

Here is a message for that case:

image

PR #511, opened 2018-10-19
Token Manager: fix empty state action

sohkai:
Fixes the empty state action button opening up the panel with '[Object object]' because it passes the event object 😅


chris-remus:
Hotfix for @bpierre to review.

Issue #506, opened 2018-10-19
Proposal rejected before it has been voted on


john-light:
Any idea why this is happening? I tried to create a proposal to assign a token to myself when I wasn't a member of the organization yet. It seems to have been rejected before any of the org members have had a chance to vote.

![aragon_vote_rejected](https://user-images.githubusercontent.com/9424721/47213957-d1062580-d39c-11e8-8315-fc6574415dcd.png)

I was told I could create a vote to perform this transaction, so that's what I tried to do and it still failed.

![aragon_vote_rejected_create](https://user-images.githubusercontent.com/9424721/47214193-82a55680-d39d-11e8-88fa-f8cba4d6381a.png)

sohkai:
@john-light Odd, might be the voting UI having a bug calculating if new votes are rejected. Which DAO is this?

john-light:
@sohkai yetanothertest. Note that when I tried to assign the token from an account that was already a member, the vote was created as expected.

PR #512, opened 2018-10-19
Voting: fix misrepresentation of open votes as closed votes

sohkai:
Closes https://github.com/aragon/aragon-apps/issues/506.

[🙈`open` is actually in the `data`](https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/App.js#L166)


chris-remus:
@bpierre to review.

PR #515, opened 2018-10-23
 Migrate arapp.json to use environments feature

izqui:
Aragon CLI `5.2.0-beta.1` hasn't been published yet but it will be published from the current master after testing it can still properly publish all the apps while generating an aragen snapshot.


izqui:
Edit: CLI 5.2.0-beta.1 is now published!

PR #518, opened 2018-10-23
Vault: move test mock to contract/test

sohkai:
Must've missed this during the migration over or something 😅


sohkai:
@bingen it was before... but now maybe all the travis machines got slower or something and all the other PRs' CI are also failing due to this timeout :(.

PR #527, opened 2018-10-25
Finance funding

bpierre:
image

image


izqui:
@bpierre Just detected a little issue with the pagination in Finance. Recent deposits aren't shown at the top of the list when there are too many transactions already:

PR #535, opened 2018-10-28
Token manager: fix the token badge when the label is too long

bpierre:
Before:

image

After:

image

### DIFF
--- a/apps/token-manager/app/src/components/TokenBadge.js
+++ b/apps/token-manager/app/src/components/TokenBadge.js
@@ -8,7 +8,10 @@ class TokenBadge extends React.PureComponent {
       <Main title={`${name} (${address})`}>
         <Label>
           <Icon src={`https://chasing-coins.com/coin/logo/${symbol}`} />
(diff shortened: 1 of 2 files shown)


apps/token-manager/app/src/components/TokenBadge.js
```diff
@@ -8,7 +8,7 @@ class TokenBadge extends React.PureComponent {
       <Main title={`${name} (${address})`}>
         <Label>
           <Icon src={`https://chasing-coins.com/coin/logo/${symbol}`} />
-          {name === symbol ? name : `${name} (${symbol})`}
+          <Name>{name === symbol ? name : `${name} (${symbol})`}</Name>
```
sohkai: I'm wondering if we should capture the symbol first, and then leave the name in the parentheses. They can both be arbitrarily long strings, but I think it's more useful to have the short representation by default (in case the name cuts off the symbol). They're both as easily spoofable, so the full badge would have to be inspected if you wanted to be absolutely 100% sure.

bpierre: Good idea 👍 

image

PR #539, opened 2018-10-29
Finance: fix vault ABI so events from the vault can be processed properly

sohkai:
🙈 Way way back I must not have realized I forgot to include the Vault's events in the ABI we pass to `app.external()`. This adds the relevant ABI and allows the Finance app's script to properly react based on changes to the Vault (mostly important if deposits or transfers happen without the Finance app knowing about it).

### DIFF
--- a/apps/finance/app/src/script.js
+++ b/apps/finance/app/src/script.js
@@ -6,8 +6,10 @@ import tokenBalanceOfAbi from './abi/token-balanceof.json'
 import tokenDecimalsAbi from './abi/token-decimals.json'
 import tokenSymbolAbi from './abi/token-symbol.json'
 import vaultBalanceAbi from './abi/vault-balance.json'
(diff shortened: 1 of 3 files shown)


apps/finance/app/src/script.js
```diff
@@ -72,7 +74,7 @@ retryEvery(retry => {
 })
 
 async function initialize(vaultAddress, ethAddress) {
-  const vaultContract = app.external(vaultAddress, vaultBalanceAbi)
+  const vaultContract = app.external(vaultAddress, vaultAbi)
```
izqui: 🙈🙈🙈🙈

sohkai:
> Or do you prefer releasing a version of Finance that works with Vault 3 first?

I guess we might as well fix it now and update them for `Vault@4` once that's released.

PR #542, opened 2018-10-29
Finance: add list of verified tokens for mainnet

sohkai:
Adds a [list of verified token addresses](https://github.com/MyCryptoHQ/MyCrypto/blob/515cd781324b26fe6990bd84da604cca7f55dfde/common/config/tokens/eth.json), so that if the app is connected to mainnet, only verified tokens get their price and logos loaded.

Increases the script's bundle by ~30kb.


apps/finance/app/src/lib/token-utils.js
```diff
@@ -1,2 +1,10 @@
+import { ETHER_TOKEN_VERIFIED_LIST } from './verified-tokens'
+
 export const ETHER_TOKEN_FAKE_ADDRESS =
   '0x0000000000000000000000000000000000000000'
+
+export const isTokenVerified = (tokenAddress, networkType) =>
+  networkType !== 'mainnet'
```
bpierre: Don’t we want the opposite?

Also we could do it like that:

```jsx
networkType !== 'mainnet' || ETHER_TOKEN_VERIFIED_LIST.has(tokenAddress.toLowerCase())
```

sohkai: Oops, forgot to revert that from testing :D. Good catch!

PR #551, opened 2018-10-30
Fix mainnet detection in apps

sohkai:
🙈 Will also need to bump the version of `@aragon/ui` once https://github.com/aragon/aragon-ui/pull/237 is released.


bpierre:
Done https://www.npmjs.com/package/@aragon/ui

bpierre:
@sohkai I upgraded `@aragon/ui`

Issue #559, opened 2018-11-01
Voting: the voting numbers should start at #1 rather than #0


bpierre:

sohkai:
Closed by #560.

PR #560, opened 2018-11-02
Voting card: fix card numbers

AquiGorka:
Ref #559


AquiGorka:
@bpierre I understand where your suggestion comes from, my thoughts were that this is more a `UI` concern because it has to do with the information rendered for a vote.
Think of a label/title, normally these use some text and some other info (like `id`s) to render information. At contract level, I'd expect votes have zero-based indexes, but labels/titles should be more human-readable friendly.

Let me know your thoughts @bpierre @izqui.

bpierre:
> At contract level, I'd expect votes have zero-based indexes, but labels/titles should be more human-readable friendly.

I agree with that, actually when opening the issue I was imagining that we were using the index of the vote from the contract. But since the contract is using a `voteId` variable, I think it should be considered as an identifier and as such, it doesn’t have to start at 0 and we should avoid making calculations on it.

I could also become a problem if:

- Other representations of the contract forget to add `1` to `voteId` (e.g. a website displaying the latest votes).
- The vote URL (when the single vote view is implemented) shows a different ID than the one displayed.
- More importantly, the ID seen in the app wouldn’t match with the ID seen in other tools. Same as a real life ballot, I believe it’s important that users are, at least in theory, able to verify themselves that what they see in the app is correct. Using a tool to keep track of events on the contract would give a `voteId` value that wouldn’t match with the one displayed in the app.

I think we could consider having a different ID in the frontend than on the contract if they were really different, it’s just the fact that they would only differ by `1` that feels a bit wrong IMO.

AquiGorka:
You are right @bpierre I will update accordingly.

sohkai:
@bpierre @AquiGorka Actually I'd prefer if this was a UI-only change.

The main reason being it'll be difficult to upgrade in the future (some Voting instances will have information at the 0th ID, others won't). On upgrade, they'd also overwrite their last vote with this change.

Some other reasons:

- Originally, we had the votes in an array, and manipulated it with an "empty vote" at the 0th slot so the ID started at 1. This was done inconsistently across the apps, and even between arrays inside the same app sometimes, leading to mild confusion and ease of introducing off-by-one errors. Mentally, it was much harder to audit the contracts.
- We've since moved on to mimicking an array with mappings, to better allow for upgrades later, and this means we do some custom out-of-bounds checks that will also need to be updated with this change.
- Array indices start at 0 for both JS and Solidity, so it feels natural for the first ID to be 0.

------------

Adding 1 to the ID feels odd, and would likely introduce more confusion if someone were to ever look at the on-chain data. I wonder if we could somehow make it feel more like an identifier, rather than just a number? Maybe using a minimum number of 0s (e.g. `000-000-000`) to make it look like IDs people are used to in the real world?

AquiGorka:
We may have more than one problem here:

- Visually have `VoteCards` display a unique `id` that's the same `id` as the one shown in the URL when a user is viewing a specific `VoteCard`.
- `VoteId`'s using incremental numerical values starting at zero feel unnatural when displayed in a list/grid of cards but natural for development.

I like Brett's idea of using a min-number-of-chars `id`, that'll help with the feel when used in lists/grids and maintain mental sanity for developers.

Before committing to any more updates, let me ask a couple of questions:

- Could it be possible to use a similar library to [uuid](https://github.com/kelektiv/node-uuid) in this case? Or was the decision of using numerical `id`s done on purpose to keep the amount of data being stored on-chain at a minimum (double checking and learning here)?
- Does it make sense to introduce these changes now? Or should we work on another visual implementation of a longer `id` and then introduce the changes?

bpierre:
@sohkai

> The main reason being it'll be difficult to upgrade in the future (some Voting instances will have information at the 0th ID, others won't). On upgrade, they'd also overwrite their last vote with this change.

Then I would rather keep it how it is now, starting at 0 and incrementing it. I think it’s fairly reasonable to assume that users will expect the first vote to be 1 (e.g. GitHub issues) , but if we can’t update this in the contract, then it’s IMO more important to keep the ID consistent between the UI and the contract.

> Originally, we had the votes in an array, and manipulated it with an "empty vote" at the 0th slot so the ID started at 1. This was done inconsistently across the apps, and even between arrays inside the same app sometimes, leading to mild confusion and ease of introducing off-by-one errors. Mentally, it was much harder to audit the contracts.

Yeah it’s too late now, but using `voteId = ++votesLength` could have solved this.

> I wonder if we could somehow make it feel more like an identifier, rather than just a number? Maybe using a minimum number of 0s (e.g. 000-000-000) to make it look like IDs people are used to in the real world?

I think starting at `000-000-000` would be as surprising as `0` from the user side. We could have considered a unique ID generated from the contract ID (e.g. `aTb2SzV`), but in this specific case, where the items are sorted by creation order, I believe it makes more sense to keep it as a simple, incremental number.

@AquiGorka

> Does it make sense to introduce these changes now? Or should we work on another visual implementation of a longer id and then introduce the changes?

At this point, I would rather not make any change, and keep it around as a not-so-simple UX problem to be solved later.

sohkai:
> Yeah it’s too late now, but using `voteId = ++votesLength` could have solved this.

This still would've caused the "empty vote" issue, and our bounds checks would have to reflect it. We do still start from index 1 in some cases, but only when we use index 0 as a placeholder for some reason (e.g. to signify a transaction does not have a payment).

> Could it be possible to use a similar library to uuid in this case? Or was the decision of using numerical ids done on purpose to keep the amount of data being stored on-chain at a minimum (double checking and learning here if this data is actually stored on-chain)?

These ids are stored on chain and are how we access information about the particular votes. In this case, we could have hashed the length of the current votes, but then our mimicking of an array wouldn't have worked (and it'd be more painful to check if a particular vote actually existed or was just empty and in which order the votes were created).

-----------

What if we prefixed the number with an "ID: " rather than "#"? I think "#" seems odd, since you expect natural numbers but can see a 0, whereas "ID: " makes it more clear it's an identifier.

bpierre:
image

It’s getting really subjective, but I think it makes the cards feel a bit “technical”, kind of like an admin UI. I would rather keep #, even without having IDs starting at 1, until we find a better solution for it.

@jounih what do you think?

jounih:
Do we even need to show the vote ID at all here? I think the users on this particular screen don't care about the ID number, they just want to:

1. see titles of votes
2. see how people voted and what has passed
3. be able to vote themselves

With the titles themselves sometimes being quite complex and unfriendly (perhaps something we could address in another issue), any more text/information will add to cognitive overload and detract from what they actually want to do.

I would propose to show the ID (and other additional info) on the details screen. If we really need to show it here, I would keep the # rather than ID: - and maybe even have it smaller, above the main title (it isn't really part of the title)

AquiGorka:
@jounih what if we removed the `id` from the title and kept it as the vanilla popover when the user leaves the cursor on top of the title?

bpierre:
I personally use it to communicate what vote I am talking about (“can you check the vote #3?”), but it’s only because we don’t have a single vote URL / view yet.

I would keep it as it is for now until we have a single vote view, just so users can keep communicating a vote easily, and then let’s remove it from the card (except from the title attribute like @AquiGorka suggests 👍).

jounih:
cool works for me!

apps/voting/app/src/components/VotingCard/VotingCard.js
```diff
@@ -41,7 +41,7 @@ class VotingCard extends React.Component {
         </Header>
         <Card>
           <Content>
-            <Question>
+            <Question title={`ID: ${id}`}>
```
bpierre: One last thing (haha it never ends sorry 😄): the `title` attribute takes precedence when being read by web browsers / screen readers, that’s why it should always reflect the full information it represents (ID + title in that case).

AquiGorka: I did not consider showing the title itself, so, sure using: `I am a title (ID: 21)` works for me.

AquiGorka: @bpierre had to go fetch the `description` one component up.

bpierre: It shouldn’t be needed, we already have `question` here! :-)

```suggestion
            <Question title={`${question} (ID: ${id})`}>
```

AquiGorka: `question` is a React component :( that's the first thing I tried, but it has 3 spans.

bpierre: Oooh sorry 🤦‍♂️ I understand now.

PR #566, opened 2018-11-07
Add Delegated Depositing

aunyks:
**NOTE:** This is a two part PR: one part adds a new feature/field and the second is a minute logical edit.

**1.**
Add an `assistedDeposit` function to `Finance.sol` that lets an account deposit funds without needing and paying for gas.  

In essence, this function accepts the same parameters as the existing `deposit` function in addition to `_nonce` and `_signature` parameters. The nonce is used to provide additional entropy when the parameters, excluding `_signature`, are hashed. The resulting hash is then verified alongside the signature to not only gauge the validity of the signature with the function parameters (to prevent a Sybil-like attack) but also recover the address of the signing account. Once this is done the token is deposited from the recovered account.  

A test has been added to account for this function.

**2.**
If an account is depositing an ERC20 token, `_deposit` will ignore its `_sender` parameter and transfer the token from `msg.sender` instead of `_sender`. It should instead listen to `_sender` so that it can properly abstract behavior.

### DIFF
--- a/apps/finance/contracts/ECDSA.sol
+++ b/apps/finance/contracts/ECDSA.sol
@@ -1,0 +1,72 @@
+pragma solidity ^0.4.24;
+
+
+/**
+ * @title Elliptic curve signature operations
+ * @dev Based on https://gist.github.com/axic/5b33912c6f61ae6fd96d6c4a47afde6d
+ * TODO Remove this library once solidity supports passing a signature to ecrecover.
+ * See https://github.com/ethereum/solidity/issues/864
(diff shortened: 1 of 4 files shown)


izqui:
Thanks for the contribution @aunyks!

We are actually thinking about adding this functionality to aragonOS in a future major version, so all functions of an Aragon app can be used in this way, without requiring each app to have its own implementation for checking signatures. You can see the progress here: https://github.com/aragon/aragonOS/pull/442

PR #571, opened 2018-11-13
Voting: add support required to identifier

sohkai:
screen shot 2018-11-13 at 10 00 01 pm

### DIFF
--- a/apps/voting/app/src/App.js
+++ b/apps/voting/app/src/App.js
@@ -16,14 +16,11 @@ import VotePanelContent from './components/VotePanelContent'
 import NewVotePanelContent from './components/NewVotePanelContent'
 import { networkContextType } from './utils/provideNetwork'
 import { settingsContextType } from './utils/provideSettings'
-import { makeEtherscanBaseUrl } from './utils'
 import { hasLoadedVoteSettings } from './vote-settings'
 import { VOTE_YEA } from './vote-types'
-import {
-  EMPTY_CALLSCRIPT,
-  isVoteOpen,
-  voteTypeFromContractEnum,
-} from './vote-utils'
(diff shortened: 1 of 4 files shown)


apps/voting/app/src/script.js
```diff
@@ -1,7 +1,7 @@
 import Aragon from '@aragon/client'
 import { of } from './rxjs'
 import voteSettings, { hasLoadedVoteSettings } from './vote-settings'
-import { EMPTY_CALLSCRIPT } from './vote-utils'
+import { EMPTY_CALLSCRIPT } from './evmscript-utils'
```
sohkai: For some reason, parcel doesn't like tree-shaking `vote-utils`, so having `EMPTY_CALLSCRIPT` there results in the bundle being much larger (~150kb minified).

apps/voting/app/src/App.js
```diff
@@ -16,14 +16,11 @@ import VotePanelContent from './components/VotePanelContent'
...
 import { settingsContextType } from './utils/provideSettings'
-import { makeEtherscanBaseUrl } from './utils'
 import { hasLoadedVoteSettings } from './vote-settings'
 import { VOTE_YEA } from './vote-types'
-import {
-  EMPTY_CALLSCRIPT,
-  isVoteOpen,
-  voteTypeFromContractEnum,
-} from './vote-utils'
+import { makeEtherscanBaseUrl } from './utils'
```
bpierre: We could start using `blockExplorerUrl()` (it’s available in this version of Aragon UI): https://github.com/aragon/aragon-ui/blob/master/src/utils/web3.js#L69

sohkai: Let's wait for the next version and then upgrade them all at the same time :).

PR #573, opened 2018-11-14
Voting: Fix the question and the description (card + panel)

bpierre:
- The card now only displays either the metadata or the radspec description.
- The title attribute has been removed since it has the same content as the children.
- Ethereum addresses are shortened on both the metadata and the description.
- Both the metadata and description render links


apps/voting/app/src/components/VotePanelContent.js
```diff
@@ -205,20 +205,18 @@ class VotePanelContent extends React.Component {
...
+              <Label>Question</Label>
+            </h2>
+            <Question>{questionNode}</Question>
+          </React.Fragment>
+          {descriptionNode && (
             <React.Fragment>
               <h2>
                 <Label>Description</Label>
               </h2>
-              <p>{this.renderDescription(description)}</p>
```
sohkai: I think we can remove `renderDescription()` now :).

apps/voting/app/src/components/VotePanelContent.js
```diff
@@ -205,20 +205,18 @@ class VotePanelContent extends React.Component {
...
-              </h2>
-              <Question>{metadata}</Question>
-            </React.Fragment>
-          )}
-          {description && (
+          <React.Fragment>
+            <h2>
+              <Label>Question</Label>
+            </h2>
+            <Question>{questionNode}</Question>
```
sohkai: Technically this could still be empty... but it's probably better to still have the "Question" label rather than removing it?

sohkai:
@bpierre I've added ec6f581 to have the cards fallback to displaying the radspec description if no metadata is available, and to hide the "Question" label in the panel if there's no question.

apps/voting/app/src/components/VotingCard/VotingCard.js
```diff
@@ -18,7 +18,7 @@ class VotingCard extends React.Component {
     const {
       options,
       endDate,
-      question,
+      description,
```
bpierre: Maybe we could use `label` here?

bpierre:
Awesome thanks @sohkai !

sohkai:
@bpierre Also added bd22eea to linkify the metadata. On AGP-1 the link is actually only in the metadata (technically it'd be very unlikely to have links in the description, since it's from radspec, but we might as well leave it there?).

PR #580, opened 2018-11-16
Agent app: EVM

### LINKED ISSUE #45: Allow to perform actions as token holder with tokens held in Vault
Sometimes the tokens being held in the Vault do not have only economic value, but also allow the holder (the DAO) to perform some actions such as voting or requesting dividends.

A way to forward actions to other contracts as the Vault to perform actions as the token holder is needed.

A more elegant approach would be just a flow of doing an approve/transferFrom with the contract that wants to perform the action as the holder, and at the end return the tokens back to the Vault. This approach wouldn't work to vote with MiniMe tokens, given that the balance is checked at a particular block height.

izqui:
Spec and design discussion: https://forum.aragon.org/t/actor-app-arbitrary-actions-from-daos/275

- [x] 1. Vault inheritance
- [x] 2. Arbitrary call execution
- [x] 3. Forwarding interface
- [x] 4. Signature handling

TODO:
- [x] Roles in arapp

Closes #45.


future-apps/actor/arapp.json
```diff
@@ -0,0 +1,33 @@
...
+      "appName": "actor.aragonpm.eth",
+      "network": "mainnet"
+    },
+    "rinkeby-old": {
+      "registry": "0xfbae32d1cde62858bc45f51efc8cc4fa1415447e",
+      "appName": "actor.aragonpm.eth",
+      "network": "rinkeby"
+    }
+  },
+  "roles": [
```
izqui: TODO

future-apps/actor/contracts/Actor.sol
```diff
@@ -0,0 +1,81 @@
...
+    bytes32 public constant RUN_SCRIPT_ROLE = keccak256("RUN_SCRIPT_ROLE");
+
+    string private constant ERROR_EXECUTE_ETH_NO_DATA = "VAULT_EXECUTE_ETH_NO_DATA";
+    string private constant ERROR_EXECUTE_TARGET_NOT_CONTRACT = "VAULT_EXECUTE_TARGET_NOT_CONTRACT";
+
+    event Execute(address indexed sender, address indexed target, uint256 ethValue, bytes data);
+
+    // TODO: requires the @decodeData helper to be implemented in radspec
+    /**
+    * @notice Execute `@decodeData(target, data)` on `target` `ethValue == 0 ? '' : '(Sending' + @formatToken(ethValue, ETH) + ')'`.
```
izqui: Radspec string won't evaluate until this helper is implemented in radspec or we decide on an alternative

future-apps/actor/contracts/Actor.sol
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    function forward(bytes _evmScript)
+        authP(RUN_SCRIPT_ROLE, arr(getScriptACLParam(_evmScript)))
+        public
+    {
+        bytes memory input = ""; // no input
+        address[] memory blacklist = new address[](0); // no addr blacklist, can interact with anything
+        runScript(_evmScript, input, blacklist);
+        // We don't need to emit an event here as EVMScriptRunner will emit ScriptResult if successful
```
izqui: There is a risk that the default script runner in the DAO won't emit this event and we wouldn't have visibility on Actor app script executions. Considering emitting an event here as well even though it will be redundant in most cases.

future-apps/actor/test/vault.js
```diff
@@ -0,0 +1,238 @@
+// Test that Actor is a fully functioning Vault by running the same tests against the Actor app
```
izqui: Copy pasting this code from the Vault tests feels dirty. @sohkai suggestions on how to do this in a clean way?

future-apps/actor/contracts/Actor.sol
```diff
@@ -0,0 +1,81 @@
...
+            default { return(ptr, size) }
+        }
+    }
+
+    function isForwarder() external pure returns (bool) {
+        return true;
+    }
+
+    function forward(bytes _evmScript)
+        authP(RUN_SCRIPT_ROLE, arr(getScriptACLParam(_evmScript)))
```
izqui: Having an explicit role for this function rather than `require(canForward())` to make the role more explicit (and help our role extraction helpers!)

future-apps/actor/contracts/Actor.sol
```diff
@@ -0,0 +1,184 @@
...
+contract Actor is Vault, IERC165, IERC1271, IForwarder {
+    bytes32 public constant EXECUTE_ROLE = keccak256("EXECUTE_ROLE");
+    bytes32 public constant RUN_SCRIPT_ROLE = keccak256("RUN_SCRIPT_ROLE");
+    bytes32 public constant PRESIGN_HASH_ROLE = keccak256("PRESIGN_HASH_ROLE");
+    bytes32 public constant DESIGNATE_SIGNER_ROLE = keccak256("DESIGNATE_SIGNER_ROLE");
+
+    bytes4 public constant ISVALIDSIG_INTERFACE_ID = 0xabababab; // TODO: Add actual interfaceId
+
+    string private constant ERROR_EXECUTE_ETH_NO_DATA = "VAULT_EXECUTE_ETH_NO_DATA";
+    string private constant ERROR_EXECUTE_TARGET_NOT_CONTRACT = "VAULT_EXECUTE_TARGET_NOT_CONTRACT";
```
bingen: This is 33 chars long, I seem to recall we were using up to 32 to avoid using extra storage.

future-apps/actor/contracts/Actor.sol
onbjerg: What is `decodeData` supposed to do?

future-apps/actor/contracts/Actor.sol
```diff
@@ -0,0 +1,184 @@
...
+    * @param _ethValue Amount of ETH from the contract that is sent with the action
+    * @param _data Calldata for the action
+    * @return Exits call frame forwarding the return data of the executed call (either error or success data)
+    */
+    function execute(address _target, uint256 _ethValue, bytes _data)
+        external // This function MUST always be external as the function performs a low level return, exiting the Actor app execution context
+        authP(EXECUTE_ROLE, arr(_target, _ethValue, uint256(getSig(_data)))) // TODO: Test that sig bytes are the least significant bytes
+    {
+        require(_ethValue == 0 || _data.length > 0, ERROR_EXECUTE_ETH_NO_DATA); // if ETH value is sent, there must be data
+        require(isContract(_target), ERROR_EXECUTE_TARGET_NOT_CONTRACT);
```
jvluso: Is there a reason we don't want the actor app to be able to send eth to someone without calling a function? Anyone able to call the execute function will be able to send funds anyway, and this makes compatibility with certain dapps slightly more difficult.

izqui: Thanks to Actor inheriting from Vault, that can be done using the `transfer` function. I just tried to separate role concerns here.

izqui:
@sohkai do you think you can take a look this next or the next one? It is important for a concierge project

future-apps/actor/contracts/Actor.sol
```diff
@@ -0,0 +1,184 @@
...
+    bytes32 public constant RUN_SCRIPT_ROLE = keccak256("RUN_SCRIPT_ROLE");
+    bytes32 public constant PRESIGN_HASH_ROLE = keccak256("PRESIGN_HASH_ROLE");
+    bytes32 public constant DESIGNATE_SIGNER_ROLE = keccak256("DESIGNATE_SIGNER_ROLE");
+
+    bytes4 public constant ISVALIDSIG_INTERFACE_ID = 0xabababab; // TODO: Add actual interfaceId
+
+    string private constant ERROR_EXECUTE_ETH_NO_DATA = "VAULT_EXECUTE_ETH_NO_DATA";
+    string private constant ERROR_EXECUTE_TARGET_NOT_CONTRACT = "VAULT_EXECUTE_TARGET_NOT_CONTRACT";
+
+    uint256 internal constant STATICCALL_MAX_GAS = 50000;
```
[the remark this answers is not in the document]
izqui: Yup

future-apps/actor/contracts/Actor.sol
```diff
@@ -0,0 +1,184 @@
...
+    * @param _target Address where the action is being executed
+    * @param _ethValue Amount of ETH from the contract that is sent with the action
+    * @param _data Calldata for the action
+    * @return Exits call frame forwarding the return data of the executed call (either error or success data)
+    */
+    function execute(address _target, uint256 _ethValue, bytes _data)
+        external // This function MUST always be external as the function performs a low level return, exiting the Actor app execution context
+        authP(EXECUTE_ROLE, arr(_target, _ethValue, uint256(getSig(_data)))) // TODO: Test that sig bytes are the least significant bytes
+    {
+        require(_ethValue == 0 || _data.length > 0, ERROR_EXECUTE_ETH_NO_DATA); // if ETH value is sent, there must be data
```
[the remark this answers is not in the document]
izqui: Yes, although I am considering removing this

future-apps/actor/contracts/Actor.sol
```diff
@@ -0,0 +1,184 @@
...
+        designatedSigner = _designatedSigner;
+
+        emit SetDesignatedSigner(msg.sender, oldDesignatedSigner, _designatedSigner);
+    }
+
+    function presignHash(bytes32 _hash)
+        external
+        authP(PRESIGN_HASH_ROLE, arr(_hash))
+    {
+        isPresigned[_hash] = true;
```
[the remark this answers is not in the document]
izqui: I don't think we should in the same way that you cannot undo an ECDSA signature. You can always upgrade the app and undo the hash, but I wouldn't make it easy

future-apps/actor/contracts/Actor.sol
```diff
@@ -0,0 +1,184 @@
...
+        return canPerform(sender, RUN_SCRIPT_ROLE, params);
+    }
+
+    function safeSupportsInterface(IERC165 target, bytes4 interfaceId) internal view returns (bool) {
+        if (!isContract(target)) {
+            return false;
+        }
+
+        bytes memory calldata = abi.encodeWithSelector(target.supportsInterface.selector, interfaceId);
+        return safeBoolStaticCall(target, calldata);
```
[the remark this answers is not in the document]
izqui: I didn't know about the two first calls, it really seems like a bit of an overkill to me.

future-apps/actor/contracts/Actor.sol
```diff
@@ -0,0 +1,184 @@
...
+import "./SignatureValidator.sol";
+import "./standards/IERC165.sol";
+import "./standards/IERC1271.sol";
+
+import "@aragon/apps-vault/contracts/Vault.sol";
+
+import "@aragon/os/contracts/common/IForwarder.sol";
+
+
+contract Actor is Vault, IERC165, IERC1271, IForwarder {
```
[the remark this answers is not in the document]
izqui: It would need to be inherited before `Vault` otherwise the linearization fails.

future-apps/actor/contracts/Actor.sol
```diff
@@ -0,0 +1,184 @@
...
+    bytes32 public constant EXECUTE_ROLE = keccak256("EXECUTE_ROLE");
+    bytes32 public constant RUN_SCRIPT_ROLE = keccak256("RUN_SCRIPT_ROLE");
+    bytes32 public constant PRESIGN_HASH_ROLE = keccak256("PRESIGN_HASH_ROLE");
```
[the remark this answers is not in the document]
izqui: I thought of `presign hash` as presign is a verb. We could change it to `ADD_PRESIGNED_HASH_ROLE `

future-apps/actor/contracts/Actor.sol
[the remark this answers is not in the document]
izqui: Changed the gas limit though

future-apps/actor/contracts/SignatureValidator.sol
```diff
@@ -0,0 +1,43 @@
...
+    enum SignatureMode {
+        EIP712,
+        GETH,
+        TREZOR
+    }
+
+    /// @dev Validates that a hash was signed by a specified signer.
+    /// @param hash Hash which was signed.
+    /// @param signer Address of the signer.
+    /// @param signature ECDSA signature along with the mode (0 = EIP712, 1 = Geth, 2 = Trezor) {mode}{v}{r}{s}.
```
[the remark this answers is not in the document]
izqui: I don't have a strong opinion on this, but given that the mode needs to be prefixed anyway, special tooling will be needed regardless.

future-apps/actor/contracts/SignatureValidator.sol
```diff
@@ -0,0 +1,43 @@
+pragma solidity ^0.4.18;
+
+// From github.com/DexyProject/protocol
+// This should probably be moved into aOS: https://github.com/aragon/aragonOS/pull/442
+
+
+library SignatureValidator {
+    // Maybe use 0x's IDs
+    enum SignatureMode {
+        EIP712,
```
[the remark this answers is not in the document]
izqui: It does implement the case, by not re-writing the `hash`

izqui:
Thanks for the thorough review @sohkai. Fixed most issues, let me know whether you think we should get anything else in, specially undecided on https://github.com/aragon/aragon-apps/pull/580#discussion_r251203707

izqui:
Looks like coverage is timing out as it does with Finance. Trying to see if it runs now, otherwise we will have to disable it

future-apps/actor/contracts/Actor.sol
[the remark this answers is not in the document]
sohkai: 👍 Had never occurred to me that was a verb. I still think `ADD_PRESIGNED_HASH_ROLE` is probably better for our sanity (as well as the UI).

future-apps/actor/test/actor.js
```diff
@@ -0,0 +1,351 @@
...
+
+contract('Actor app', (accounts) => {
+  let daoFact, actorBase, acl, actor, actorId
+
+  let ETH, ANY_ENTITY, APP_MANAGER_ROLE, EXECUTE_ROLE, RUN_SCRIPT_ROLE, PRESIGN_HASH_ROLE, DESIGNATE_SIGNER_ROLE, ISVALIDSIG_INTERFACE_ID
+
+  const root = accounts[0]
+
+  const encodeFunctionCall = (contract, functionName, ...params) =>
+    contract[functionName].request(...params).params[0]
```
sohkai: 👀Just curious, where did you find this API? Is it documented anywhere?

sohkai:
@izqui @jvluso Added to resolve compile warnings as well as clean up / add a few more tests.

future-apps/actor/contracts/SignatureValidator.sol
```diff
@@ -0,0 +1,52 @@
+pragma solidity 0.4.24;
+
+// From github.com/DexyProject/protocol
+// This should probably be moved into aOS: https://github.com/aragon/aragonOS/pull/442
+
+
+library SignatureValidator {
+    // Maybe use 0x's IDs
```
sohkai: Yes, I think this makes sense, especially in regards to their `ILLEGAL` and `INVALID` types. We should also check the length using the same enum trick they do.

future-apps/actor/contracts/SignatureValidator.sol
```diff
@@ -0,0 +1,52 @@
...
+            v += 27;
+        }
+
+        if (v != 27 && v != 28) {
+            return false;
+        }
+
+        if (mode == SignatureMode.GETH) {
+            hash = keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", hash));
+        } else if (mode == SignatureMode.TREZOR) {
```
sohkai: Note that Trezor implements the same signing prefix now: https://github.com/ethereum/go-ethereum/issues/14794#issuecomment-425633910

future-apps/actor/test/actor.js
izqui: I discovered this API when working on Aragon 0.1's BylawsLib and I have been using it ever since lol.

izqui: Pretty sure it is an internal thing we are not supposed to be using

future-apps/actor/contracts/Actor.sol
izqui: See https://github.com/aragon/radspec/pull/55

future-apps/agent/contracts/Agent.sol
```diff
@@ -0,0 +1,207 @@
...
+        // and ecrecover is more expensive than an SLOAD.
+        if (isPresigned[hash]) {
+            return true;
+        }
+
+        // Checks if designatedSigner is a contract, and if it supports the isValidSignature interface
+        if (safeSupportsInterface(IERC165(designatedSigner), ISVALIDSIG_INTERFACE_ID)) {
+            // designatedSigner.isValidSignature(hash, signature) as a staticall
+            IERC1271 signerContract = IERC1271(designatedSigner);
+            bytes memory calldata = abi.encodeWithSelector(signerContract.isValidSignature.selector, hash, signature);
```
bingen: I would rename this to `calldata_` or something like this. I was having problems with truffle 5 importing this as a dependency. Same for the 2 other occurrences below.

sohkai: Actually surprised this compiles; `calldata` should be a reserved word, no?

bingen: Yes, it's surprising, I guess it's using the context or something. Anyway, it doesn't work with Truffle 5, and it could be confusing.

future-apps/agent/contracts/standards/ERC1271.sol
```diff
@@ -0,0 +1,41 @@
+pragma solidity 0.4.24;
+
+
+// ERC1271 on Feb 12th, 2019: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1271.md
```
sohkai: Should use a commit hash if you want it to be dated

future-apps/agent/contracts/standards/ERC1271.sol
```diff
@@ -0,0 +1,41 @@
...
+    * @param _hash Arbitrary length data signed on the behalf of address(this)
+    * @param _signature Signature byte array associated with _data
+    * @return A bytes4 magic value 0x20c13b0b if the signature check passes, 0x00000000 if not
+    *
+    * MUST NOT modify state (using STATICCALL for solc < 0.5, view modifier for solc > 0.5)
+    * MUST allow external calls
+    */ 
+    function isValidSignature(bytes32 _hash, bytes memory _signature) public view returns (bytes4);
+
+    function isValidSignatureReturn(bool isValid) internal pure returns (bytes4) {
```
sohkai: I'd name this like `returnIsValidSignatureBytes()` or something. Otherwise it looks too similar to `isValidSignature()`.

future-apps/agent/contracts/SignatureValidator.sol
```diff
@@ -42,18 +80,58 @@ library SignatureValidator {
...
+     * L = l - 1
+     * Input:  ([l1][l2]....[l32][b1][b2]....[bn])
+     * Output: [00]([L1][L2]....[L32][b2]....[bn])
+     */
+    function popFirstByte(bytes memory input) internal pure returns (bytes memory output, byte firstByte) {
+        assembly {
+            let length := mload(input)
+            firstByte := mload(add(input, 0x20))
+            output := add(input, 0x01)
+            mstore8(input, 0x00) // Probably safe to remove, as it will most already likely be 0
```
sohkai: Not sure I get the intention of `mstore8`; why not use a full `mstore`?

future-apps/agent/contracts/SignatureValidator.sol
```diff
@@ -42,18 +80,58 @@ library SignatureValidator {
...
+     * Memory layout: bytes in parenthesis are returned
+     * L = l - 1
+     * Input:  ([l1][l2]....[l32][b1][b2]....[bn])
+     * Output: [00]([L1][L2]....[L32][b2]....[bn])
+     */
+    function popFirstByte(bytes memory input) internal pure returns (bytes memory output, byte firstByte) {
+        assembly {
+            let length := mload(input)
+            firstByte := mload(add(input, 0x20))
+            output := add(input, 0x01)
```
sohkai: Don't we want `add(input, 0x20)` here?

future-apps/agent/contracts/SignatureValidator.sol
sohkai: Ahh nevermind, realizing we're only popping a byte and not a full word

future-apps/agent/contracts/SignatureValidator.sol
```diff
@@ -0,0 +1,137 @@
...
+        if (mode == SignatureMode.EIP712) {
+            return ecVerify(hash, signer, signature);
+        } else if (mode == SignatureMode.EthSign) {
+            return ecVerify(
+                keccak256(abi.encodePacked("\x19Ethereum Signed Message:\n32", hash)),
+                signer,
+                signature
+            );
+        } else if (mode == SignatureMode.ERC1271) {
+            // Pop the mode byte before sending it down the validation chain
```
izqui: `if (signature.length <= 1) return false`

future-apps/agent/contracts/SignatureValidator.sol
```diff
@@ -1,36 +1,74 @@
...
+
     /// @dev Validates that a hash was signed by a specified signer.
     /// @param hash Hash which was signed.
     /// @param signer Address of the signer.
-    /// @param signature ECDSA signature along with the mode (0 = Invalid, 1 = EIP712, 2 = EthSign) {mode}{r}{s}{v}.
+    /// @param signature ECDSA signature along with the mode (0 = Invalid, 1 = EIP712, 2 = EthSign, 3 = ERC1271) {mode}{r}{s}{v}.
     /// @return Returns whether signature is from a specified user.
-    function isValidSignature(bytes32 hash, address signer, bytes signature) internal pure returns (bool) {
+    function isValidSignature(bytes32 hash, address signer, bytes signature) internal view returns (bool) {
         if (signature.length == 0) {
```
sohkai: This should be `<= 1` since mode is always required

future-apps/agent/test/agent.js
```diff
@@ -407,18 +425,18 @@ contract('Agent app', (accounts) => {
...
 
-      it('isValidSignature returns false if designated signer doesnt support the interface', async () => {
+      it('isValidSignature returns true even if the designated signer doesnt support the interface', async () => {
         // false - not ERC165 interface compliant
         // true  - any sigs are valid
         // false - doesn't revert on checking sig
         // false - doesn't modify state on checking sig
         await setDesignatedSignerContract(false, true, false, false)
 
         // Requires ERC165 compliance before checking isValidSignature
```
sohkai: Should amend this comment too; from what I've understood EIP 1271 doesn't require EIP 165?

future-apps/agent/contracts/SignatureValidator.sol
sohkai: This is pretty scary haha. Basically had to run this through remix to see what it would actually do to the memory.

I can see why people prefer to pop the last byte instead of the first byte now (perhaps we should move the mode to the end too?).

future-apps/agent/contracts/standards/ERC1271.sol
```diff
@@ -5,7 +5,7 @@ pragma solidity 0.4.24;
 // Rationale: https://github.com/ethereum/EIPs/issues/1271#issuecomment-462719728
 
 contract ERC1271 {
-    bytes4 constant public ERC1271_INTERFACE_ID = this.isValidSignature.selector;
+    bytes4 constant public ERC1271_INTERFACE_ID = 0xfb855dc9; // this.isValidSignature.selector
```
sohkai: Note that this was giving a warning; I guess solidity doesn't know this is known at compile-time

future-apps/agent/contracts/SignatureValidator.sol
izqui: Will modify to copying into a new array for safety

future-apps/agent/contracts/SignatureValidator.sol
izqui: `== 0` == `<= 1`, I think I don't understand what you mean here

izqui: Length being 1 is valid for ERC1271 mode signatures. We could perform the check before sending it over to the erc1271 contract, but that contract may have logic which allows empty signatures

future-apps/agent/contracts/SignatureValidator.sol
izqui: Not needed, see https://github.com/aragon/aragon-apps/pull/580#discussion_r257318004

future-apps/agent/contracts/SignatureValidator.sol
izqui: Solved in 0d3e2ecde4bd86f90dca8959f2e36efa186d4bc2

izqui:
Branch ready. @sohkai PTAL and let's get this 🚢🚢🚢

apps/agent/contracts/SignatureValidator.sol
```diff
@@ -134,4 +131,24 @@ library SignatureValidator {
 
         return ret;
     }
+
+    // From: https://github.com/Arachnid/solidity-stringutils/blob/master/src/strings.sol
+    function memcpy(uint dest, uint src, uint len) private pure {
```
sohkai: Let's use `uint256` consistently :+1:

apps/agent/contracts/SignatureValidator.sol
```diff
@@ -134,4 +131,24 @@ library SignatureValidator {
...
+            dest += 32;
+            src += 32;
+        }
+
+        // Copy remaining bytes
+        uint mask = 256 ** (32 - len) - 1;
+        assembly {
+            let srcpart := and(mload(src), not(mask))
+            let destpart := and(mload(dest), mask)
+            mstore(dest, or(destpart, srcpart))
```
sohkai: Technically this could be optimized since we don't need to make sure the remaining bytes in `dest` are kept, but let's keep it the same as the library :+1:.

apps/agent/contracts/SignatureValidator.sol
```diff
@@ -84,30 +83,28 @@ library SignatureValidator {
...
-     * @dev DANGEROUS FUNCTION: Modifies input in place, making it invalid after using this function
-     * Memory layout: bytes in parenthesis are returned
-     * L = l - 1
-     * Input:  ([l1][l2]....[l32][b1][b2]....[bn])
-     * Output: [00]([L1][L2]....[L32][b2]....[bn])
-     */
-    function popFirstByte(bytes memory input) internal pure returns (bytes memory output, byte firstByte) {
+    function popFirstByte(bytes memory input) private pure returns (bytes memory output) {
+        uint256 inputLength = input.length;
+        require(inputLength > 0);
```
sohkai: What do you think about adding an error reason here?

apps/agent/contracts/SignatureValidator.sol
izqui: Good observation, would rather not modify the code

PR #584, opened 2018-11-22
Voting: vote analysis script

izqui:


apps/voting/scripts/vote-stats.js
```diff
@@ -0,0 +1,95 @@
...
+      'Supports',
+      'Stake',
+      'Block number',
+      'Transaction hash',
+      'Used Aragon client'
+    ]
+
+    const csvHeader = keys.join(',') + '\n'
+    const csvBody = proccessedVotes.map(
+      vote => Object.values(vote).join(',')
```
sohkai: Just a note that this may not work as expected, as [enumeration order may not be guaranteed](http://2ality.com/2015/10/property-traversal-order-es6.html#traversing-the-own-keys-of-an-object).

A way to guarantee the keys and values line up would be to instead use arrays. It would also make this bit simpler:

```js
const headerKeys = [
  'Voter address',
  'Supports',
  'Stake',
  'Block number',
  'Transaction hash',
  'Used Aragon client'
]

const proccessedVotes = await Promise.all(votes.map(async ({ transactionHash, args }) => {
  const { voter, supports, stake } = args
  const { blockNumber, input } = await getTransaction(transactionHash)
  return [
    voter,
    supports,
    stake: formatNumber(stake, decimals),
    blockNumber,
    transactionHash,
    clientVote: input.endsWith('1')
  ]
}))

const csvData = [headerKeys, ...processedVotes]
const csvText = csvData.map(data => data.join(',')).join('\n')

// write to file
```

PR #599, opened 2018-11-28
Vertically center empty screens

### LINKED ISSUE #598: Empty screens are not centered anymore
Not vertically centered anymore

image

bpierre:
Fixes #598

### DIFF
--- a/apps/finance/app/src/App.js
+++ b/apps/finance/app/src/App.js
@@ -99,65 +99,72 @@ class App extends React.Component {
     return (
       <PublicUrl.Provider url="./aragon-ui/">
         <BaseStyles />
-        <AppView
-          appBar={
-            <AppBar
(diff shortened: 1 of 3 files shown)


apps/finance/app/src/App.js
```diff
@@ -99,65 +99,72 @@ class App extends React.Component {
...
+            />
+          </SidePanel>
+        </Main>
       </PublicUrl.Provider>
     )
   }
 }
 
+const Main = styled.div`
+  height: 100vh;
```
sohkai: Just wanted to double check this doesn't run into the same problem we had with `vh` due to the banner. I think it shouldn't be a problem since this is would be inside the iframe?

bpierre: Yes exactly 👍

PR #603, opened 2018-12-08
Finance: add asterisks to all required input fields

sohkai:
Fixes https://github.com/aragon/aragon/issues/508, adding asterisks to these panels:

screen shot 2018-12-08 at 12 35 52 pm

-----

screen shot 2018-12-08 at 12 35 59 pm

-----

screen shot 2018-12-08 at 12 36 04 pm

------

Also makes the label unselectable in cases where we used a custom label.

### DIFF
--- a/apps/finance/app/src/components/NewTransfer/Withdrawal.js
+++ b/apps/finance/app/src/components/NewTransfer/Withdrawal.js
@@ -9,6 +9,7 @@ import {
   Text,
   TextInput,
   theme,
+  unselectable,
 } from '@aragon/ui'
 import { toDecimals } from '../../lib/math-utils'
 import { addressPattern, isAddress } from '../../lib/web3-utils'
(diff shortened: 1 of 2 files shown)


apps/finance/app/src/components/NewTransfer/Withdrawal.js
```diff
@@ -202,6 +203,16 @@ const CombinedInput = styled.div`
   }
 `
 
+const StyledAsterisk = styled.span.attrs({
+  children: '*',
+  title: 'Required',
+})`
+  color: ${theme.accent};
+  float: right;
```
AquiGorka: I won't be able to sleep anymore :(
Can I convince you to use `justify-content: space-between` (either `flex` or `grid`) on the parent container?

sohkai: Haha actually this was taken from [@aragon/ui's Field](https://github.com/aragon/aragon-ui/blob/master/src/components/Field/Field.js#L14) but we can look at changing that.

sohkai: See https://github.com/aragon/aragon-ui/pull/275 :)

PR #606, opened 2018-12-10
Responsive menu panel for iframe apps

AquiGorka:
These changes for the `Voting`, `Token Manager` and `Finance` apps add the mechanism to open/close the `MenuPanel` as shown [here](https://github.com/aragon/aragon/pull/539)

### DIFF
--- a/apps/finance/app/src/components/MenuButton/IconMenu.js
+++ b/apps/finance/app/src/components/MenuButton/IconMenu.js
@@ -1,0 +1,15 @@
+import React from 'react'
+
+const Menu = props => (
+  <svg width="18px" height="14px" viewBox="0 0 18 14" {...props}>
+    <g transform="translate(-19, -25)" stroke="currentColor">
+      <g transform="translate(20, 25)">
+        <path d="M0,1 L16,1" />
+        <path d="M0,7 L16,7" />
+        <path d="M0,13 L16,13" />
+      </g>
+    </g>
+  </svg>
+)
+
+export default Menu
--- a/apps/token-manager/app/src/App.js
+++ b/apps/token-manager/app/src/App.js
@@ -12,10 +12,12 @@ import {
   SidePanel,
   font,
   observe,
+  BreakPoint,
 } from '@aragon/ui'
 import EmptyState from './screens/EmptyState'
 import Holders from './screens/Holders'
 import AssignVotePanelContent from './components/Panels/AssignVotePanelContent'
+import MenuButton from './components/MenuButton/MenuButton'
 import { networkContextType } from './provide-network'
 import { hasLoadedTokenSettings } from './token-settings'
 import { makeEtherscanBaseUrl } from './utils'
@@ -29,6 +31,7 @@ const initialAssignTokensConfig = {
 class App extends React.Component {
   static propTypes = {
     app: PropTypes.object.isRequired,
+    sendMessageToWrapper: PropTypes.func.isRequired,
   }
   static defaultProps = {
     appStateReady: false,
@@ -88,6 +91,9 @@ class App extends React.Component {
       sidepanelOpened: true,
     })
   }
+  handleMenuPanelOpen = () => {
+    this.props.sendMessageToWrapper('menuPanel', true)
+  }
   handleSidepanelClose = () => {
     this.setState({ sidepanelOpened: false })
   }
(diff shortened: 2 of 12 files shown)


apps/token-manager/app/src/App.js
```diff
@@ -57,7 +59,7 @@ class App extends React.Component {
   getHolderBalance = address => {
     const { holders } = this.props
     const holder = holders.find(holder =>
-      addressesEqual(holder.address, address)
+      addressesEqual(holder.address, address),
```
bpierre: I realize we don’t have a `.prettierrc` per app, but we should use the ES5 style trailing comma ([`trailingComma: "es5"`](https://github.com/aragon/aragon-apps/blob/master/apps/finance/app/.prettierrc#L4)) for consistency.

apps/finance/app/src/components/MenuButton/IconMenu.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+const Menu = props => (
+  <svg width="18px" height="14px" viewBox="0 0 18 14" {...props}>
+    <g transform="translate(-19.000000, -25.000000)" stroke="currentColor">
```
bpierre:
```suggestion
    <g transform="translate(-19, -25)" stroke="currentColor">
```

apps/finance/app/src/components/MenuButton/IconMenu.js
```diff
@@ -0,0 +1,15 @@
...
+      <g transform="translate(20.000000, 25.000000)">
```
bpierre:
```suggestion
      <g transform="translate(20, 25)">
```

AquiGorka:
@bpierre all comments addressed too :)

apps/voting/app/src/components/MenuButton/MenuButton.js
```diff
@@ -0,0 +1,31 @@
...
+  height: 32px;
+  width: 32px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  padding: 0;
+  cursor: pointer;
+  outline: none;
+
+  &:focus {
```
bpierre: Ideally we should use `:focus-visible` here, but it is not supported by Chrome yet, so we have [`<FocusVisible />`](https://github.com/aragon/aragon-ui/blob/master/src/components/FocusVisible/FocusVisible.js) that can be used in the meantime.

Usage:

```jsx
<FocusVisible>
  (({ focusVisible }) => (
    <div css={`
      :focus {
        border: `${Number(focusVisible)}px solid blue`;
      }
    `} />
  ))
</FocusVisible>
```

apps/voting/app/src/components/MenuButton/MenuButton.js
```diff
@@ -0,0 +1,31 @@
...
+    border: 2px solid ${theme.accent};
+  }
+  &:active {
```
bpierre: If we use `FocusVisible`, we don’t need this anymore.

apps/voting/app/src/components/MenuButton/MenuButton.js
AquiGorka: Quick question: when Chrome starts supporting `:focus-visible` the `css` will work no?
What I am aiming for is that the solution would be to: a)keep the `css` rules, b) use the `FocusVisible` container and c) Remove `FocusVisible` when Chrome supports `:focus-visible`

Right? (what a mouth full, jajaja)

bpierre: The day Chrome starts supporting `:focus-visible`, using `FocusVisible` + `:focus` will still work, like it does in Firefox today. 👍

At this point, we will move away from `FocusVisible` + `:focus` and use `:focus-visible` instead.

e.g. in this particular case, we could have this today:

```jsx
 const StyledButton = styled.button`
  /* … */
  outline: 0;
  
  /* either wrap the entire :focus declaration: */
  ${p => p.focusVisible? `
    &:focus {
      outline: 2px solid ${theme.accent}`;
    }
  ` : ''}

  /* or on the value since there is only one: */
  &:focus {
    outline: ${p => p.focusVisible? `2px solid ${theme.accent}` : '0'};
  }
`

export default props => (
  <FocusVisible>
    (({ focusVisible }) => (
      <StyledButton focusVisible={focusVisible} {...props}>
        <IconMenu />
      </StyledButton>
    ))
  </FocusVisible>
)
```

And this tomorrow:


```jsx
 const StyledButton = styled.button`
  /* … */
  outline: 0;

  &:focus-visible {
    outline: 2px solid ${theme.accent};
  }
`

export default props => (
  <StyledButton focusVisible={focusVisible} {...props}>
    <IconMenu />
  </StyledButton>
)
```

PR #611, opened 2018-12-14
Merge @next contract changes

sohkai:
Long-lived PR to help us track contract changes in the `next` branch for future contract deployments.

**Note**: No release was made immediately as a result of merging these changes, as more changes were added afterwards that also required security review.


tintinweb:
Hey guys,

Adding comments directly to the PR. Let me know if you have any specific format preferences for reporting issues. Here are some observations:


##### 1. (medium) - General: survey.js/finance.js/voting.js/token-manager.js/a.o. - missing security specific unittests
###### description
* Verify that all externally-accessible, state-changing functionality is protected by authentication and the ACL system is set-up correctly. Consider adding ACL/permission specific unittests. add pos/negative checks to verify that only permitted accounts are allowed to call privileged methods.
* e.g. see `survey.sol` below; unittests are set-up to grant the permissions `CREATE_SURVEYS_ROLE`, `MODIFY_PARTICIPATION_ROLE` to `ANY_ENTITY` essentially allowing anyone to perform authenticated actions not testing if the app specific ACL configuration is set up correctly.
###### location
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/survey/test/survey.js#L52-L53
###### suggested fix (incomplete example)
```js
const { assertInvalidOpcode, assertRevert } = require('@aragon/test-helpers/assertThrow')
const getBlockNumber = require('@aragon/test-helpers/blockNumber')(web3)
const timeTravel = require('@aragon/test-helpers/timeTravel')(web3)

const pct16 = x => new web3.BigNumber(x).times(new web3.BigNumber(10).toPower(16))
const surveyEvent = receipt => receipt.logs.filter(x => x.event == 'StartSurvey')[0].args
const createdSurveyId = receipt => surveyEvent(receipt).surveyId

contract('Security - Survey app ACL system', accounts => {
  let daoFact, dao, surveyBase, survey

  let ABSTAIN_VOTE, ANY_ENTITY
  let APP_MANAGER_ROLE, CREATE_SURVEYS_ROLE, MODIFY_PARTICIPATION_ROLE

  const root = accounts[0]
  const holder19 = accounts[0]
  const holder31 = accounts[1]
  const holder50 = accounts[2]
  const nonHolder = accounts[4]
  const unprivileged_account = accounts[5]
  const privileged_account = accounts[6]

  const NULL_ADDRESS = '0x00'

  const SURVEY_APP_ID = '0x1234'
  const SURVEY_MOCK_APP_ID = '0x12341234'
  const minimumAcceptanceParticipationPct = pct16(20)
  const surveyTime = 1000

  before(async () => {
    const kernelBase = await getContract('Kernel').new(true) // petrify immediately
    const aclBase = await getContract('ACL').new()
    daoFact = await getContract('DAOFactory').new(kernelBase.address, aclBase.address, NULL_ADDRESS)
    surveyBase = await getContract('Survey').new()

    // Setup constants
    ANY_ENTITY = await aclBase.ANY_ENTITY()
    APP_MANAGER_ROLE = await kernelBase.APP_MANAGER_ROLE()
    ABSTAIN_VOTE = await surveyBase.ABSTAIN_VOTE()
    CREATE_SURVEYS_ROLE  = await surveyBase.CREATE_SURVEYS_ROLE()
    MODIFY_PARTICIPATION_ROLE  = await surveyBase.MODIFY_PARTICIPATION_ROLE()
  })

  beforeEach(async () => {
    const r = await daoFact.newDAO(root)
    dao = getContract('Kernel').at(r.logs.filter(l => l.event == 'DeployDAO')[0].args.dao)
    const acl = getContract('ACL').at(await dao.acl())

    await acl.createPermission(root, dao.address, APP_MANAGER_ROLE, root, { from: root })

    const receipt = await dao.newAppInstance(SURVEY_APP_ID, surveyBase.address, '0x', false, { from: root })
    survey = getContract('Survey').at(receipt.logs.filter(l => l.event == 'NewAppProxy')[0].args.proxy)

    /* check survey specific ACL */
    await acl.createPermission(privileged_account, survey.address, CREATE_SURVEYS_ROLE, root, { from: root })
    await acl.createPermission(privileged_account, survey.address, MODIFY_PARTICIPATION_ROLE, root, { from: root })
  })

  context('normal token supply', () => {
    let token

    beforeEach(async () => {
      token = await getContract('MiniMeToken').new(NULL_ADDRESS, NULL_ADDRESS, 0, 'n', 0, 'n', true) // empty parameters minime

      await token.generateTokens(holder19, 19)
      await token.generateTokens(holder31, 31)
      await token.generateTokens(holder50, 50)
    })
    it('anyone can initialize survey (potential race?)', async () => {
      /* */ 
      await survey.initialize(token.address, minimumAcceptanceParticipationPct, surveyTime, { from: unprivileged_account })
    })

    it('privileged account - can change minimum acceptance participation', async () => {
      await survey.initialize(token.address, minimumAcceptanceParticipationPct, surveyTime, { from: unprivileged_account })

      const receipt = await survey.changeMinAcceptParticipationPct(1,  {from: privileged_account })
      const events = receipt.logs.filter(x => x.event == 'ChangeMinParticipation')

      assert.equal(events.length, 1, 'should have emitted ChangeMinParticipation event')
      assert.equal(await survey.minParticipationPct(), 1, 'should have change acceptance participation')
    })

    it('unprivileged account - cannot change minimum acceptance participation', async () => {
      await survey.initialize(token.address, minimumAcceptanceParticipationPct, surveyTime, { from: unprivileged_account })
      await assertRevert(async () => {
        await survey.changeMinAcceptParticipationPct(1, {from: unprivileged_account})
      })
    })

    it('root - cannot change minimum acceptance participation', async () => {
      await survey.initialize(token.address, minimumAcceptanceParticipationPct, surveyTime, { from: unprivileged_account })
      await assertRevert(async () => {
        await survey.changeMinAcceptParticipationPct(1, {from: root})
      })
    })

    context('creating survey', () => {
      const optionsCount = 10e9 // lots of options
      let surveyId

      beforeEach(async () => {
        await survey.initialize(token.address, minimumAcceptanceParticipationPct, surveyTime, { from: unprivileged_account })
      })

      it('privileged account - create survey succeeds and getSurvey returns valid data', async () => {
        createdEvent = surveyEvent(await survey.newSurvey('metadata', optionsCount, { from: privileged_account }))
        surveyId = createdEvent.surveyId
        creator = createdEvent.creator
        metadata = createdEvent.metadata

        const [isOpen, startDate, snapshotBlock, minParticipationPct, votingPower, participation, options] = await survey.getSurvey(surveyId)

        assert.isTrue(isOpen, 'survey should be open')
        assert.equal(creator, privileged_account, 'creator should be correct')
        assert.equal(snapshotBlock, await getBlockNumber() - 1, 'snapshot block should be correct')
        assert.deepEqual(minParticipationPct, minimumAcceptanceParticipationPct, 'min participation should be survey min participation')
        assert.equal(votingPower, 100, 'voting power should be 100')
        assert.equal(participation, 0, 'initial participation should be 0') // didn't vote even though creator was holder
        assert.equal(options, optionsCount, 'number of options should be correct')
        assert.equal(metadata, 'metadata', 'should have returned correct metadata')
        const voterState = await survey.getVoterState(surveyId, privileged_account)
        assert.equal(voterState[0].length, 0, 'nonHolder should not have voted (options)')
        assert.equal(voterState[1].length, 0, 'nonHolder should not have voted (stakes)')
      })

      it('unprivileged account - create survey fails', async () => {
        await assertRevert(async () => {
          await survey.newSurvey('metadata', optionsCount, { from: unprivileged_account })
        })
      })      

      it('root account - create survey fails', async () => {
        await assertRevert(async () => {
          await survey.newSurvey('metadata', optionsCount, { from: root })
        })
      })    
    })
  })

  context('before init', () => {
    it('fails creating a survey before initialization', async () => {
      await assertRevert(async () => {
        await survey.newSurvey('metadata', 10, {from: privileged_account})
      })
    })
    it('fails changing MinAcceptParticipationPct before initialization', async () => {
      await assertRevert(async () => {
        await survey.changeMinAcceptParticipationPct(1, {from: privileged_account})
      })
    })
  })  
})
```
----------------------------

##### 2. (info) - survey.sol -  var can be declared as uint64
###### description
const `PCT_BASE` fits into uint64. consider declaring it as `uint256` --> `uint64`
###### location
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/survey/contracts/Survey.sol#L22

----------------------------

##### 3. (info) - voting.sol - `_minAcceptQuorumPct > 0` and `_supportRequiredPct <= PCT_BASE`
###### description
* enforce `_minAcceptQuorumPct` is not zero. (would disable min quorum; should this be allowed?)
* allow `_supportRequiredPct` to be 100%
###### location
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/voting/contracts/Voting.sol#L90-L91
###### suggested fix
```diff
--- a/apps/voting/contracts/Voting.sol
+++ b/apps/voting/contracts/Voting.sol
@@ -87,8 +87,8 @@ contract Voting is IForwarder, AragonApp {
     {
         initialized();
 
-        require(_minAcceptQuorumPct <= _supportRequiredPct, ERROR_INIT_PCTS);
-        require(_supportRequiredPct < PCT_BASE, ERROR_INIT_SUPPORT_TOO_BIG);
+        require(_minAcceptQuorumPct > 0 && _minAcceptQuorumPct <= _supportRequiredPct, ERROR_INIT_PCTS);
+        require(_supportRequiredPct <= PCT_BASE, ERROR_INIT_SUPPORT_TOO_BIG);
```
###### location
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/voting/contracts/Voting.sol#L122
###### suggested fix
```diff
-        require(_minAcceptQuorumPct <= supportRequiredPct, ERROR_CHANGE_QUORUM_PCTS);
+        require(_minAcceptQuorumPct > 0 && _minAcceptQuorumPct <= supportRequiredPct, ERROR_CHANGE_QUORUM_PCTS);
```
###### location
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/voting/contracts/Voting.sol#L108
###### suggested fix
* see previous instance
-----------------------------

##### 4. (info) - voting.sol - access modifiers - consider declaring functions external (optimization)
###### description
* function is not used internally and can therefore be declared external
* also note the mismatching declaration of `isForwarder()` compared to `IForwarder`
###### location
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/voting/contracts/Voting.sol#L178
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/voting/contracts/Voting.sol#L187
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/voting/contracts/Voting.sol#L233
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/voting/contracts/Voting.sol#L264
###### suggested fix
* declare function `external` unless it is planned to be called internally as well.

-------------------------------

##### 5. (info) - survey.sol - access modifiers - consider declaring functions external (optimization)
###### description
* function is not used internally and can therefore be declared external
###### location
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/survey/contracts/Survey.sol#L240
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/survey/contracts/Survey.sol#L259
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/survey/contracts/Survey.sol#L305
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/survey/contracts/Survey.sol#L312
###### suggested fix
* declare function `external` unless it is planned to be called internally as well.

##### 6. (info) - survey.sol - local variable shadows state var
###### description
* local var `minParticipationPct` (declared in returns) shadows state var with same name.
###### location
https://github.com/aragon/aragon-apps/blob/25d458508d80c633c7c44e54e85dd5b8b4ab5f66/apps/survey/contracts/Survey.sol#L278
###### suggested fix
* rename local var to avoid shadowing state variables (readability)

sohkai:
Thanks for the thorough review, @tintinweb! What do you think about making separate comments for each of the issues so that we can respond directly to them in a thread?

> 1. (medium) - General: survey.js/finance.js/voting.js/token-manager.js/a.o. - missing security specific unittests

The ACL authentication is something we've [tested thoroughly as part of aragonOS](https://github.com/aragon/aragonOS/blob/dev/test/app_acl.js), but I do agree each app should have its own tests to ensure functionality it wants to protect is indeed protected by a role. I wouldn't consider this a blocker, and is more related to tech debt.

I believe this was brought up as part of the last CD audit as well 😄... so I should get on doing this.

> 2. (info) - survey.sol / vote.sol - inconsistent declaration of PCT_BASE

Indeed, we choose to keep them as `uint256` due to it being lower in gas (using `uint64` would cause extra masks). Perhaps a comment would suffice here?

> 3. (info) - voting.sol - `_minAcceptQuorumPct > 0` and `_supportRequiredPct <= PCT_BASE`

These conditions are meant to be "strict" percentage checks (ie. failing on equality), to enforce things like "strict majority" (where you need > 50% to pass, rather than >= 50%). As it is, `minAcceptQuorumPct == 0` is acceptable—there is simply no quorum requirement, and `supportRequiredPct == PCT_BASE` is not acceptable—this causes a deadlock in the voting execution (even at 100% the vote would be considered failed).

See https://github.com/aragon/aragon-apps/pull/459 for this particular change and its reasoning.

> 4. (info) - voting.sol - access modifiers - consider declaring functions external (optimization)
> 5. (info) - survey.sol - access modifiers - consider declaring functions external (optimization)

Ahh, we forgot to make `isForwarder()` external after moving to `solidity@0.4.24`—I'll do that. Taking another look through, I did notice a number of other functions that could be divided into `external` and `internal` versions; I'll also make a PR for this.

For others though, we've made conscious decisions to keep some functions `public` in the case our app contracts are ever used as a superclass by other contracts (you'll notice in Finance and Token Manager we keep a lot of the getters as `public` as well).

> 6. (info) - survey.sol - local variable shadows state var

Fixed in https://github.com/aragon/aragon-apps/pull/684 :).

tintinweb:
Thanks for the quick response @sohkai!

Since some of the issues raised weren't actually part of the PRs change I wasn't able to comment them inline (preferred), hence the all-in-one comment. I'll try to add comments directly to the changed lines if possible and will provide comments on issues not directly as individual comments to the PR in the future.

ad 1) even though it is not a blocker I'd suggest to treat this with priority (aligned with your development plan) otherwise accidental changes to the apps individual auth settings might stay undetected. In contrast to the ACL unit-tests which focus on general ACL base functionality the important part here is to focus on the app specific authentication configuration (i.e. verify the assumption that function X a) can only be called by permitted account providing correct args b) cannot be called by permitted account providing mismatching args b) cannot be called by other accounts). Thanks for tracking this as #685 👍 

ad 2) unfort. the inconsistent declaration of `PCT_BASE` in `vote.sol` and `survey.sol` is part of this PR. I can file a general issue for that unless you think it is an overkill :)

ad 3) thanks for the clarification!

ad 4) In case you're creating a PR for this and want to review other instances as well I can probably provide a list for review.

sohkai:
> ad 2) unfort. the inconsistent declaration of `PCT_BASE` in `vote.sol` and `survey.sol` is part of this PR. I can file a general issue for that unless you think it is an overkill :)

Ahh good catch, it looks like https://github.com/aragon/aragon-apps/pull/473 changed them both but then reverted it in the Survey app. Although using `uint256` results in one less mask every time `PCT_BASE` is used, I think keeping it `uint64` makes sense since the compiler will complain if we try to set a value that's too high (especially useful given how close 10^18 is to 2^64): https://github.com/aragon/aragon-apps/pull/687.

.travis.yml
```diff
@@ -38,7 +42,10 @@ env:
     - TASK=test:shared
 matrix:
   allow_failures:
+    - env: TASK=coverage:agent
```
tintinweb: failed to manually produce coverage statistics for agent. Assumed it was allowed to fail for the same reason `coverage:finance` is in `allow_failures` (timeout) but observed this behaviour where the agent **unittests pass** without error but under instrumentation for **coverage** analysis four **tests fail** (therefore wrong coverage stats). 

Is this a known issue and the reason for `coverage:agent` being allowed to fail?

```  
 78 passing (53s)
   4 failing

   1) Contract: Agent app
        > Signing messages
          > Designated signer
            > Signature mode: ERC1271 -> EIP712
              isValidSignature returns true to a valid signature:
       Expected signature to be valid (returned 0x00000000)
       + expected - actual
       -0x00000000
       +0x20c13b0b
       
       at assertIsValidSignature (test/agent.js:268:14)
       at Context.it (test/agent.js:396:13)
       at process.internalTickCallback (internal/process/next_tick.js:77:7)
 
   2) Contract: Agent app
        > Signing messages
          > Designated signer
            > Signature mode: ERC1271 -> EIP712
              isValidSignature returns true to a valid signature with legacy version:
       Expected signature to be valid (returned 0x00000000)
       + expected - actual
       -0x00000000
       +0x20c13b0b
       
       at assertIsValidSignature (test/agent.js:268:14)
       at Context.it (test/agent.js:401:13)
       at process.internalTickCallback (internal/process/next_tick.js:77:7)
 
   3) Contract: Agent app
        > Signing messages
          > Designated signer
            > Signature mode: ERC1271 -> EthSign
              isValidSignature returns true to a valid signature:
       Expected signature to be valid (returned 0x00000000)
       + expected - actual
       -0x00000000
       +0x20c13b0b
       
       at assertIsValidSignature (test/agent.js:268:14)
       at Context.it (test/agent.js:396:13)
       at process.internalTickCallback (internal/process/next_tick.js:77:7)
 
   4) Contract: Agent app
        > Signing messages
          > Designated signer
            > Signature mode: ERC1271 -> EthSign
              isValidSignature returns true to a valid signature with legacy version:
       Expected signature to be valid (returned 0x00000000)
       + expected - actual
       -0x00000000
       +0x20c13b0b
       
       at assertIsValidSignature (test/agent.js:268:14)
       at Context.it (test/agent.js:401:13)
       at process.internalTickCallback (internal/process/next_tick.js:77:7)
 
 -------------------------|----------|----------|----------|----------|----------------|
 File                     |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
 -------------------------|----------|----------|----------|----------|----------------|
  contracts/              |    96.92 |       90 |    94.12 |    97.37 |                |
   Agent.sol              |    95.45 |    83.33 |       90 |    95.83 |            102 |
   SignatureValidator.sol |    97.67 |    91.67 |      100 |    98.08 |            126 |
  contracts/standards/    |       50 |      100 |       50 |       50 |                |
   ERC1271.sol            |       50 |      100 |       50 |       50 |             42 |
   IERC165.sol            |      100 |      100 |      100 |      100 |                |
 -------------------------|----------|----------|----------|----------|----------------|
 All files                |    95.52 |       90 |    89.47 |    96.15 |                |
 -------------------------|----------|----------|----------|----------|----------------|
```

.travis.yml
```diff
@@ -38,7 +42,10 @@ env:
...
     - env: TASK=coverage:finance
```
tintinweb: ℹ️for reference: manually produced coverage stats for `coverage:finance`:

* Finance 
```
--------------|----------|----------|----------|----------|----------------|
File          |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
--------------|----------|----------|----------|----------|----------------|
 contracts/   |    99.38 |    92.65 |     97.3 |      100 |                |
  Finance.sol |    99.38 |    92.65 |     97.3 |      100 |                |
--------------|----------|----------|----------|----------|----------------|
All files     |    99.38 |    92.65 |     97.3 |      100 |                |
--------------|----------|----------|----------|----------|----------------|
```

.travis.yml
sohkai: @tintinweb Yes, this is a known issue that we haven't been too concerned about yet (since the whole thing isn't _exactly_ ready and we're honestly a bit stumped at how this could be failing coverage). It does also timeout, because `solidity-coverage` goes crazy on its instrumentation.

apps/agent/contracts/Agent.sol
```diff
@@ -0,0 +1,147 @@
...
+        params[0] = getScriptACLParam(evmScript);
+        return canPerform(sender, RUN_SCRIPT_ROLE, params);
+    }
+
+    function getScriptACLParam(bytes evmScript) internal pure returns (uint256) {
+        return uint256(keccak256(abi.encodePacked(evmScript)));
+    }
+
+    function getSig(bytes data) internal pure returns (bytes4 sig) {
+        assembly { sig := add(data, 0x20) }
```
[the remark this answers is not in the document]
izqui: 🤦🏻‍♀️ good catch, thanks

apps/agent/contracts/Agent.sol
```diff
@@ -0,0 +1,147 @@
...
+        bytes memory input = ""; // no input
+        address[] memory blacklist = new address[](0); // no addr blacklist, can interact with anything
+        runScript(_evmScript, input, blacklist);
+        // We don't need to emit an event here as EVMScriptRunner will emit ScriptResult if successful
+    }
+
+    function isValidSignature(bytes32 hash, bytes signature) public view returns (bytes4) {
+        // Short-circuit in case the hash was presigned. Optimization as performing calls
+        // and ecrecover is more expensive than an SLOAD.
+        if (isPresigned[hash]) {
```
tintinweb: Referring to [this discussion](https://forum.aragon.org/t/agent-app-arbitrary-actions-from-daos/275)

1. > A standard for contracts ‘signing messages’ (already live in 0x v2 3), is for the contract to expose a isValidSignature function that gets called for verifying whether the contract approves a given signature as its own:

      /* nitpicking alert */: I am a bit unsure about this one. the method is called `isValidSignature` but the provided signature is in some cases actually never checked. Instead the method performs a static hash lookup assuring that a certain anonymous hash is known by the Agent. It is kind of equivalent to a signature check given that only a tightly controlled role can add hashes but also works as kind of a backdoor to 'sign' arbitrary hashes (even downstream in the name of potentially configured signature providers; hash lookup takes precedence). What is the use-case for presigned hashes?

2. there is no way to remove presigned hashes from the map which might be an issue if someone accidentally "signs" a hash. How likely is the case that someone accidentally adds a wrong hash? Is adding a hash done by user interaction or in an automated fashion (contract, ...)? What is the worst case scenario if this happens (redeploy app?)? From the discussion I learned that this mechanism is intended to be irrevocable mimicking ordinary ecdsa signatures. 

3.  > 4.1.0 Designated signer is not set
  Return false unless the hash has been pre-signed (see section 4.2)
  
    spec mismatch: With the current implementation `SignatureValidator.isValidSignature` is still being called to verify the signature even though `designatedSigner` might be addr(0).

apps/agent/test/agent.js
```diff
@@ -0,0 +1,577 @@
+const Agent = artifacts.require('Agent')
+
```
Corsaire: It looks like all the tests are going through the Agent contract. There should be dedicated tests for SignatureValidator contract that will test all its public functions.

apps/agent/contracts/Agent.sol
Corsaire: @izqui @sohkai 
`isValidSignature` function encapsulates the usage of two different mechanisms of signature validation. According to the [discussion](https://forum.aragon.org/t/agent-app-arbitrary-actions-from-daos/275), it’s intended behaviour that one of those mechanisms processes irrevocable signatures (presigned) and the other one works with signatures that can be revoked. The problem is that there is no well-defined way to know whether this particular signature is revocable or not. Different types of signatures represent different types of commitments, so they should often be used for different purposes and there should be a way of differing these types of signatures. So I would suggest 2 possible solutions here:
1. Keep single signature type and work with only revocable (or irrevocable) signatures.
2. Define a transparent mechanism that will distinguish revocable signatures from irrevocable.

P.S. Right now it's possible to find out if a signature is irrevocable by looking straight into public `isPresigned` storage variable, but I'm not sure if it's transparent enough. I think there should be an explicit function with clear naming for that.

P.P.S. Since this issue should be very common, we might also consider adding it to ERC1271

apps/agent/contracts/Agent.sol
```diff
@@ -0,0 +1,157 @@
...
+        emit PresignHash(msg.sender, _hash);
+    }
+
+    function isForwarder() external pure returns (bool) {
+        return true;
+    }
+
+    function supportsInterface(bytes4 interfaceId) external pure returns (bool) {
+        return
+            interfaceId == ERC1271_INTERFACE_ID ||
```
Corsaire: ERC1271 is still a draft. Using unfinished ERC may lead to making multiple updates to the system before it stabilizes. And making serious upgrades is always a security consideration.

apps/agent/contracts/SignatureValidator.sol
```diff
@@ -0,0 +1,156 @@
...
+
+        output = new bytes(inputLength - 1);
+
+        uint256 inputPointer;
+        uint256 outputPointer;
+        assembly {
+            inputPointer := add(input, 0x21)
+            outputPointer := add(output, 0x20)
+        }
+        memcpy(outputPointer, inputPointer, inputLength);
```
Corsaire: I think there should be `memcpy(outputPointer, inputPointer, inputLength - 1);` since the length of the new array should be `inputLength - 1`

apps/agent/contracts/SignatureValidator.sol
```diff
@@ -0,0 +1,156 @@
...
+        if (v != 27 && v != 28) {
+            badSig = true;
+        }
+    }
+
+    function popFirstByte(bytes memory input) private pure returns (bytes memory output) {
+        uint256 inputLength = input.length;
+        require(inputLength > 0, ERROR_INVALID_LENGTH_POP_BYTE);
+
+        output = new bytes(inputLength - 1);
```
izqui: Early exit here if `inputLength == 1`

apps/agent/contracts/SignatureValidator.sol
```diff
@@ -0,0 +1,156 @@
...
+import "./standards/ERC1271.sol";
+
+
+library SignatureValidator {
+    enum SignatureMode {
+        Invalid, // 0x00
+        EIP712,  // 0x01
+        EthSign, // 0x02
+        ERC1271, // 0x03
+        NMode    // 0x04, to check if mode is specified, leave at the end
```
izqui: Consider checking the storage presigned hashes only if a specific mode is used

apps/agent/contracts/SignatureValidator.sol
```diff
@@ -0,0 +1,160 @@
...
+        (bool badSig, bytes32 r, bytes32 s, uint8 v) = unpackEcSig(signature);
+
+        if (badSig) {
+            return false;
+        }
+
+        return signer == ecrecover(hash, v, r, s);
+    }
+
+    function unpackEcSig(bytes memory signature) private pure returns (bool badSig, bytes32 r, bytes32 s, uint8 v) {
```
facuspagnuolo: May be interesting to have in mind [this](https://github.com/OpenZeppelin/openzeppelin-solidity/pull/1622) signature malleability case.

apps/voting/test/voting.js
```diff
@@ -223,12 +223,12 @@ contract('Voting App', accounts => {
                     assert.isTrue(isOpen, 'vote should be open')
                     assert.isFalse(isExecuted, 'vote should not be executed')
                     assert.equal(creator, holder51, 'creator should be correct')
-                    assert.equal(snapshotBlock, await getBlockNumber() - 1, 'snapshot block should be correct')
-                    assert.equal(supportRequired.toNumber(), neededSupport.toNumber(), 'required support should be app required support')
-                    assert.equal(minQuorum.toNumber(), minimumAcceptanceQuorum.toNumber(), 'min quorum should be app min quorum')
+                    assert.equal(snapshotBlock.toString(), await getBlockNumber() - 1, 'snapshot block should be correct')
+                    assert.equal(supportRequired.toString(), neededSupport.toString(), 'required support should be app required support')
+                    assert.equal(minQuorum.toString(), minimumAcceptanceQuorum.toString(), 'min quorum should be app min quorum')
```
facuspagnuolo: Another alternative could be to use `BN#eq`: 
```js
assert(minQuorum.eq(minimumAcceptanceQuorum), 'min quorum should be app min quorum')
```

apps/finance/contracts/Finance.sol
```diff
@@ -380,11 +383,11 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
     * @return success Boolean indicating whether the accounting period is the correct one (if false,
     *                 maxTransitions was surpased and another call is needed)
     */
-    function tryTransitionAccountingPeriod(uint64 _maxTransitions) public isInitialized returns (bool success) {
+    function tryTransitionAccountingPeriod(uint64 _maxTransitions) external isInitialized returns (bool success) {
```
facuspagnuolo: Given that `_tryTransitionAccountingPeriod` works with `uint256`, is there any reason to use `uint64` here?

apps/vault/contracts/Vault.sol
```diff
@@ -94,7 +97,7 @@ contract Vault is EtherTokenConstant, AragonApp, DepositableStorage {
             // Deposit is implicit in this case
             require(msg.value == _value, ERROR_VALUE_MISMATCH);
         } else {
-            require(ERC20(_token).transferFrom(msg.sender, this, _value), ERROR_TOKEN_TRANSFER_FROM_REVERTED);
+            require(ERC20(_token).safeTransferFrom(msg.sender, this, _value), ERROR_TOKEN_TRANSFER_FROM_REVERTED);
```
facuspagnuolo: Nit, `this` references to the contract instance, it's more appropiate to do `address(this)` in  this case. Same applies for the rest of the incidences referencing `this` as the address of the contract.

apps/agent/contracts/Agent.sol
```diff
@@ -0,0 +1,157 @@
...
+import "./SignatureValidator.sol";
+import "./standards/IERC165.sol";
+import "./standards/ERC1271.sol";
+
+import "@aragon/apps-vault/contracts/Vault.sol";
+
+import "@aragon/os/contracts/common/IForwarder.sol";
+
+
+contract Agent is IERC165, ERC1271Bytes, IForwarder, IsContract, Vault {
```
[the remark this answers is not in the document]
sohkai: In this particular case, do you mean adding some information as to what the app should do?

We will move towards having individual readmes in the apps (https://github.com/aragon/aragon-apps/issues/689) and perhaps that would alleviate this?

apps/agent/contracts/Agent.sol
```diff
@@ -0,0 +1,157 @@
...
+
+    string private constant ERROR_EXECUTE_ETH_NO_DATA = "AGENT_EXEC_ETH_NO_DATA";
+    string private constant ERROR_EXECUTE_TARGET_NOT_CONTRACT = "AGENT_EXEC_TARGET_NO_CONTRACT";
+    string private constant ERROR_DESIGNATED_TO_SELF = "AGENT_DESIGNATED_TO_SELF";
+
+    mapping (bytes32 => bool) public isPresigned;
+    address public designatedSigner;
+
+    event Execute(address indexed sender, address indexed target, uint256 ethValue, bytes data);
+    event PresignHash(address indexed sender, bytes32 indexed hash);
```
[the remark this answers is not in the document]
sohkai: Perhaps indexing `hash` is unnecessary, but I could see someone looking up if they (or someone else) added a specific hash as a presign before cc @izqui.

apps/agent/contracts/standards/ERC1271.sol
```diff
@@ -0,0 +1,44 @@
...
+    */
+    function isValidSignature(bytes32 _hash, bytes memory _signature) public view returns (bytes4);
+
+    function returnIsValidSignatureMagicNumber(bool isValid) internal pure returns (bytes4) {
+        return isValid ? ERC1271_RETURN_VALID_SIGNATURE : ERC1271_RETURN_INVALID_SIGNATURE;
+    }
+}
+
+
+contract ERC1271Bytes is ERC1271 {
```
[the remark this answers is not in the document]
sohkai: Mostly because `ERC1271Bytes` may become `ERC1271`; we're not sure what the exact interface will be but we've tried to cover both cases here.

apps/voting/test/voting.js
sohkai: It's actually using `Bignumber.js` since we're still on truffle@4, but I do like using the string comparison since it means less migrations in the future :)

apps/finance/contracts/Finance.sol
sohkai: Will fix in another PR.

apps/vault/contracts/Vault.sol
sohkai: Will fix in another PR.

sohkai:
Created https://github.com/aragon/aragon-apps/issues/720 to track the improvements to the Agent app.

PR #612, opened 2018-12-14
Responsive token manager

AquiGorka:
These changes make the token manager work responsively and add tabbed navigation to the mobile view.

Pending:
- [x] Action button to Icon
- [x] Animate tab change
- [x] Token manager signer panel


apps/token-manager/app/src/screens/Holders.js
```diff
@@ -64,15 +98,125 @@ class Holders extends React.Component {
...
+
+  handleSelectTab = index => {
+    this.setState({ selectedTab: index })
+  }
+}
+
+const Screen = ({
+  offset,
+  children,
+  component: Component,
```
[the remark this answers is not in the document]
AquiGorka: Yeah! For some reason I thought I wasn't able to use a capitalized prop directly.

apps/token-manager/app/src/components/HolderRow.js
```diff
@@ -81,12 +85,38 @@ class HolderRow extends React.Component {
...
+      &&& {
+        border-left-width: 1px;
+        border-right-width: 1px;
+
+        :first-child, :last-child {
+          border-radius: 3px;
+        }
+      }
+    `
+  )};
```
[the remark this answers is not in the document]
AquiGorka: Yeah, for a next iteration I'd like to move to `aragon/ui` common Components & changes introduced to all the apps.

AquiGorka:
@bpierre all comments addressed 😄

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -64,15 +99,125 @@ class Holders extends React.Component {
...
+  margin-top: 16px;
+
+  & ul {
+    border-bottom: none !important;
+  }
+  & li {
+    padding: 0 20px;
+  }
+
+  ${breakpoint('medium', `display: none`)};
```
sohkai: Rather than hiding this with css, could we just avoid rendering this component if `tabbedNavigation` is falsey?

apps/token-manager/app/src/components/SideBar.js
```diff
@@ -40,15 +40,17 @@ class SideBar extends React.Component {
   render() {
     const {
       holders,
+      shorten,
```
sohkai: `shortenAddresses` :smile:

bpierre:
Another thing: at 320px (the minimum we should support), there is not enough space to display the button:

image

We could remove the margin, but it would still be very close, and the token badge can be any size. What about shrinking the app title (and add it as a `title` attribute so it can still be accessed)? e.g.

image

We should also ensure that this doesn’t happen when the width is below the supported minimum (which is probably related to `AppView`):

image

/cc @jounih

jounih:
> Another thing: at 320px (the minimum we should support without horizontal scrolling), there is not enough space to display the button:
> 
> image
> 
> We could remove the margin, but it would still be very close, and the token badge can be any size. What about shrinking the app title (and add it as a `title` attribute so it can still be accessed)? e.g.
> 
> image
> 

I think shortening the title with ... is a good solution. I do think we should lose the margin too (between the action icon and the tag), it looks very wide for mobile.

Another solution would be to stack the title and the tag on top of each other, if they are too long to show side by side.

mobile token manager copy 2 2x


What do you think @bpierre @AquiGorka ?

bpierre:
> Another solution would be to stack the title and the tag on top of each other, if they are too long to show side by side.

Yes I like this solution even better 👍 and if we reuse this for every app, we could keep the ellipsis on the app title in case it gets too long for some apps in the future.

apps/token-manager/app/src/WindowSizeProvider.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react'
```
sohkai: 😍

bpierre:
@AquiGorka @sohkai Ready ✅

PR #614, opened 2018-12-17
Finance, Vault: Use SafeERC20

sohkai:
Requires https://github.com/aragon/aragonOS/pull/469.

Sets up the Finance and Vault apps so they are usable with all existing token interfaces.


sohkai:
> > I guess that as TokenManager is using MiniMe, we assume those tokens will be ok, right?

Yeah... that's my assumption. If you modified minime to do weird things, and expected it to work with the token manager... 🤷‍♂️. At least for now, my security model around the Token Manager is that it always uses our version of the Minime token since so much of its functionality is derived from the token itself.

Although we could sprinkle the `static` version of `balanceOf` in the Token Manager...

> I wonder if we could put those Token Mocks in some shared place to avoid repeating them so much.

Will think about this... it probably makes most sense to put it as part of the eventual "test utils" package.

> Maybe it's github, but it seems indentation got messed up with some test js files.

Good catch, fixed now!

sohkai:
Updated to use the static token utils in https://github.com/aragon/aragonOS/pull/469.

izqui:
I would love for this to be merged before we release the Actor app, so the first version already has the fix

PR #616, opened 2018-12-19
Known tokens fallback

### LINKED ISSUE #561: Finance: support tokens returning their name and symbol as 'bytes32'
For some reason, Dai uses a `bytes32` state variable for its [name and symbol](https://etherscan.io/token/0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359#readContract) 🤷‍♀️

The Finance app doesn't support this edge case and the UI looks confusing right now:

2018-11-03 at 3 17 35 pm
2018-11-03 at 3 17 44 pm

https://mainnet.aragon.org/#/ravedao.aragonid.eth/0xc5ef25d8f0678ddca01e49201cbea521ece81587

bpierre:
Builds on #615, fixes #561 

Some known tokens don’t strictly follow ERC-20 and it would be difficult to adapt to every situation. The data listed in this map is used as a fallback if either the name or the symbol can’t be determined from the contract only.

image

### DIFF
--- a/apps/finance/app/src/script.js
+++ b/apps/finance/app/src/script.js
@@ -10,6 +10,23 @@ import vaultBalanceAbi from './abi/vault-balance.json'
 import vaultGetInitializationBlockAbi from './abi/vault-getinitializationblock.json'
 import vaultEventAbi from './abi/vault-events.json'
 
+// Some known tokens don’t strictly follow ERC-20 and it would be difficult to
+// adapt to every situation. The data listed in this map is used as a fallback
+// if either the name or the symbol can’t be determined from the contract only.
+const KNOWN_TOKENS_FALLBACK = new Map([
+  [
+    '0x89d24A6b4CcB1B6fAA2625fE562bDD9a23260359',
+    { symbol: 'DAI', name: 'Dai Stablecoin v1.0', decimals: '18' },
+  ],
+])
+
+const tokenFallback = (address, fieldName) => {
+  if (!KNOWN_TOKENS_FALLBACK.has(address)) {
+    return null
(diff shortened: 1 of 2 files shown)


apps/finance/app/src/script.js
```diff
@@ -10,6 +10,23 @@ import vaultBalanceAbi from './abi/vault-balance.json'
 import vaultGetInitializationBlockAbi from './abi/vault-getinitializationblock.json'
 import vaultEventAbi from './abi/vault-events.json'
 
+// Some known tokens don’t strictly follow ERC-20 and it would be difficult to
+// adapt to every situation. The data listed in this map is used as a fallback
+// if either the name or the symbol can’t be determined from the contract only.
+const KNOWN_TOKENS_FALLBACK = new Map([
```
sohkai: I would put this in another file, so we can also use in [`Deposit`](https://github.com/aragon/aragon-apps/blob/master/apps/finance/app/src/components/NewTransfer/Deposit.js#L148).

apps/finance/app/src/script.js
```diff
@@ -318,18 +337,19 @@ function loadTokenName(tokenContract) {
         .subscribe(
           name => {
             tokenName.set(tokenContract, name)
-            resolve(name)
+            resolve(name || fallback)
           },
           () => {
             // Name is optional
-            resolve('')
+            resolve(fallback)
```
sohkai: I just realized we don't return errors through aragonAPI's RPC... so a failing call here actually just returns `undefined` in the result handler. I think leaving these error handlers here is still good though :+1:.

PR #618, opened 2019-01-07
bytes32 symbol

rperez89:
Fixes https://github.com/aragon/aragon-apps/issues/561.

Fix for the token symbol in bytes32 (DAI)

### DIFF
--- a/apps/finance/app/src/abi/token-symbol-bytes.json
+++ b/apps/finance/app/src/abi/token-symbol-bytes.json
@@ -1,0 +1,16 @@
+[
+    {
+      "constant": true,
+      "inputs": [],
+      "name": "symbol",
+      "outputs": [
+        {
+          "name": "",
+          "type": "bytes32"
+        }
+      ],
+      "payable": false,
+      "stateMutability": "view",
+      "type": "function"
+    }
+  ]
--- a/apps/finance/app/src/abi/token-symbol-bytes.json
+++ b/apps/finance/app/src/abi/token-symbol-bytes.json
@@ -1,16 +1,16 @@
 [
-    {
-      "constant": true,
-      "inputs": [],
-      "name": "symbol",
-      "outputs": [
-        {
-          "name": "",
-          "type": "bytes32"
-        }
-      ],
-      "payable": false,
-      "stateMutability": "view",
-      "type": "function"
-    }
-  ]
+  {
+    "constant": true,
+    "inputs": [],
+    "name": "symbol",
+    "outputs": [
+      {
+        "name": "",
+        "type": "bytes32"
+      }
+    ],
+    "payable": false,
+    "stateMutability": "view",
+    "type": "function"
+  }
+]
--- a/apps/finance/app/src/components/NewTransfer/Deposit.js
+++ b/apps/finance/app/src/components/NewTransfer/Deposit.js
@@ -182,9 +182,7 @@ class Deposit extends React.Component {
         tokenData.decimals = parseInt(tokenDecimals, 10)
       }
 
-      resolve({
-        ...tokenData,
-      })
+      resolve(tokenData)
     })
   }
   validateInputs({ amount, selectedToken } = {}) {
--- a/apps/finance/app/src/components/NewTransfer/Deposit.js
+++ b/apps/finance/app/src/components/NewTransfer/Deposit.js
@@ -160,10 +160,10 @@ class Deposit extends React.Component {
         tokenDataFallback(address, 'symbol', network.type) || ''
 
       const tokenData = {
-        symbol: symbolFallback,
         userBalance,
         decimals: parseInt(decimalsFallback, 10),
         loading: false,
+        symbol: symbolFallback,
       }
 
       const [tokenSymbol, tokenDecimals] = await Promise.all([
@@ -174,7 +174,7 @@ class Deposit extends React.Component {
           .toPromise(),
       ])
 
-      // no token symbol or decimals: resolve the fallback data
+      // If symbol or decimals are resolved, overwrite the fallbacks
       if (tokenSymbol) {
         tokenData.symbol = tokenSymbol
       }
--- a/apps/finance/app/src/components/NewTransfer/Deposit.js
+++ b/apps/finance/app/src/components/NewTransfer/Deposit.js
@@ -21,6 +21,7 @@ import provideNetwork from '../../lib/provideNetwork'
 import {
   ETHER_TOKEN_FAKE_ADDRESS,
   tokenDataFallback,
+  getTokenSymbol
 } from '../../lib/token-utils'
 import { addressesEqual, isAddress } from '../../lib/web3-utils'
 import { combineLatest } from '../../rxjs'
@@ -158,17 +159,23 @@ class Deposit extends React.Component {
         tokenDataFallback(address, 'decimals', network.type) || '0'
       const symbolFallback =
         tokenDataFallback(address, 'symbol', network.type) || ''
-
-      combineLatest(token.decimals(), token.symbol())
+      const tokenSymbol = getTokenSymbol(app,address)
+      if (tokenSymbol){
+        combineLatest(token.decimals(), token.symbol())
         .first()
         .subscribe(
-          ([decimals = decimalsFallback, symbol = symbolFallback]) =>
+          ([decimals = decimalsFallback, symbol]) =>
             resolve({
-              symbol,
+              symbol: tokenSymbol,
               userBalance,
               decimals: parseInt(decimals, 10),
               loading: false,
-            }),
+            })
+        )
+      }else {
+        combineLatest(token.decimals(), token.symbol())
+        .first()
+        .subscribe(
           () => {
             // Decimals and symbols are optional
             resolve({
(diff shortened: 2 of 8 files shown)


apps/finance/app/src/lib/token-utils.js
```diff
@@ -38,3 +44,30 @@ export const tokenDataFallback = (tokenAddress, fieldName, networkType) => {
   }
   return fallbacksForNetwork.get(addressWithoutChecksum)[fieldName] || null
 }
+
+export async function getTokenSymbol (app, address){
```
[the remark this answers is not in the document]
rperez89: yes the only thing that i am not sure about this is that we also use this function from Deposit.js, if importing a function from that script file is not something bad sure let's move it!

apps/finance/app/src/lib/token-utils.js
```diff
@@ -38,3 +44,30 @@ export const tokenDataFallback = (tokenAddress, fieldName, networkType) => {
...
+  const tokenAbi = [].concat(tokenBalanceOfAbi,tokenDecimalsAbi,tokenSymbolAbi)
+  const token = app.external(address,tokenAbi)
+  const tokenSymbol = await token
+    .symbol()
+    .first()
+    .toPromise()
+  if (tokenSymbol){
+    return tokenSymbol
+
+  }else{
```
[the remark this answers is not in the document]
bpierre: Oh yes sorry I should have been more explicit, yes I am only talking about removing the `else` and not the block itself, because the `if` block returns immediately.

Basically converting this kind of structure:

```js
if (a) {
  return 'a'
} else {
  /* do stuff */
  if (b) {
    return 'b'
  } else {
    return 'c'
  }
}
```

Into this (to fit the style we are using in other places):

```js
if (a) {
  return 'a'
}
/* do stuff */
if (b) {
  return 'b'
}
return 'c'
```

Or even:

```js
if (a) {
  return 'a'
}
/* do stuff */
return b ? 'b' : 'c'
```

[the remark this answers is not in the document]
rperez89: ohhh yes sure!

apps/finance/app/src/components/NewTransfer/Deposit.js
```diff
@@ -158,17 +159,23 @@ class Deposit extends React.Component {
...
-            }),
+            })
+        )
+      }else {
+        combineLatest(token.decimals(), token.symbol())
+        .first()
+        .subscribe(
           () => {
             // Decimals and symbols are optional
             resolve({
```
bpierre: It seems we could resolve directly here, without the observer?

apps/finance/app/src/components/NewTransfer/Deposit.js
```diff
@@ -158,17 +159,23 @@ class Deposit extends React.Component {
...
         tokenDataFallback(address, 'symbol', network.type) || ''
-
-      combineLatest(token.decimals(), token.symbol())
+      const tokenSymbol = getTokenSymbol(app,address)
+      if (tokenSymbol){
+        combineLatest(token.decimals(), token.symbol())
         .first()
         .subscribe(
-          ([decimals = decimalsFallback, symbol = symbolFallback]) =>
+          ([decimals = decimalsFallback, symbol]) =>
```
bpierre: We are not using `symbol` anymore, so we can probably remove it (and the `token.symbol()` call).

apps/finance/app/src/lib/token-utils.js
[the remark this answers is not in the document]
bpierre: Oh yes, I missed `Deposit.js` for some reason, sorry about that!

apps/finance/app/src/lib/token-utils.js
```diff
@@ -1,4 +1,10 @@
 import { ETHER_TOKEN_VERIFIED_ADDRESSES } from './verified-tokens'
+import { toUtf8 } from './web3-utils'
+import tokenSymbolBytes from '../abi/token-symbol-bytes.json'
+import tokenDecimalsAbi from '../abi/token-decimals.json'
+import tokenBalanceOfAbi from '../abi/token-balanceof.json'
+import tokenNameAbi from '../abi/token-name.json'
```
bpierre: We can remove this, it is not used in the file :-)

apps/finance/app/src/components/NewTransfer/Deposit.js
```diff
@@ -159,26 +160,31 @@ class Deposit extends React.Component {
...
+      const [tokenSymbol, tokenDecimals] = await Promise.all([
+        getTokenSymbol(app, address),
+        token
+          .decimals()
+          .first()
+          .toPromise(),
+      ])
+
+      // no token symbol or decimals: resolve the fallback data
+      if (!tokenSymbol || !tokenDecimals) {
```
sohkai: I think we should just overwrite `tokenData`'s fields here, if either of the two exist:

```js
if (tokenSymbol) {
  tokenData.symbol = tokenSymbol
}
if (tokenDecimals) {
  tokenData.decimals = parseInt(tokenDecimals, 10)
}
```

Otherwise if we get one but not the other, we'll end up overwriting the other (from its fallback) by accident.

apps/finance/app/src/lib/token-utils.js
```diff
@@ -38,3 +43,24 @@ export const tokenDataFallback = (tokenAddress, fieldName, networkType) => {
   }
   return fallbacksForNetwork.get(addressWithoutChecksum)[fieldName] || null
 }
+
+export async function getTokenSymbol(app, address) {
+  let tokenAbi = [].concat(tokenBalanceOfAbi, tokenDecimalsAbi, tokenSymbolAbi)
```
sohkai: Technically we only need `[tokenSymbolAbi]` here and `[tokenSymbolBytesAbi]` later.

apps/finance/app/src/lib/token-utils.js
```diff
@@ -38,3 +43,24 @@ export const tokenDataFallback = (tokenAddress, fieldName, networkType) => {
...
+  let token = app.external(address, tokenAbi)
+  let tokenSymbol = await token
```
sohkai: We should probably add comments here that we currently don't need a `try/catch` block here because aragon.js doesn't throw the error back at us.

That being said, I was thinking we could program a bit defensively and use a `try/catch` anyway so it won't break when aragon.js actually does return errors:

```js
let tokenSymbol

try {
  // try string
  tokenSymbol = await token.symbol()
  ...
  // Note: aragon.js currently doesn't throw errors when a call fails (returns undefined instead) so we manually throw
  if (!tokenSymbol) {
     throw new Error()
  }
} catch () {
  // try bytes32
  try {
    tokenSymbol = await token.symbol()
    ...
    tokenSymbol ? toUtf8(tokenSymbol) : null
  } catch () {
     // ignore
  }
}

return tokenSymbol
```

apps/finance/app/src/abi/token-symbol-bytes.json
```diff
@@ -0,0 +1,16 @@
+[
+    {
+      "constant": true,
+      "inputs": [],
+      "name": "symbol",
```
izqui: We should do token name as well in this PR, as `ds-token` also uses `bytes32` for `token.name()`

apps/finance/app/src/lib/token-utils.js
rperez89: So instead of 
`let tokenAbi = [].concat(tokenBalanceOfAbi, tokenDecimalsAbi, tokenSymbolAbi)`
this:
`let token = app.external(address, tokenSymbolAbi)` ?

sohkai: The API I believe still requires the ABI to be input as an array I think, so it'd be `let token = app.external(address, [tokenSymbolAbi])`

apps/finance/app/src/components/NewTransfer/Deposit.js
```diff
@@ -176,14 +176,15 @@ class Deposit extends React.Component {
...
+      }
+      if (tokenDecimals) {
+        tokenData.decimals = parseInt(tokenDecimals, 10)
       }
 
       resolve({
-        ...tokenData,
-        symbol: tokenSymbol,
-        decimals: parseInt(tokenDecimals, 10),
+        ...tokenData,
```
sohkai: No need to allocation a new object here, we can just `resolve(tokenData)`

apps/finance/app/src/lib/token-utils.js
```diff
@@ -62,3 +62,22 @@ export async function getTokenSymbol(app, address) {
...
+export async function getTokenName(app, address) {
+  let token = app.external(address, tokenNameAbi)
+  let tokenName = await token
+    .name()
+    .first()
+    .toPromise()
+  if (tokenName) {
+    return tokenName
+  }
+  // Now we try with symbol as bytes32
```
sohkai: with **name** 😄

apps/finance/app/src/lib/token-utils.js
rperez89: mmm strange i tested it and it didn't  fail... But maybe it was something cached in my browser..

sohkai:
@rperez89 Oops, I missed that the ABIs we export are already arrays so we don't need to add more to them. I've reverted that in 05f95f1 :).

PR #619, opened 2019-01-07
Responsive finance

AquiGorka:
Pending:

- [x] Toggle animate Filters
- [x] New Transfer Token dropdown spans wider than available space


apps/finance/app/src/components/ToggleFiltersButton.js
```diff
@@ -0,0 +1,42 @@
...
+    border: 2px solid ${theme.accent};
+  }
+  &:active {
+    border: none;
+  }
+`
+
+export default props => (
+  <StyledButton {...props}>
+    <svg
```
bpierre: Just as a note, we should move this icon in Aragon UI (but we can keep it like this until it gets released).

bpierre: Optimized version of this SVG (not tested, I just ran it through [`svgr`](https://github.com/smooth-code/svgr)):

```jsx
<svg width="16" height="14">
  <path
    d="M10.421 2.667H.631C.275 2.667 0 2.377 0 2c0-.378.274-.667.632-.667h9.79C10.694.556 11.388 0 12.21 0c.822 0 1.537.556 1.79 1.333h1.368c.358 0 .632.29.632.667 0 .378-.274.667-.632.667H14C13.726 3.444 13.032 4 12.21 4c-.82 0-1.536-.556-1.789-1.333zM16 7c0 .378-.274.667-.632.667h-9.79C5.327 8.444 4.612 9 3.79 9 2.97 9 2.274 8.444 2 7.667H.632C.274 7.667 0 7.377 0 7c0-.378.274-.667.632-.667H2C2.253 5.556 2.968 5 3.79 5c.82 0 1.515.556 1.789 1.333h9.79c.357 0 .631.29.631.667zm-.632 4.333H9.79C9.537 10.556 8.821 10 8 10s-1.516.556-1.79 1.333H.633c-.358 0-.632.29-.632.667 0 .378.274.667.632.667H6.21C6.463 13.444 7.179 14 8 14s1.516-.556 1.79-1.333h5.578c.358 0 .632-.29.632-.667 0-.378-.274-.667-.632-.667z"
    fill="#000"
  />
</svg>
```

apps/finance/app/src/components/ToggleFiltersButton.js
```diff
@@ -0,0 +1,42 @@
...
+  width: 14px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  padding: 0;
+  cursor: pointer;
+  outline: none;
+
+  &:focus {
+    border: 2px solid ${theme.accent};
```
bpierre: We should use `outline` here so that it doesn’t impact the box:

image

And also remove the inner border in Firefox:

```css
&::-moz-focus-inner {
  border: 0;
}
```

apps/finance/app/src/components/ToggleFiltersButton.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+  &:active {
+    border: none;
```
bpierre: Just as a note, we should move to [`FocusVisible`](https://github.com/aragon/aragon-ui/blob/master/src/components/FocusVisible/FocusVisible.js) in the future.

apps/finance/app/src/components/Transfers.js
```diff
@@ -101,27 +111,55 @@ class Transfers extends React.Component {
...
+            <BreakPoint to="medium">
+              <ToggleFiltersButton
+                title="Toggle Filters"
+                onClick={this.handleToggleFiltersClick}
+              />
+            </BreakPoint>
+          </Title>
+          <Spring
+            native
+            config={springs.clamp}
```
bpierre: `springs.clamp` doesn’t exist 🙈

AquiGorka:
> At certain sizes, we can see the button for the menu panel while it is still visible (maybe we could use GetWindowSize only for this?):

This is the issue I was referring to when I said the inner _iframe_ does not consider the _parent_'s dimensions.

bpierre:
> This is the issue I was referring to when I said the inner iframe does not consider the parent's dimensions.

Yes which is what we want for everything but this button, so I was wondering if we could use the window size for this specifically, to have it in sync with the parent? We could also send this information from the parent, using `postMessage`.

> I've been keeping the same padding/margin for all headers to the same spacing.

We should probably keep it like this for now then, and update them all at the same time later 👍

AquiGorka:
@bpierre 

- [x] The `IdentityBadge` should be shortened earlier, so that the table doesn’t get cut
- [x] The labels should probably be wrapped so that they are not 1fr (50%) width
- [x] We should still have a minimum width (= horizontal scrollbar) when the viewport gets to a certain point (320px)
- [x] The toggle transition has a little jump at the end, and should use something like springs.smooth or springs.swift as a config
- [x] The buttons in the header should be closer to the edges, and I think it would be nice to have the entire area clickable (from top to bottom, and including the padding to the edge), especially for touch screens (agreed to come back to this and review all the spacing).

>At certain sizes, we can see the button for the menu panel while it is still visible (maybe we could use GetWindowSize only for this?)

This is super hard when this happens the iframe's width is < 768 and there is no way to know if the menu panel exists and the window's width > 768. If it is alright with you I'll post this as an issue as it something that concerns all the iframe apps.

bpierre:
@AquiGorka Awesome!

> If it is alright with you I'll post this as an issue as it something that concerns all the iframe apps.

👍

apps/finance/app/src/App.js
```diff
@@ -150,39 +176,70 @@ class App extends React.Component {
...
+  </React.Fragment>
+)
+
+const AppBarContainer = styled.div`
+  display: flex;
+  width: 100%;
+  height: 100%;
+  justify-content: space-between;
+  align-items: center;
+  justify-content: safe;
```
sohkai: Hmm, does this need to be with an [alignment keyword](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content#Formal_syntax)?

apps/finance/app/src/components/Transfers.js
```diff
@@ -101,27 +111,52 @@ class Transfers extends React.Component {
...
+                <Filters
+                  style={{
+                    overflow: progress.interpolate(v =>
+                      v === 1 ? 'unset' : 'hidden'
+                    ),
+                    height: progress.interpolate(v => `${80 * v}px`),
+                  }}
+                >
+                  <FilterLabel>
+                    <Label>Token:</Label>
```
sohkai: Let's remove the `:`

apps/finance/app/src/components/Transfers.js
```diff
@@ -101,27 +111,52 @@ class Transfers extends React.Component {
...
+                    <DropDown
+                      items={['All', ...symbols]}
+                      active={selectedToken}
+                      onChange={this.handleTokenChange}
+                    />
+                  </FilterLabel>
+                  <FilterLabel>
+                    <Label>Transfer type:</Label>
```
sohkai: And the `:` here too 😄

apps/finance/app/src/components/Transfers.js
```diff
@@ -160,7 +197,7 @@ class Transfers extends React.Component {
                         key={transfer.transactionHash}
                         token={tokenDetails[toChecksumAddress(transfer.token)]}
                         transaction={transfer}
-                        wideMode={width > 800}
+                        wideMode={width > 768 + 219}
```
sohkai: Wondering if we should use a constant instead of `219` (is it the width of the side panel)?

AquiGorka: I did this on purpose so we would not miss it.

apps/finance/app/src/components/Transfers.js
```diff
@@ -181,37 +218,74 @@ class Transfers extends React.Component {
...
-  margin-bottom: 20px;
+  margin: 20px 20px 10px 20px;
   font-weight: 600;
+  display: flex;
+  justify-content: space-between;
+
+  ${breakpoint(
+    'medium',
+    `
+      margin: 10px 30px 20px 0;
```
sohkai: Hmm, is the right margin here necessary?

Actually, I was thinking of increasing the top margin to align it with the filters on larger screens. Thoughts @bpierre @AquiGorka @jounih?

Currently:

image

With more top margin (`30px`):

image

apps/finance/app/src/components/Balances.js
```diff
@@ -92,24 +92,53 @@ const ScrollView = styled.div`
...
 
 const List = styled.ul`
-  display: flex;
-  list-style: none;
-  padding: 0 10px;
+  ${breakpoint(
+    'medium',
+    `
+      display: flex;
+      list-style: none;
```
sohkai: Should the `list-style: none` be on the small breakpoint too?

apps/finance/app/src/components/TransferRow.js
```diff
@@ -116,6 +176,70 @@ class TransferRow extends React.Component {
...
 
+const Wrap = styled.div`
+  display: inline-flex;
+
+  ${breakpoint(
+    'medium',
+    `
+      display: block;
+    `
+  )};
```
sohkai: Do we need this breakpoint definition if `Wrap` is only used in `renderSmallScreens()`?

apps/finance/app/src/components/Transfers.js
AquiGorka: Can we push this for the backlog?

bpierre: Proposed fix: https://github.com/aragon/aragon-apps/pull/767

PR #625, opened 2019-01-10
Agent: Allow execute to send Eth

jvluso:
I think this is important for a few reasons

1. It will make the Aragon Keyring simpler
2. Preventing execute from just sending eth doesn't add any security - functions can be created to sidestep this. It also might provide a false sense of security.
3. The validation costs gas


izqui:
The reasons make sense, need more time to think about it though

sohkai:
Merging into https://github.com/aragon/aragon-apps/pull/651, as that adds the tests for this change.

PR #627, opened 2019-01-14
Fixed issue #582.

### LINKED ISSUE #582: Change "New vote" to "New poll"
In the voting app the button says "New vote" and that "Votes are informative". But the binding votes are also called "votes", so calling these the same thing while defining them differently is confusing.

What is called a "vote" in the "New vote" panel should really be called a "poll", since it is an informative question with a simple "yes" or "no" answer.

screenshot_2018-11-21 aragon

dev1644:
Renamed `New Vote` to `New Poll`, closes #582.

### DIFF
--- a/apps/voting/app/src/App.js
+++ b/apps/voting/app/src/App.js
@@ -228,7 +228,7 @@ class App extends React.Component {
                 title="Voting"
                 endContent={
                   <Button mode="strong" onClick={this.handleCreateVoteOpen}>
-                    New Vote
+                    New Poll
                   </Button>
                 }
               />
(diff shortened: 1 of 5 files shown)


sohkai:
Hi @dev1644, could you `npm install` and run the prettier fixer on the changes?

In any case, pinging @aragon/design for this as it's not something we've discussed we wanted to change (given that the app is called "Voting", perhaps it makes sense to leave it at "New Vote" even if its a subjective, non-binding vote).

dev1644:
So, should I revert the changes?

sohkai:
@dev1644 We had a discussion about how to handle copy changes like this, and in the future we'll have better copy guidelines to make quicker decisions on suggestions like this. Until then we're putting this in the holding tank, and I'll close this PR for now.

Thanks for contributing though! There are some more [good first issues](https://github.com/aragon/aragon/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22good+first+issue%22) available on aragon/aragon :).

Issue #636, opened 2019-01-25
Use blockExplorerUrl() from @aragon/ui


sohkai:
Following https://github.com/aragon/aragon/issues/457, we should upgrade all the apps to use `blockExplorerUrl()` instead of `makeEtherscanUrl()`.

sohkai:
Fixed by #639.

PR #639, opened 2019-01-28
feat(): replaced `etherscanBaseUrl` with `blockExplorerUrl`

MarkGeeRomano:
Addresses #636 

- Left the `survey` app alone, since it's not using a version of UI that has the `blockExplorerUrl` function
- There's quite a few evaluations of `etherscanBaseUrl` to conditionally render components. I switched this to just `network.type`


MarkGeeRomano:
> It's better to conditionally render based on if we actually have the URL or not (you can imagine where we might have a valid type but the block explorer doesn't support it or etc.).

Gotcha. Updated!

sohkai:
@MarkGeeRomano Thanks Mark! I noticed this bit of the network tooling wasn't used in the Survey app anyway, so I removed it from that app in https://github.com/aragon/aragon-apps/pull/682 :)

PR #648, opened 2019-02-06
Voting: fix snapshot block for new vote

sohkai:
Testing this is a bit contrived, but it does fail the previous contract version.

### DIFF
--- a/apps/voting/contracts/test/mocks/VotingMock.sol
+++ b/apps/voting/contracts/test/mocks/VotingMock.sol
@@ -20,4 +20,16 @@ contract VotingMock is Voting {
     function isValuePct(uint256 _value, uint256 _total, uint256 _pct) external pure returns (bool) {
         return _isValuePct(_value, _total, _pct);
     }
+
+    // Create a token and a vote in the same transaction to test snapshot block valuess are correct
+    function newTokenAndVote(address _holder, uint256 _tokenAmount, string _metadata)
+        external
+        returns (uint256 voteId)
+    {
+        token.generateTokens(_holder, _tokenAmount);
+
+        bytes memory noScript = new bytes(0);
+        voteId = _newVote(noScript, _metadata, false, false);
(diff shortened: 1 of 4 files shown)


apps/voting/contracts/test/mocks/VotingMock.sol
```diff
@@ -20,4 +20,16 @@ contract VotingMock is Voting {
     function isValuePct(uint256 _value, uint256 _total, uint256 _pct) external pure returns (bool) {
         return _isValuePct(_value, _total, _pct);
     }
+
+    // Create a token and a vote in the same transaction to test snapshot block valuess are correct
```
bingen: Besides the minor typo in "valuess", "Create a token" sounds confusing to me, I'd rather say something like "Mint tokens" (and create a vote)

apps/voting/test/voting.js
```diff
@@ -495,6 +495,47 @@ contract('Voting App', accounts => {
...
+        it('uses the correct snapshot value if tokens are minted in the same block', async () => {
+            // Create vote and generate some tokens in the same transaction
+            // Requires the voting mock to be the token's owner
+            await token.changeController(voting.address)
+            const voteId = createdVoteId(await voting.newTokenAndVote(holder2, 1, 'metadata'))
+
+            const [isOpen, isExecuted, startDate, snapshotBlock, supportRequired, minQuorum, y, n, votingPower, execScript] = await voting.getVote(voteId)
+
+            assert.equal(snapshotBlock.toString(), await getBlockNumber() - 1, 'snapshot block should be correct')
+            assert.equal(votingPower.toString(), (await token.totalSupplyAt(snapshotBlock)).toString(), 'total voters should be correct')
```
bingen: We could even check that those values match the expected total number of tokens (3, I guess)

sohkai: Actually 2, since it's the block before, but 👍.

PR #651, opened 2019-02-07
Agent: parameterize execution tests for ETH deposit amount

sohkai:
Adds a few more tests for the Actor app to invoke fallback functionality, like an EOA would be able to do, and then parameterizes the `execute()` tests for deposit amounts.

Note that this does rely on https://github.com/aragon/aragon-apps/pull/625 for all the tests to pass, which I think is a reasonable change to `execute()`:

- Inherited `Vault.transfer()` is gas-limited and safe from reentry, and does not allow a complex contract interaction to happen in the destination's fallback function
- `Actor.execute()` can already invoke contract functionality with ETH, which, in terms of security, would be a superset of allowing just the fallback function with ETH


izqui:
Needs updating given the rename

sohkai:
@izqui Does this mean we should also merge https://github.com/aragon/aragon-apps/pull/625?

izqui:
TBH i am still undecided, what's your take?

sohkai:
> Note that this does rely on #625 for all the tests to pass, which I think is a reasonable change to `execute()`:

> * Inherited `Vault.transfer()` is gas-limited and safe from reentry, and does not allow a complex contract interaction to happen in the destination's fallback function
> * `Actor.execute()` can already invoke contract functionality with ETH, which, in terms of security, would be a superset of allowing just the fallback function with ETH

future-apps/actor/contracts/test/mocks/DestinationMock.sol
```diff
@@ -2,9 +2,16 @@ pragma solidity 0.4.24;
 
 
 contract DestinationMock {
-    uint256 test;
+    bool public expensiveFallback;
+    uint256 public counter;
+
+    constructor(bool _expensiveFallback) {
```
bingen: Although it's just a test, it wouldn't hurt to add `public` here to avoid a warning.

future-apps/actor/contracts/test/mocks/DestinationMock.sol
```diff
@@ -2,9 +2,16 @@ pragma solidity 0.4.24;
...
+        expensiveFallback = _expensiveFallback;
+    }
 
     function () external payable {
-        test = test + 1;
+        if (expensiveFallback) {
+            counter = counter + 1;
```
bingen: This is trying to break through the 2300 limit, right? We may put a comment here.

future-apps/actor/test/actor.js
```diff
@@ -81,118 +82,121 @@ contract('Actor app', (accounts) => {
...
+          const nonContract = accounts[8] // random account
+          const randomData = '0x12345678'
+          const noData = '0x'
+
+          await assertRevert(() =>
+            actor.execute(nonContract, depositAmount, randomData, { from: executor })
+          )
+
+          await assertRevert(() =>
+            actor.execute(nonContract, depositAmount, noData, { from: executor })
```
bingen: Why is this failing if `require(isContract(_target))` was removed?

sohkai: Good catch, fixed now in 7c68809.

PR #652, opened 2019-02-08
control apps frontend installation via environment variable

### LINKED ISSUE #602: Fix CI
`lerna bootstrap` is causing the CI to fail because it takes too long to install.

The best way to solve this right now is likely to have a `install:contracts` command that ignores each app's frontend installation as we don't currently test them. This could be done via a separate install script that checks if a given env var is on (e.g. `INSTALL_FRONTENDS`) before going ahead and running `npm install` in the app's dir.

danielcaballero:
Fixes #602

Please note that it **requires '@aragon/apps-shared-scripts' to be published before merging**.

I have tested the changes in here https://travis-ci.org/danielcaballero/aragon-apps/builds/490015186 and, as expected, the build times have decreased notably.


shared/scripts/prepare.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
sohkai: Let's use a node script for this instead, so that it's more portable.

You can see an example of one in https://github.com/aragon/aragon/blob/master/scripts/start.

.travis.yml
```diff
@@ -20,19 +20,22 @@ notifications:
...
-  - TASK=coverage:voting
-  - TASK=test:shared
+  global:
+    - INSTALL_FRONTEND=false
+  matrix:
+    - TASK=lint
+    - TASK=test:finance
+    - TASK=test:survey
+    - TASK=test:token-manager
+    - TASK=test:token-manager:app INSTALL_FRONTEND=true
```
sohkai: I think this can be removed?

shared/scripts/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "name": "@aragon/apps-shared-scripts",
+  "version": "1.0.0",
+  "description": "",
+  "scripts": {
+    "test": "echo \"No tests\""
+  }
```
sohkai: At this point, I think we can keep this package `private` until we absolutely need to make it public (e.g. for the apps to be `npm install`ed correctly).

shared/scripts/prepare.sh
danielcaballero: done

.travis.yml
danielcaballero: `INSTALL_FRONTEND` needs to be `true` in this case as `test:token-manager:app` actually does test the frontend part (if you don't install the frontend then you get the error https://travis-ci.org/danielcaballero/aragon-apps/jobs/489963180)

Unless you meant removing this line all together? Let me know.

shared/scripts/package.json
danielcaballero: done

.travis.yml
sohkai: Ahh gotcha :+1:. Forgot there were tests in just the Token Manager's frontend.

sohkai:
Turns out there's a `lerna bootstrap --no-scripts` option that turns off all npm scripts, but I think having an environment variable for this is better in case we ever need to use more scripts in the future.

sohkai:
@danielcaballero I've pushed a few commits to upgrade lerna, and only ignore the apps on CI environments. Once CI confirms this works, I'll merge.

Thank you for the contribution ❤️ !

danielcaballero:
> @danielcaballero I've pushed a few commits to upgrade lerna, and only ignore the apps on CI environments. Once CI confirms this works, I'll merge.
> 
> Thank you for the contribution ❤️ !

Happy to help!

PR #657, opened 2019-02-08
Payroll: Add hard cap for allowed tokens, to avoid OOG attacks

### LINKED ISSUE #649: Payroll: ALLOWED_TOKENS_MANAGER_ROLE can lock the app with an OOG attack
In both `determineAllocation` and `payday` the `allowedTokens` array is iterated in a for loop. Someone with the `ALLOWED_TOKENS_MANAGER_ROLE` can add a large enough number of tokens that could cause these functions to fail due to OOG.

Mitigation options:
1. Hardcode an arbitrary maximum number of allowed tokens.
2. Allow users of these functions to provide the token addresses relevant to them, so there is no need to iterate over the array. This wouldn't need to be the default function that is normally used, but we should have a safe version of these functions that doesn't iterate the array.

bingen:
Fixes #649.

One thing to take into account is that right now is not possible to remove tokens from the allowed list.

### DIFF
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -33,12 +33,14 @@ contract Payroll is EtherTokenConstant, IsContract, AragonApp { //, IForwarder {
 
     uint128 internal constant ONE = 10 ** 18; // 10^18 is considered 1 in the price feed to allow for decimal calculations
     uint64 internal constant MAX_UINT64 = uint64(-1);
+    uint8 internal constant MAX_ALLOWED_TOKENS = 25; // payday for loop is spending ~220k per token
     uint256 internal constant MAX_ACCRUED_VALUE = 2**128;
 
     string private constant ERROR_NO_EMPLOYEE = "PAYROLL_NO_EMPLOYEE";
     string private constant ERROR_EMPLOYEE_DOES_NOT_MATCH = "PAYROLL_EMPLOYEE_DOES_NOT_MATCH";
     string private constant ERROR_FINANCE_NOT_CONTRACT = "PAYROLL_FINANCE_NOT_CONTRACT";
     string private constant ERROR_TOKEN_ALREADY_ALLOWED = "PAYROLL_TOKEN_ALREADY_ALLOWED";
+    string private constant ERROR_MAX_ALLOWED_TOKENS = "PAYROLL_MAX_ALLOWED_TOKENS";
     string private constant ERROR_ACCRUED_VALUE_TOO_BIG = "PAYROLL_ACCRUED_VALUE_TOO_BIG";
     string private constant ERROR_TOKEN_ALLOCATION_MISMATCH = "PAYROLL_TOKEN_ALLOCATION_MISMATCH";
     string private constant ERROR_NO_ALLOWED_TOKEN = "PAYROLL_NO_ALLOWED_TOKEN";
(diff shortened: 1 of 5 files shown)


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -33,12 +33,14 @@ contract Payroll is EtherTokenConstant, IsContract, AragonApp { //, IForwarder {
 
     uint128 internal constant ONE = 10 ** 18; // 10^18 is considered 1 in the price feed to allow for decimal calculations
     uint64 internal constant MAX_UINT64 = uint64(-1);
+    uint8 internal constant MAX_ALLOWED_TOKENS = 25; // payday for loop is spending ~220k per token
```
izqui:
```suggestion
    uint8 internal constant MAX_ALLOWED_TOKENS = 25; // for loop in `payday()` uses ~220k gas per available token
```

future-apps/payroll/test/payroll_allowed_tokens.js
```diff
@@ -0,0 +1,95 @@
...
+
+    const startDate = parseInt(await payroll.getTimestampPublic.call(), 10) - 2628005 // now minus 1/12 year
+    // add employee
+    const receipt = await payroll.addEmployee(employee, SALARY, "Kakaroto", 'Saiyajin', startDate)
+  })
+
+  it('fails adding one more token', async () => {
+    const erc20Token = await deployErc20TokenAndDeposit(owner, finance, vault, 'Extra token', ERC20_TOKEN_DECIMALS)
+
+    return assertRevert(async () => {
```
izqui: No need to be async, it can just be: `return assertRevert(() => payroll.addAllowedToken(erc20Token.address))` with proper indentation

future-apps/payroll/test/payroll_allowed_tokens.js
```diff
@@ -0,0 +1,95 @@
...
+
+  let erc20Tokens = []
+  const ERC20_TOKEN_DECIMALS = 18
+  const SALARY = (new web3.BigNumber(1)).times(10 ** Math.max(ERC20_TOKEN_DECIMALS - 4, 1))
+  const RATE_EXPIRY_TIME = 1000
+  const NOW_MOCK = new Date().getTime()
+  let ETH
+  let MAX_ALLOWED_TOKENS
+
+  const MAX_GAS_USED = 7e6
```
izqui:
```suggestion
  const MAX_GAS_USED = 6e6
```

izqui: Let's play it a bit safer, I don't like functions that need more than 80% of the current block gas limit

future-apps/payroll/test/payroll_allowed_tokens.js
```diff
@@ -0,0 +1,95 @@
...
+    )
+
+    await payroll.mockSetTimestamp(NOW_MOCK)
+
+    // adds allowed tokens
+    await addAllowedTokens(payroll, erc20Tokens)
+
+    const startDate = parseInt(await payroll.getTimestampPublic.call(), 10) - 2628005 // now minus 1/12 year
+    // add employee
+    const receipt = await payroll.addEmployee(employee, SALARY, "Kakaroto", 'Saiyajin', startDate)
```
sohkai: We could add an assert here that we've reached the max number of tokens allowed

future-apps/payroll/test/payroll_allowed_tokens.js
sohkai: I've been thinking about this recently and my recommendation is that it's probably best as

```
await assertRevert(async () => {
  await ...
})
```

Since the chances of screwing up later (if we add more statements to the test) are lower if we're explicit about the `async` and `await`s.

future-apps/payroll/test/payroll_allowed_tokens.js
```diff
@@ -0,0 +1,95 @@
...
+      finance,
+      erc20Tokens[0],
+      priceFeed,
+      RATE_EXPIRY_TIME
+    )
+
+    await payroll.mockSetTimestamp(NOW_MOCK)
+
+    // adds allowed tokens
+    await addAllowedTokens(payroll, erc20Tokens)
```
sohkai: Wasn't immediately obvious that this brought the allowed tokens all the way to the max. Potentially we could add a length getter to the test mock that would let us check the current `allowTokens` length against the max value.

future-apps/payroll/test/payroll_allowed_tokens.js
izqui: That makes sense

PR #659, opened 2019-02-11
(Voting) more descriptive empty state card

### LINKED ISSUE #245: Language suggestions for "No data" use cases of Aragon apps
Right now for the Aragon apps, when an object does not exist it says "Nothing here". I suggest to improve the terminology to be more descriptive in relation to what app you are looking at.

Here is an example of a suggested improvement for the Voting app:

image

The second option would be to just have a single line of text using the same style (instead of one line being bold/blue and the other black.

For example: 

image

wslyvh:
Fixes #245.

Updated the empty state card for the voting App with more descriptive terminology as suggested in #245.

Will update the other Apps accordingly, once this one is approved.

### DIFF
--- a/apps/voting/app/src/screens/EmptyState.js
+++ b/apps/voting/app/src/screens/EmptyState.js
@@ -6,8 +6,8 @@ import emptyIcon from '../assets/empty-card-icon.svg'
 const EmptyState = ({ onActivate }) => (
   <Main>
     <EmptyStateCard
-      title="Nothing here."
-      text="Create a new vote to start using the app."
+      title="There are no votes yet"
+      text="Add a vote to your organization to get started."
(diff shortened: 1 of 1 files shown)


dizzypaty:
Hi @wslyvh, thank you so much for your contribution! ✨ It's looking great and consistent throughout the different apps.

These onboarding cards should encourage interaction, nudging users to take the first key action. The copy used in the card should match the action that will be performed when users interact with the primary button. In this case users can create new votes, not add them to the organization.

The change you proposed for the title copy is perfect but my suggestion for the text in this scenario is: “Create a new vote to get started."

What do you think?

wslyvh:
Makes sense. Updated

PR #663, opened 2019-02-11
(Finance) more descriptive empty state card

### LINKED ISSUE #661: More descriptive Empty card state for the Finance App
Similar to #245 and #660 but for the Finance App.

image

Suggestion is to change text to
Title="There are no funds yet"
Text="Add funds to your organization to get started."

And changing the button text to match with the SidePanel 'New Transfer'.
Button="New Transfer"

wslyvh:
Fixes #661.

Updated the empty state card for the Token Manager with more descriptive terminology as suggested in #661.


dizzypaty:
Following the same reasoning as #659, my suggestion for the text in this scenario is: 
text=“Create a new transfer to get started."

And the change you proposed for the title and button label is perfect!

wslyvh:
Updated.

PR #666, opened 2019-02-13
Use the Viewport component

bpierre:
- Remove uses of BreakPoint.
- Remove GetWindowSize.
- Remove WindowSizeProvider.
- Moved the common apps structure into a new component, [AppLayout](https://github.com/aragon/aragon-apps/pull/666/files#diff-a5f61ff0a88b4aaa17a7a3cfcf42537e), which is duplicated in every app for now.
- Moved the main action icons into a icon components.


apps/voting/app/src/App.js
```diff
@@ -23,6 +11,8 @@ import NewVotePanelContent from './components/NewVotePanelContent'
 import AutoLink from './components/AutoLink'
 import MenuButton from './components/MenuButton/MenuButton'
 import NewVoteButton from './components/NewVoteButton/NewVoteButton'
```
AquiGorka: This shouldn't be here.

apps/finance/app/src/App.js
```diff
@@ -125,137 +110,61 @@ class App extends React.Component {
...
-                    title="Add funds to your organization"
-                    text="There are no funds yet - add funds easily"
-                    actionText="Add funds"
-                    onActivate={this.handleNewTransferOpen}
-                  />
-                </EmptyScreen>
-              )}
-            </AppView>
-            <SidePanel
+      <div css="min-width: 320px">
```
AquiGorka: Should this be part of `AppLayout`?

apps/finance/app/src/components/AppLayout.js
```diff
@@ -0,0 +1,104 @@
...
+        <AppView
+          padding={below('medium') ? smallViewPadding : 30}
+          appBar={
+            <AppBar>
+              <AppBarContainer
+                style={{ padding: below('medium') ? '0' : '0 30px' }}
+              >
+                <Title>
+                  {below('medium') && <MenuButton onClick={onMenuOpen} />}
+                  <TitleLabel>{title}</TitleLabel>
```
AquiGorka: How could this work for an app like `Permissions` that has a `NavigationBar`?

apps/finance/app/src/components/AppLayout.js
```diff
@@ -0,0 +1,104 @@
...
+                </Title>
+                {mainButton &&
+                  (below('medium') ? (
+                    <ButtonIcon
+                      onClick={mainButton.onClick}
+                      title={mainButton.label}
+                      css={`
+                        width: auto;
+                        height: 100%;
+                        padding: 0 20px 0 10px;
```
AquiGorka: `padding: 0 20px`?

apps/finance/app/src/components/AppLayout.js
```diff
@@ -0,0 +1,104 @@
...
+  align-items: center;
+  height: 100%;
+`
+
+const TitleLabel = styled.span`
+  flex: 0 1 auto;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+  margin-right: 10px;
```
AquiGorka: Why is this needed? There nothing after the finance title

apps/finance/app/src/components/AppLayout.js
```diff
@@ -0,0 +1,104 @@
...
+  ${font({ size: 'xxlarge' })};
+`
+
+const AppBarContainer = styled.div`
+  display: flex;
+  width: 100%;
+  height: 100%;
+  justify-content: space-between;
+  align-items: center;
+  justify-content: safe;
```
AquiGorka: There's a duplicate declaration for `justify-content` (I had already spotted this from my changes)

apps/finance/app/src/components/TransferRow.js
```diff
@@ -48,136 +48,110 @@ class TransferRow extends React.Component {
...
-      </React.Fragment>
-    )
-  }
-
-  renderSmallScreens() {
     const {
-      network: { type },
-      token: { decimals, symbol },
+      network,
+      token,
```
AquiGorka: `token: { decimals, symbol },`?

apps/token-manager/app/src/App.js
```diff
@@ -124,134 +111,66 @@ class App extends React.Component {
...
-              </Main>
-            </PublicUrl.Provider>
-          </ToastHub>
-        </WindowSizeProvider>
-      </Root.Provider>
+      <div css="min-width: 320px">
+        <Main>
+          <AppLayout
+            title="Token Manager"
+            afterTitle={tokenSymbol && <Badge.App>{tokenSymbol}</Badge.App>}
```
AquiGorka: Maybe `title` accepts an element/node instead of using _after_?

apps/voting/app/src/components/MenuButton/MenuButton.js
```diff
@@ -1,31 +1,17 @@
...
-
-  &:focus {
-    border: 2px solid ${theme.accent};
-  }
-  &:active {
-    border: none;
-  }
-`
-
 export default props => (
```
AquiGorka: Why didn't you use this approach for all the apps?

apps/finance/app/src/App.js
bpierre: I wanted to, but `SidePanel` need the `min-width` too 😔. A solution could be to have a `sidePanels` prop on `AppLayout`, but it would make the panels appear before the content, so I left it like this.

apps/finance/app/src/components/AppLayout.js
bpierre: This component is only meant to be used on apps that don’t use `NavigationBar` for now. I’d like to think a bit about `AppView`, `NavigationBar` and this component, to see how we can make the API a bit nicer for users (and us).

apps/finance/app/src/components/AppLayout.js
bpierre: It really should be `10px` on both sides, the `20px` on the right side is to reach the right side of the screen, the same way `height: 100%` reaches the top and bottom of the header. This is to make the touch target bigger.

apps/finance/app/src/components/AppLayout.js
bpierre: We could remove it, but `AppLayout` is strictly the same component in the three apps, I would rather keep it that way rather than having different versions.

apps/token-manager/app/src/App.js
bpierre: Because `title` need to be wrapped separately: https://github.com/aragon/aragon-apps/pull/666/files#diff-2e20014b3077a7461c11aabfbe2c53abR27

apps/voting/app/src/App.js
bpierre: Oops 🙈

apps/voting/app/src/components/MenuButton/MenuButton.js
bpierre: It is exactly the same for all the apps? [Finance](https://github.com/aragon/aragon-apps/blob/use-viewport/apps/finance/app/src/components/MenuButton/MenuButton.js) / [Token Manager](https://github.com/aragon/aragon-apps/blob/use-viewport/apps/token-manager/app/src/components/MenuButton/MenuButton.js) / [Voting](https://github.com/aragon/aragon-apps/blob/use-viewport/apps/voting/app/src/components/MenuButton/MenuButton.js)

bpierre:
> One commit? @bpierre these changes are too big for one commit only.

We squash our PRs, so we have very big commits in the history 😄

> Let's discuss a team convention for this but for me reviewing your changes it helps me a lot to read through the commit history looking at independent changes.

Sure, let’s discuss this. Just as a note, this is how I review PRs:

- On the first review, I check the entire diff, and test the changes using the latest commit of the PR. I don’t review commits individually because one commit might be obsoleted by another one happening after, and it might not be testable (e.g. if there was a mistake fixed later).
- On next reviews happening on the PR, I review commits beacause it helps me spot the changes since the initial review. For small PRs, I still have a look at the entire diff.

I assumed it made sense to make it all a single PR since it’s about aligning the three apps together, and `AppLayout` is exactly the same component in the three apps. But I understand it can be a lot to review. I’m going to close this PR and open separate ones to make it easier.

AquiGorka:
No need to close this one, I've already reviewed the big diff, we can continue with incremental commits.

> We squash our PRs, so we have very big commits in the history 😄

On master and with links to PRs which can potentially include the independent commits

Is the `MenuPanel` working for you?

bpierre:
> No need to close this one, I've already reviewed the big diff, we can continue with incremental commits.

Ah I just opened them… but the code is the same, so we can continue the review in these invididual PRs:

https://github.com/aragon/aragon-apps/pull/671
https://github.com/aragon/aragon-apps/pull/672
https://github.com/aragon/aragon-apps/pull/673

> Is the `MenuPanel` working for you?

Yes, how is it for you? Does it not open?

image

sohkai:
> Sure, let’s discuss this. Just as a note, this is how I review PRs:

> * On the first review, I check the entire diff, and test the changes using the latest commit of the PR. I don’t review commits individually because one commit might be obsoleted by another one happening after, and it might not be testable (e.g. if there was a mistake fixed later).
> * On next reviews happening on the PR, I review commits beacause it helps me spot the changes since the initial review. For small PRs, I still have a look at the entire diff.

This is exactly what I do too, I think having the PR linked in the commit is usually enough so if you had to, you can look back at the individual commits there.

PR #668, opened 2019-02-13
Actor: reuse Vault tests

sohkai:
The Vault now exports its tests via a sharable utility.

See f41519a for the changes made to the original tests to make them sharable.


future-apps/agent/test/vault.js
```diff
@@ -1,238 +1,6 @@
...
-      })
-
-      it('fails when attempting to recover tokens out of the vault', async () => {
-        await token.transfer(vault.address, 10)
-        assert.equal((await token.balanceOf(vault.address)), 10, 'vault should have 10 balance')
-        await assertRevert(() => vault.transferToVault(token.address))
-      })
-    })
-  })
+  runSharedVaultTests('Vault', { accounts, artifacts, web3 })
```
bingen: Shouldn't this be `'Agent'` instead?

PR #669, opened 2019-02-13
test: ignore all test contracts from coverage

sohkai:


sohkai:
Oops, turns out this was already done in `next`

PR #671, opened 2019-02-14
Finance: move to Viewport

bpierre:
- Remove uses of BreakPoint.
- Remove GetWindowSize.


apps/finance/app/src/components/Transfers.js
```diff
@@ -173,19 +176,23 @@ class Transfers extends React.Component {
...
-                    <BreakPoint from="medium">
-                      <TableRow>
-                        <DateHeader title="Date" />
-                        <SourceRecipientHeader title="Source / Recipient" />
-                        <ReferenceHeader title="Reference" />
-                        <AmountHeader title="Amount" align="right" />
-                        <TableHeader />
-                      </TableRow>
-                    </BreakPoint>
+                    <Viewport>
```
sohkai: Do we need this extra `Viewport` when we already have an instance above?

apps/finance/app/src/components/Transfers.js
```diff
@@ -199,12 +206,12 @@ class Transfers extends React.Component {
                         key={transfer.transactionHash}
                         token={tokenDetails[toChecksumAddress(transfer.token)]}
                         transaction={transfer}
-                        wideMode={width > 768 + 219}
+                        wideMode={width > breakpoints.medium + 219}
```
sohkai: Could we set a constant for `219`?

apps/finance/app/src/components/AppLayout.js
```diff
@@ -0,0 +1,103 @@
...
+  afterTitle,
+  smallViewPadding,
+  onMenuOpen,
+  mainButton,
+}) => {
+  return (
+    <Viewport>
+      {({ below }) => (
+        <AppView
+          padding={below('medium') ? smallViewPadding : 30}
```
sohkai: Maybe we could add a `largeViewPadding` that defaults to `30`?

apps/finance/app/src/App.js
```diff
@@ -125,137 +110,61 @@ class App extends React.Component {
...
-                    actionText="Add funds"
-                    onActivate={this.handleNewTransferOpen}
-                  />
-                </EmptyScreen>
-              )}
-            </AppView>
-            <SidePanel
+      <div css="min-width: 320px">
+        <Main>
+          <AppLayout
```
sohkai: So clean ✨

apps/finance/app/src/components/TransferRow.js
```diff
@@ -48,136 +48,110 @@ class TransferRow extends React.Component {
...
+                  <TextOverflow style={{ marginTop: '5px' }}>
+                    {reference}
+                  </TextOverflow>
+                  <Amount positive={isIncoming} style={{ marginTop: '5px' }}>
+                    {formattedAmount} {token.symbol}
+                  </Amount>
+                </Grid>
+              </StyledTableCell>
+            </TableRow>
+          ) : (
```
sohkai: I would say in this case it might make sense to have explicit `if/else` blocks, since the tree is so large and it's hard to spot the `:`

apps/finance/app/src/components/TransferRow.js
```diff
@@ -48,136 +48,110 @@ class TransferRow extends React.Component {
...
+                <ActionsWrapper>
+                  {network.etherscanBaseUrl && (
+                    <ContextMenu>
+                      <ContextMenuItem onClick={this.handleViewTransaction}>
+                        <IconTokens />
+                        <ActionLabel>View Transaction</ActionLabel>
+                      </ContextMenuItem>
+                    </ContextMenu>
+                  )}
+                  {showCopyTransferMessage && (
```
sohkai: Noticed you removed the commented out context menu (don't remember why we commented it out in the first place), but should we just remove all of the code related to it?

I find copying the transaction hash to be more useful, and worst case opening the etherscan URL gives you that anyway.

apps/finance/app/src/components/Transfers.js
bpierre: 🤦‍♂️ thanks!

apps/finance/app/src/components/AppLayout.js
bpierre: Yes good idea!

apps/finance/app/src/components/TransferRow.js
bpierre: Ah yes. I’m not sure why we decided to remove it, maybe for iframe permissions yes. I’m going to remove the code, and we’ll probably have a better solution soon within the transaction popover.

apps/finance/app/src/components/Transfers.js
bpierre: I ended up removing this entirely, since we don’t want `IdentityBadge` to be in wide mode anymore.

apps/finance/app/src/components/Transfers.js
```diff
@@ -205,7 +202,7 @@ class Transfers extends React.PureComponent {
                         key={transfer.transactionHash}
                         token={tokenDetails[toChecksumAddress(transfer.token)]}
                         transaction={transfer}
-                        wideMode={width > WIDE_MODE_FROM}
+                        smallViewMode={below('medium')}
```
sohkai: cc @AquiGorka on this, just to be sure this is safe 😄

bpierre: Oh yeah these two props seem similar but are unrelated:

- `wideMode` is an old one, that was used to make the badge shortened / not shortened (it was the first use of what became `Viewport`), and we don’t need it anymore since the badge should always be shortened now.
- `smallViewMode` is the small view mode that @AquiGorka introduced: I made it as an external prop to not have unnecessary instances of `Viewport`.

apps/finance/app/src/components/AppLayout.js
```diff
@@ -0,0 +1,106 @@
...
+  largeViewPadding: 30,
+}
+
+AppLayout.propTypes = {
+  children: PropTypes.node,
+  title: PropTypes.node.isRequired,
+  afterTitle: PropTypes.node,
+  onMenuOpen: PropTypes.func.isRequired,
+  smallViewPadding: PropTypes.number,
+  largeViewPadding: PropTypes.number,
```
AquiGorka: Same comment as on the other PR, should this be `mediumViewPadding`?

apps/finance/app/src/components/Transfers.js
AquiGorka: What @bpierre said is half-correct: `wideMode` was used to set the shorten property for the `IdentityBadge` which is lost [now](https://github.com/aragon/aragon-apps/pull/671/files#diff-160b40bf78932dbc027b7f8d9567d1a8R79)? This might break spacing for mobile, should it be set to `true` directly as the default mode?

PR #672, opened 2019-02-14
Voting: move to Viewport

bpierre:
- Remove uses of BreakPoint.


apps/voting/app/src/components/VotePanelContent.js
```diff
@@ -236,7 +230,7 @@ class VotePanelContent extends React.Component {
             <Label>Created By</Label>
           </h2>
           <Creator>
-            <IdentityBadge entity={creator} shorten={shortAddresses} />
```
sohkai: Do we not need this to be shortened at all now?

bpierre: I think it could be shortened at all times (which is the default), now that it opens with the full address. /cc @aragon/design

jounih: yeah I think the shortened version as default is good (for 0x addresses)

sohkai: Good point, that may actually make it easier to verify for most people (I assume most check 3-4 at the start and at the end), if they check at all.

PR #673, opened 2019-02-14
Token Manager: move to Viewport

bpierre:
- Remove uses of BreakPoint.
- Remove uses of WindowSizeProvider.


apps/token-manager/app/src/components/AppLayout.js
```diff
@@ -0,0 +1,106 @@
...
+          {children}
+        </AppView>
+      )}
+    </Viewport>
+  )
+}
+
+AppLayout.defaultProps = {
+  smallViewPadding: 20,
+  largeViewPadding: 30,
```
AquiGorka: Should this be `mediumViewPadding`?

bpierre: In the context of this component, there are only two modes: it doesn’t (and shouldn’t IMO) know about the global breakpoints. That’s why I was previously avoiding using “small” and “large”, and go for distinct names like “compact” instead.

PR #675, opened 2019-02-15
Survey: add old participation percent as part of MODIFY_PARTICIPATION_ROLE's params

sohkai:
Oops, looks like we forgot to add the old participation percent to the role's parameters. The `arapp.json`'s role params were correct before.


izqui:
Is this an issue that the CLI detected? If not, we should have a warning on `apm publish`

sohkai:
No, this wasn't; made https://github.com/aragon/aragon-cli/issues/370.

PR #676, opened 2019-02-15
 Survey: emit current option power on ResetVote events instead of previous option powers

sohkai:
Makes it easier for frontends to correctly respond to votes being reset (if they're trying to correctly reflect the options' powers) and the previous value is always known from `previousStake + currentPower`.


apps/survey/contracts/Survey.sol
```diff
@@ -157,9 +157,10 @@ contract Survey is AragonApp {
             for (uint256 i = 1; i <= previousVote.optionsCastedLength; i++) {
                 OptionCast storage previousOptionCast = previousVote.castedVotes[i];
                 uint256 previousOptionPower = survey.optionPower[previousOptionCast.optionId];
```
bingen: I don't know how the compiler handles this, but `previousOptionPower` is only used once, and although it's more readable this way maybe a comment would suffice.

PR #680, opened 2019-02-16
Survey: fix named returns from getSurvey()

sohkai:
Oops, looks like these return values were mistaken to be input params.

### DIFF
--- a/apps/survey/contracts/Survey.sol
+++ b/apps/survey/contracts/Survey.sol
@@ -261,24 +261,24 @@ contract Survey is AragonApp {
         view
         surveyExists(_surveyId)
         returns (
-            bool _open,
-            uint64 _startDate,
-            uint64 _snapshotBlock,
-            uint64 _minParticipationPct,
-            uint256 _votingPower,
-            uint256 _participation,
-            uint256 _options
+            bool open,
+            uint64 startDate,
+            uint64 snapshotBlock,
+            uint64 minParticipationPct,
+            uint256 votingPower,
+            uint256 participation,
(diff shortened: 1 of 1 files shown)


apps/survey/contracts/Survey.sol
```diff
@@ -261,24 +261,24 @@ contract Survey is AragonApp {
         view
         surveyExists(_surveyId)
         returns (
-            bool _open,
-            uint64 _startDate,
-            uint64 _snapshotBlock,
-            uint64 _minParticipationPct,
```
bingen: Too late, but I don't like this change that much, because it raises a compiler warning:
```
,/home/bingen/workspace/aragon-apps/apps/survey/contracts/Survey.sol:267:13: Warning: This declaration shadows an existing declaration.
            uint64 minParticipationPct,
            ^------------------------^
/home/bingen/workspace/aragon-apps/apps/survey/contracts/Survey.sol:66:5: The shadowed declaration is here:
    uint64 public minParticipationPct;
    ^-------------------------------^
```
That's why the underscores were used.

sohkai: Ahh, hadn't noticed the warning when I made this change.

The named return variable should just be `minParticipation`, similar to [how the Voting app does it](https://github.com/aragon/aragon-apps/blob/master/apps/voting/contracts/Voting.sol#L257).

The main thing about the underscore was that it looked like an input, since it was used as the prefix, and also looked weird for frontends consuming the ABI (all the returns were prefixed with an underscore).

PR #683, opened 2019-02-18
Finance: fix breakpoint definition in BalanceToken

sohkai:


AquiGorka:
❤️ @sohkai did not see the typo there

PR #686, opened 2019-02-20
Finance, Token Manager, Survey, Voting: fix function-visibility

sohkai:
From https://github.com/aragon/aragon-apps/pull/611#issuecomment-465090647:

> * (info) - voting.sol - access modifiers - consider declaring functions external (optimization)
> * (info) - survey.sol - access modifiers - consider declaring functions external (optimization)

---------

This has mostly cosmetic changes, to make the function grouping based on visibility more clear to reviewers and readers of the code.

The main changes to the contracts are:

- 08420fd: Changes `isForwarder()` to external, now that the interface is declared with `external`
- c46151e: Changes a few public functions in Finance to be external, since they either had internal variants subclasses could use or should not be used by subclasses
- 6ccfaa0: Modernizes the authentication strategy in Survey, to one where only the `external` functionality is protected by modifiers and other sanity checks, and the main logic is executed in `internal` functions

### DIFF
--- a/apps/survey/contracts/Survey.sol
+++ b/apps/survey/contracts/Survey.sol
@@ -167,7 +167,6 @@ contract Survey is AragonApp {
         external
         surveyExists(_surveyId)
     {
-        require(_optionIds.length == _stakes.length && _optionIds.length > 0, ERROR_VOTE_WRONG_INPUT);
         require(canVote(_surveyId, msg.sender), ERROR_CAN_NOT_VOTE);
 
         _voteOptions(_surveyId, _optionIds, _stakes);
@@ -301,6 +300,8 @@ contract Survey is AragonApp {
     * @dev Assumes the survey exists and that msg.sender can vote
     */
     function _voteOptions(uint256 _surveyId, uint256[] _optionIds, uint256[] _stakes) internal {
+        require(_optionIds.length == _stakes.length && _optionIds.length > 0, ERROR_VOTE_WRONG_INPUT);
+
         SurveyStruct storage survey = surveys[_surveyId];
 
         // Revert previous votes, if any
--- a/apps/survey/contracts/Survey.sol
+++ b/apps/survey/contracts/Survey.sol
@@ -145,11 +145,131 @@ contract Survey is AragonApp {
 
     /**
     * @notice Reset previously casted vote in survey #`_surveyId`, if any.
+    * @dev Initialization check is implicitly provided by `surveyExists()` as new surveys can only
+    *      be created via `newSurvey(),` which requires initialization
     * @param _surveyId Id for survey
     */
-    function resetVote(uint256 _surveyId) public isInitialized surveyExists(_surveyId) {
+    function resetVote(uint256 _surveyId) external surveyExists(_surveyId) {
         require(canVote(_surveyId, msg.sender), ERROR_CAN_NOT_VOTE);
 
+        _resetVote(_surveyId);
+    }
+
+    /**
+    * @notice Vote for multiple options in survey #`_surveyId`.
+    * @dev Initialization check is implicitly provided by `surveyExists()` as new surveys can only
+    *      be created via `newSurvey(),` which requires initialization
+    * @param _surveyId Id for survey
+    * @param _optionIds Array with indexes of supported options
+    * @param _stakes Number of tokens assigned to each option
+    */
+    function voteOptions(uint256 _surveyId, uint256[] _optionIds, uint256[] _stakes)
+        external
+        surveyExists(_surveyId)
+    {
+        require(_optionIds.length == _stakes.length && _optionIds.length > 0, ERROR_VOTE_WRONG_INPUT);
+        require(canVote(_surveyId, msg.sender), ERROR_CAN_NOT_VOTE);
+
+        _voteOptions(_surveyId, _optionIds, _stakes);
+    }
+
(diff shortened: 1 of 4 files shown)


apps/survey/contracts/Survey.sol
```diff
@@ -145,100 +145,46 @@ contract Survey is AragonApp {
 
     /**
     * @notice Reset previously casted vote in survey #`_surveyId`, if any.
+    * @dev Initialization check is implicitly provided by `surveyExists()` as new surveys can only
+    *      be created via `newSurvey(),` which requires initialization
     * @param _surveyId Id for survey
     */
-    function resetVote(uint256 _surveyId) public isInitialized surveyExists(_surveyId) {
+    function resetVote(uint256 _surveyId) external surveyExists(_surveyId) {
         require(canVote(_surveyId, msg.sender), ERROR_CAN_NOT_VOTE);
```
bingen: Shouldn't this be better inside the internal function? I think it's part of the main logic, it feels safer (making sure that vote won't ever be reset if it can not be reset) and makes the code easier to follow as there's no need to check all the places the internal function is called from.

sohkai: Going to push back on this one; `canVote()` feels similar to `surveyExists()` and is quite expensive (does a call to the token), so I'd prefer if it was only guarded on the external interfaces.

I've added a comment that users of `_voteOptions` and `_resetVote` should have already done precondition checks before invoking them in 0a8f7c0.

bingen: Ok, although I don't fully get second part of the argument: isn't the same cost if you call it from the internal function or from the external one?
Btw, thinking about that I realized that `token.balanceOfAt` is called 3 times besides from `canVote`: from `_resetVote`, `_voteOptions` and `voteOption`. I wonder if that could be further optimized, although I'm afraid it would make the code uglier.

sohkai: > Ok, although I don't fully get second part of the argument: isn't the same cost if you call it from the internal function or from the external one?

It'll get called once from the external function (e.g. `voteOptions`), and then again from the internal one.

Good catch on the `token.balanceOfAt()` though, perhaps we can pass that down but it feels a bit sketchy to do so :/

bingen: Well, my proposal was to put the check only in the internal one, and now that you mentioned the cost, actually only in `_resetVote`, as it's called at the beginning by `_voteOptions` (plus a comment noticing that the check is not needed because it's done in that other function, like we do with `isInitialized` modifier often).
But nevermind, if you like it better this way it's ok. It's true that feels more consistent with the way we are using modifiers like `surveyExists`. Probably because of not being so used to them, but I've never been a big fan of modifiers, haha.

tintinweb: > Ok, although I don't fully get second part of the argument: isn't the same cost if you call it from the internal function or from the external one?

gas consumption is different (https://ethereum.stackexchange.com/a/55426/3544) although unchecked if this is the heavy part when calling `reset_vote`.

tintinweb: regarding where to put the check.
I usually advocate for putting the check where the actual modification happens (innermost basicblock/function). Reason is that by putting the check to the inner function you are still guarded in case you're adding a new unprotected externally reachable function that at some point calls the -now protected- internal method. If you can guarantee that this won't happen you're safe anyway. Adding a comment and relying that someone understands the risk is often not a secure by default approach (can be critical if multiple devs are working on the same codebase). 
on the flipside, bailing early safes some gas but I'd expect the internal call not to consume a lot here.

bingen: I completely agree, it's what I meant by "it feels safer (...)", but way better explained.
About the gas, the difference would only be when it reverts right? I mean, if the transaction succeeds there wouldn't be any difference, right? (As it will always call the internal function) If that's the case, I don't think it matters too much: when used from the frontend, you won't be allowed to attempt a transaction for a non-open vote, and this is what regular users will do. In general, honest use should check first if the vote is open and mistakes here should be really rare. I expect hitting this check mostly from malicious attempts, and in that case I'm not concerned about some extra gas used from the attacker.

tintinweb: yeah thats right, front-end checks are done for convenience and not security (can easily be bypassed) to save users from running into calls we already know will fail. the contract actually implements the security check protecting state. ideally most issues should be caught in the front-end (convenience, avoiding to waste gas) and if someone fails in a transaction (due to security checks in the contract) this is likely a malicious attempt or bug that needs to be investigated.

regarding gas consumption (unverified assumption, but we can if needed):
there is a difference between an external call to `public` and `external` function as well as an internal call to an `internal` function. the internal call will probably just be a JUMP (something around 3(push4)+8(jump)+x gas) and thats negligible. the difference of public/external for the external call in gas consumption according to the link pasted earlier is unlikely to be very significant (abt. 250 gas more for `public` if the values are accurate) for your function (`canVote` is calling an external address, `_reset` is calling an external address). If the require bails excess gas will [not be consumed](https://solidity.readthedocs.io/en/v0.4.24/control-structures.html#error-handling-assert-require-revert-and-exceptions).

we can therefore expect a slightly lower gas footprint by switching to `external`. gas consumption will be at max in the success case, and will only consume as much as needed if it bails.

PR #688, opened 2019-02-21
Agent: fix execute signature ACL param

izqui:
Found during security review. See https://github.com/aragon/aragon-apps/pull/611#discussion_r

The `signature` ACL parameter for the `EXECUTE_ROLE` contains a memory pointer rather than the actual value. The ACL param shouldn't be trusted until the Agent instance has been upgraded to a version with this patch.

Pending tests.


izqui:
I love how if we had actually tried to remove the TODO we would have found the issue

PR #692, opened 2019-02-26
fix: update usage of @aragon'ui's Main component to use relative asset paths

sohkai:
cc @bpierre should we just make [`Main` default to using relative paths](https://github.com/aragon/aragon-ui/blob/master/src/components/Main/Main.js#L28)? (From @2color).

### DIFF
--- a/apps/finance/app/src/App.js
+++ b/apps/finance/app/src/App.js
@@ -109,7 +109,7 @@ class App extends React.Component {
 
     return (
       <div css="min-width: 320px">
-        <Main>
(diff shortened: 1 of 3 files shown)


bpierre:
Yes let’s do that. I don’t like the idea of having a default configuration that doesn’t work with URLs, but it is how it is in apps and aragon/aragon at the moment, and I really want `Main` to work without any prop by default. We can think about it again whenever we’ll support both location hashes and paths. 👍

PR #695, opened 2019-02-27
Voting: default decimals to 0 if token contract doesn't specify them

sohkai:
Defaults the token decimals to 0 ("safest" assumption) if the token contract doesn't provide them.

In some cases, this can make the values look quite odd (if they're in the yuge range), and this includes a bit of logic to handle these exponents.

We'll leave the decimal matching as something to tackle with https://github.com/aragon/aragon/issues/589.

Numbers in sci. notation:


sohkai:
> Shall we start to move math-utils.js in a shared directory?

Will do separately :+1:.

PR #701, opened 2019-03-08
Add metadata to apps

bpierre:
Adds:

- New icons (the colorful ones).
- Long description (`details.md`).
- Screenshots.

Other changes:

- A `meta/` directory has been created. It contains the assets used to describe an app: its icons, screenshots, and long description.
- The `icon.svg` files have been renamed `icon-menu.svg` (and moved to `meta/`). It is not included in the manifest.json file because we recently agreed to only have one type of icon for the apps, in the next iteration of the Aragon menu. That way we can keep trying to load them until the Aragon menu gets updated, by trying to load them directly.
- The icon of the vault (and its details.md) is now in public/meta, to align it with the webapp-apps.


john-light:
@bpierre @sohkai  could I get a screenshot of the copy in context? This will help me determine if e.g. there is too much or too little text. Based on the "details" pages I see here, it looks like quite a lot of text for the description section shown in the screenshot below, for example.

Screenshot_2019-03-31 Figma

dizzypaty:
Based on a quick chat I had with @bpierre  yesterday, I tweaked slightly the layout on the app details view to better fit large amount of content (using 2 columns to display the 'Description' section).

These changes are minor and Pierre confirmed that they won’t impact the estimated development time of the App Center. 

WebApp-1366px - App detail view 2

[Figma file with assets here.](https://www.figma.com/file/c9UcAQIEsNOgaBQKY7NFPOfw/Apps-Center?node-id=452%3A2)

sohkai:
@john-light We'll merge this for now, so we can use the current text (that's fairly outdated and inaccurate) to test the app center. Let's make sure to review and modify the text before the App Center is actually released.

john-light:
@sohkai looking at this now, will PR off master. Is there a reason Survey has a manifest file but does not have a details.md file?

sohkai:
> Is there a reason Survey has a manifest file but does not have a details.md file?

Primarily because it's not an entry in the App Center yet, but we should write some text and describe it 👍.

> In what context is the copy in the Manifest file rendered

The markdown should be in the big "Description" section in https://github.com/aragon/aragon-apps/pull/701#issuecomment-480211685.

>  This may be pre-mature but I think there should be a section on the description page that specifically details a timeline of security audits

We had discussed an "audit" badge that would link you to the relevant security audit report. Probably don't need a timeline; you probably don't care when the audit started 😄

> Also, is there a way to revert an update or go down to a lower version if needed?

Yes, this is possible (but not exposed in the UI right now).

cc @dizzypaty for the text change.

john-light:
Ok, thanks.

>  > In what context is the copy in the Manifest file rendered
> 
> The markdown should be in the big "Description" section in #701 (comment).

@sohkai the Manifest file only has a short description of the app, see e.g. https://github.com/aragon/aragon-apps/pull/784/files#diff-36647c603cfd401f394003214d31c521

>  Probably don't need a timeline; you probably don't care when the audit started smile

I mean a history of audits: Audit 1 happened on date x, Audit 2 happened on date y, etc etc. Maybe also link to any "official" responses to the audits.

bpierre:
For reference, this other PR updates the copy https://github.com/aragon/aragon-apps/pull/784

PR #703, opened 2019-03-11
Token Manager: enforce all token controller methods to only be accessible via the managed token contract

sohkai:


apps/token-manager/contracts/TokenManager.sol
```diff
@@ -62,6 +62,11 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
     event NewVesting(address indexed receiver, uint256 vestingId, uint256 amount);
     event RevokeVesting(address indexed receiver, uint256 vestingId, uint256 nonVestedAmount);
 
+    modifier onlyToken() {
+        require(msg.sender == address(token), ERROR_ON_TRANSFER_WRONG_SENDER);
```
bingen: I guess the `ON_TRANSFER` part of the name should be changed now, plus now `ERROR_PROXY_PAYMENT_WRONG_SENDER` is not used anymore.

apps/token-manager/contracts/TokenManager.sol
```diff
@@ -239,9 +247,7 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
     * @param _amount The amount of the transfer
     * @return False if the controller does not authorize the transfer
     */
-    function onTransfer(address _from, address _to, uint _amount) public isInitialized returns (bool) {
```
bingen: I would add an explicit comment about why `isInitialized` is not needed here.

PR #704, opened 2019-03-11
Token Manager: disallow assigning tokens to self

sohkai:
_Technically_ this isn't really a problem but seems like confusing behaviour we should protect against.

### DIFF
--- a/apps/token-manager/contracts/TokenManager.sol
+++ b/apps/token-manager/contracts/TokenManager.sol
@@ -31,6 +31,7 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
 
     string private constant ERROR_NO_VESTING = "TM_NO_VESTING";
     string private constant ERROR_TOKEN_CONTROLLER = "TM_TOKEN_CONTROLLER";
+    string private constant ERROR_ASSIGN_TO_SELF = "TM_ASSIGN_TO_SELF";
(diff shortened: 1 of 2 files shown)


sohkai:
@bingen The `assign()` action is specifically to transfer tokens held by the Token Manager to another address. It is not an arbitrary movement of funds.

In this case, it is awkward that you can successfully do:

1. A no-op, transferring tokens from itself to itself (note that typical tokens, including MiniMe do allow this to succeed)
1. Create a vesting contract for itself, when it already holds those tokens

bingen:
> The `assign()` action is specifically to transfer tokens held by the Token Manager to another address. It is not an arbitrary movement of funds.

Yes, that's true, so the last part of my comment doesn't make any sense, but I still don't see the point of adding this check, to prevent something that does no harm and it would just save some gas in case somebody does the mistake of doing a non-sense assignment at the cost of increasing (even if it's only a tiny amount) gas cost for all regular and correct assignments. But as you want.

sohkai:
> It would make probably less sense to assign to the token address or to zero address for instance, and we are not protecting against it.

> Prevent something that does no harm and it would just save some gas in case somebody does the mistake of doing a non-sense assignment at the cost of increasing (even if it's only a tiny amount) gas cost for all regular and correct assignments

Hmm, yes, I am agreeing now that this is a losing battle for little gain and more gas cost per invocation. Assigning vested tokens to the token manager is similarly a no op, since it can always transfer its balances.

PR #705, opened 2019-03-11
Survey: optimize voteOption()

sohkai:
Optimizes `_voteOption()` by caching the `MultiOptionVote` storage variable, rather than recalculating its position multiple times.

### DIFF
--- a/apps/survey/contracts/Survey.sol
+++ b/apps/survey/contracts/Survey.sol
@@ -181,8 +181,8 @@ contract Survey is AragonApp {
     * @param _optionId Index of supported option
     */
     function voteOption(uint256 _surveyId, uint256 _optionId) external surveyExists(_surveyId) {
-        require(_optionId != ABSTAIN_VOTE, ERROR_VOTE_WHOLE_WRONG_OPTION);
         require(canVote(_surveyId, msg.sender), ERROR_CAN_NOT_VOTE);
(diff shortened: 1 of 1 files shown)


apps/survey/contracts/Survey.sol
```diff
@@ -181,8 +181,8 @@ contract Survey is AragonApp {
     * @param _optionId Index of supported option
     */
     function voteOption(uint256 _surveyId, uint256 _optionId) external surveyExists(_surveyId) {
-        require(_optionId != ABSTAIN_VOTE, ERROR_VOTE_WHOLE_WRONG_OPTION);
         require(canVote(_surveyId, msg.sender), ERROR_CAN_NOT_VOTE);
+        require(_optionId != ABSTAIN_VOTE, ERROR_VOTE_WHOLE_WRONG_OPTION);
```
bingen: What's the rationale for this change?
Besides, I realize now that we are already doing this check in `_voteOptions`, in L321, and in a more complete way as we also check `&& optionId <= survey.options`, so maybe we can omit it here.

sohkai: Main rationale was to move authentication checks as far up the stack as possible (see https://github.com/aragon/aragon-apps/pull/710).

However, yes, I think we should remove this check since it's done internally (and checks are applied closest to where it's used).

PR #706, opened 2019-03-11
Finance: optimize SafeMath

sohkai:
It's very unlikely incrementing uint256 array lengths will be a problem (we probably have bigger ones if we ever get to the point where these overflow), but still better practice to be safe.

Also removes SafeMath from a few places where we've already done the check implicitly.


sohkai:
👍 Makes sense; I'll remove the additional `SafeMath` additions. I was mainly thinking if we enabled the SMT checker these increments would fail, but I agree, these are basically impossible to hit.

PR #709, opened 2019-03-11
Finance, TokenManager: move auth modifier to after state sanity checks

sohkai:
Reorders `authP()` checks so that they're only applied after sanity checking state.

This is important so that the error message we provide is accurate in the case of parameterized permissions; if the state we expect to use as a parameter is non-existent, that should be the error instead of an authentication error.

### DIFF
--- a/apps/token-manager/contracts/TokenManager.sol
+++ b/apps/token-manager/contracts/TokenManager.sol
@@ -179,8 +179,8 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
     */
     function revokeVesting(address _holder, uint256 _vestingId)
         external
-        authP(REVOKE_VESTINGS_ROLE, arr(_holder))
         vestingExists(_holder, _vestingId)
+        authP(REVOKE_VESTINGS_ROLE, arr(_holder))
     {
         TokenVesting storage v = vestings[_holder][_vestingId];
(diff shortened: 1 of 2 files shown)


apps/token-manager/contracts/TokenManager.sol
```diff
@@ -179,8 +179,8 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
     */
     function revokeVesting(address _holder, uint256 _vestingId)
         external
-        authP(REVOKE_VESTINGS_ROLE, arr(_holder))
         vestingExists(_holder, _vestingId)
+        authP(REVOKE_VESTINGS_ROLE, arr(_holder))
```
bingen: I don't fully get why the state check has to take precedence over the authentication one, specially in a case like this one where they are more disjoint, but I don't see any strong reason for the opposite either, so I'm fine with these changes.

sohkai:
> Although I can see the rationale behind the change in Finance.executePayment(), I don't think it is too bad that we are sending 0 as the amount of an inexistent payment.

This is really the only case I was worried about, but yes, on second thought keeping the auth checks first is more consistent with all the other checks we do once we enter into functions :+1:.

I can see other cases where you're more likely fail the auth check if something was non-existent, but you're likely to realize you accessed non-existent state once you start digging in.

PR #710, opened 2019-03-11
Finance, Survey, Token Manager, Voting: Clarify internal protection scheme

sohkai:
Builds on https://github.com/aragon/aragon-apps/pull/702.

An attempt at an answer to https://github.com/aragon/aragon-apps/pull/686#discussion_r259411928.

This will be added as a guideline to hack.aragon.org (soon), but in short, my current thoughts on where to place security modifiers in Aragon apps:

- "Lowest" in the call stack:
  - Place internal state assertions as close as possible to where they're required or used
  - Place external call assertions as close as possible to where the call happened
- "Highest" in the call stack:
  - Place authentication assertions as close as possible to the start of the externally accessible function
  - Place parameter assertions as close as possible to the start of the externally accessible function
- In places where an expensive state assertion will be repeated because a caller wants to branch based on the assertion (rather than allowing it to revert), add an "unsafe" version of the internal call that skips checks

### DIFF
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -149,6 +149,7 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
      * @notice Deposit ETH to the Vault, to avoid locking them in this Finance app forever
      */
     function () external payable isInitialized transitionsPeriod {
+        require(msg.value > 0, ERROR_DEPOSIT_AMOUNT_ZERO);
         _deposit(
             ETH,
             msg.value,
@@ -192,6 +193,7 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
     * @param _reference Reason for payment
     */
     function deposit(address _token, uint256 _amount, string _reference) external payable isInitialized transitionsPeriod {
+        require(_amount > 0, ERROR_DEPOSIT_AMOUNT_ZERO);
         _deposit(
             _token,
             _amount,
@@ -534,7 +536,6 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
     // Internal fns
 
     function _deposit(address _token, uint256 _amount, string _reference, address _sender, bool _isExternalDeposit) internal {
-        require(_amount > 0, ERROR_DEPOSIT_AMOUNT_ZERO);
         _recordIncomingTransaction(
             _token,
             _sender,
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -546,15 +546,15 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
         // External deposit will be false when the assets were already in the Finance app
         // and just need to be transferred to the vault
         if (_isExternalDeposit) {
-            if (_token != ETH) {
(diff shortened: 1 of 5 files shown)


apps/voting/contracts/Voting.sol
```diff
@@ -306,11 +305,15 @@ contract Voting is IForwarder, AragonApp {
...
 
         if (_executesIfDecided && _canExecute(_voteId)) {
-            _executeVote(_voteId);
+            _executeVote(_voteId, true);
         }
     }
 
-    function _executeVote(uint256 _voteId) internal {
+    function _executeVote(uint256 _voteId, bool _skipChecks) internal {
+        if (!_skipChecks) {
```
bingen: To me this makes the code unnecessarily more complex.

apps/finance/contracts/Finance.sol
```diff
@@ -616,11 +599,15 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
...
-        string _reference
+        string _reference,
+        bool _skipChecks
     )
         internal
     {
-        require(_getRemainingBudget(_token) >= _amount, ERROR_REMAINING_BUDGET);
+        if (!_skipChecks) {
+            require(_getRemainingBudget(_token) >= _amount, ERROR_REMAINING_BUDGET);
+        }
```
bingen: Is there a reason to not write it as:
```
require(_skipChecks || _getRemainingBudget(_token) >= _amount, ERROR_REMAINING_BUDGET);
```
?

bingen: I would actually move this check to `newPayment` so we can avoid the use of this `_skipChecks` parameter at all.

apps/finance/contracts/Finance.sol
```diff
@@ -546,12 +548,12 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
         // External deposit will be false when the assets were already in the Finance app
         // and just need to be transferred to the vault
         if (_isExternalDeposit) {
-            if (_token != ETH) {
-                // Get the tokens to Finance
-                require(ERC20(_token).safeTransferFrom(msg.sender, this, _amount), ERROR_TOKEN_TRANSFER_FROM_REVERTED);
-            } else {
+            if (_token == ETH) {
                 // Ensure that the ETH sent with the transaction equals the amount in the deposit
                 require(msg.value == _amount, ERROR_VALUE_MISMATCH);
```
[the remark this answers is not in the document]
sohkai: Done in aa7017b

apps/token-manager/contracts/TokenManager.sol
```diff
@@ -324,6 +323,10 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
     }
 
     function _mint(address _receiver, uint256 _amount) internal {
+        // Max balance doesn't apply to the token manager itself
+        if (_receiver != address(this)) {
```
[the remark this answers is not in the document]
sohkai: See https://github.com/aragon/aragon-apps/pull/734.

sohkai:
@izqui @bingen I've moved to using `unsafe` variants of the internal methods instead.

PR #711, opened 2019-03-12
Finance: consistent payment creation

sohkai:
Splits up `newPayment()` into:

- `newScheduledPayment()`: _always_ saves a `RecurringPayment`, for repeating or scheduled future payments
- `newImmediatePayment()`: _only_ for single, immediately-payable transactions

It also modifies the execution behaviour of payments. Previously, it was possible to call `executePayment()` or `receiverExecutePayment()` even if no payments were made (only a single `PaymentFailure` event would be emitted). Now calling these two functions will cause a revert, with more obvious error messages (either you hadn't waited long enough, or there wasn't enough balance in the Vault).

We still allow `newScheduledPayment()` to create a "failing" payment, as I think it'd still be useful to schedule payments even if you can't immediately pay them out.


apps/finance/contracts/Finance.sol
```diff
@@ -322,9 +345,14 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
...
-        require(_nextPaymentTime(_paymentId) <= getTimestamp64(), ERROR_EXECUTE_PAYMENT_TIME);
-
-        _executePayment(_paymentId);
+        uint256 paid = _executePayment(_paymentId);
+        if (paid == 0) {
+            if (_nextPaymentTime(_paymentId) > getTimestamp64()) {
+                revert(ERROR_EXECUTE_PAYMENT_TIME);
+            } else {
+                revert(ERROR_EXECUTE_PAYMENT_NUM);
+            }
```
bingen: This could be written in 2 lines, as:
```
require(_nextPaymentTime(_paymentId) <= getTimestamp64(), ERROR_EXECUTE_PAYMENT_TIME);
revert(ERROR_EXECUTE_PAYMENT_NUM);
```
Not saying it is better this way, it just came up to my mind and found it kind of nice, but maybe it is clearer as it is. :wink:

apps/finance/test/finance.js
```diff
@@ -13,6 +14,11 @@ const TokenReturnMissingMock = artifacts.require('TokenReturnMissingMock')
 const getContract = name => artifacts.require(name)
 
 const getEventData = (receipt, event, arg) => receipt.logs.filter(log => log.event == event)[0].args[arg]
+const assertPaymentFailure = receipt => {
+    const filteredLogs = receipt.logs.filter(log => log.event == 'PaymentFailure')
+    assert.equal(filteredLogs.length, 1, 'should have logged payment failure')
+}
```
bingen: Couldn't we use `assertEvent` here?

apps/finance/contracts/Finance.sol
```diff
@@ -322,9 +345,14 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
...
+        }
```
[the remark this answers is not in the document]
bingen: Now I see that error messages are slightly different, but I don't think that is necessary, as you usually know which function are you calling, right? (and those functions are both `external`)

apps/finance/contracts/Finance.sol
```diff
@@ -258,11 +256,36 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
...
+    /**
+    * @notice Create a new payment of `@tokenAmount(_token, _amount)` to `_receiver` for '`_reference`'
+    * @param _token Address of token for payment
+    * @param _receiver Address that will receive payment
+    * @param _amount Tokens that are paid every time the payment is due
+    * @param _reference String detailing payment reason
+    */
+   function newPaymentTransaction(address _token, address _receiver, uint256 _amount, string _reference)
+        external
+        authP(CREATE_PAYMENTS_ROLE, arr(_token, _receiver, _amount, MAX_UINT, 1))
```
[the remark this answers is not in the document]
izqui: I think most users will parametrize it with a `>`, as setting a payment with a very low interval can drain the vault.

apps/finance/contracts/Finance.sol
```diff
@@ -566,15 +600,15 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
         }
     }
 
-    function _executePayment(uint256 _paymentId) internal {
+    function _executePayment(uint256 _paymentId) internal returns (uint256) {
```
izqui: We could add an extra return value that is true if the `PaymentFailure` event was emitted, which would remove the need to perform a redundant `_nextPaymentTime(_paymentId) > getTimestamp64()` to decide the revert reason.

apps/finance/contracts/Finance.sol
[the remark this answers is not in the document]
bingen: Yes, that makes sense, but actually both could be set using an `AND` operator, and then this function would not work.

apps/finance/test/finance.js
```diff
@@ -1,3 +1,4 @@
+const assertEvent = require('@aragon/test-helpers/assertEvent')
```
izqui: I think we should check revert reasons in this contract to ensure that things are reverting with the error that we expected. See https://github.com/aragonone/aragon-court/blob/6d0b322becfb5ff528437c14087ebd07cc918520/test/helpers/assert-revert.js which uses Truffle 4

apps/finance/contracts/Finance.sol
```diff
@@ -228,24 +233,17 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
         returns (uint256 paymentId)
```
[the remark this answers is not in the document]
sohkai: Good thing nobody's using the params!

apps/finance/contracts/Finance.sol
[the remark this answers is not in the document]
sohkai: I chose `MAX_UINT64` against `0` since most people will probably want `>`.

I'll document it, but another thing we could do is use `MAX_UINT256` instead here, as a sentinel interval value for "never".

apps/finance/contracts/Finance.sol
```diff
@@ -258,11 +256,36 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
         payment.maxRepeats = _maxRepeats;
         payment.createdBy = msg.sender;
 
-        if (_nextPaymentTime(paymentId) <= getTimestamp64()) {
-            _executePayment(paymentId);
-        }
+        // We skip checking how many times the new payment was executed to allow creating new
+        // recurring payments before having enough vault balance
```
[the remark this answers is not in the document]
sohkai: `executePayment()` will revert if you try to execute a payment that results in no actual transactions.

Here, in `newPayment()`, I'd like to explicitly allow this, so you can "store" a payment that should be immediately payable (but you can't pay yet, perhaps due to a lack of funds).

The intention is that it's kind of like signing up for a recurring service fee even if you can't pay the first bill when you initially set up the recurring payment.

apps/finance/contracts/Finance.sol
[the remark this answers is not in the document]
sohkai: I think this is reasonable; the function signature should be easy to spot.

apps/finance/contracts/Finance.sol
sohkai: Another alternative is to use a ternary, but some quick checking with this on remix suggests the ternary is more expensive than the if/else 🤷‍♂️.

apps/finance/contracts/Finance.sol
sohkai: Was looking at this and I don't think this covers the right cases. `PaymentFailure` is only emitted if at least one payment was due, but the funds weren't available, and needs the `_nextPaymentTime(_paymentId) > getTimestamp64()` in each iteration.

In the check, we want to detect if the payment wasn't due at all for deciding the revert reason.

sohkai:
@izqui @bingen Ready for a re-review.

Changes:

- 1fd6ade: modifies some radspec strings, given the scope of the new functions
- 7d00e37: Moved the execution error checking to a new internal function, `_executePaymentAtLeastOnce()`
- 0748ed3: Renamed `newPaymentTransaction()` to `newImmediatePayment()`
- e302575: Renamed `newPayment()` to `newScheduledPayment()` (see https://github.com/aragon/aragon-apps/pull/735 for another renaming for the structs given this change)

apps/finance/contracts/Finance.sol
```diff
@@ -781,6 +820,18 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
         return nextPayment;
     }
 
+    // Syntax sugar
+
+    function _arr(address _a, address _b, uint256 _c, uint256 _d, uint256 _e, uint256 _f) internal pure returns (uint256[] r) {
```
sohkai: This is a bit awkward; for some reason Solidity won't let me use just the `arr()` name :(.

Don't really think we need to port this over back to aragonOS; this is a really specific helper.

apps/finance/contracts/Finance.sol
izqui: Might not be getting the full picture and this is a very small optimization, but the `PaymentFailure` event can only be emitted if there was a payment due that couldn't be performed. If we returned one extra value (`bool lastPaymentFailed`) then in `_executePaymentAtLeastOnce` we could just decide on the revert reason depending on `lastPaymentFailed` rather than re-checking if there was a payment due.

apps/finance/contracts/Finance.sol
```diff
@@ -228,28 +234,22 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
         string _reference
     )
         external
-        authP(CREATE_PAYMENTS_ROLE, arr(_token, _receiver, _amount, _interval, _maxRepeats))
+        authP(CREATE_PAYMENTS_ROLE, _arr(_token, _receiver, _amount, uint256(_initialPaymentTime), uint256(_interval), uint256(_maxRepeats)))
```
izqui: I think new params should always be added at the end.

apps/finance/contracts/Finance.sol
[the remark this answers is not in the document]
izqui: Even if we think so, we shouldn't change the param orders as it is a breaking change.

apps/finance/contracts/Finance.sol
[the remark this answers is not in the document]
izqui: That makes sense

apps/finance/contracts/Finance.sol
bingen: Yes, it looks like. Although that seems to me like a hard question (which ones to include in aragonOS helpers), the possible combinations of parameters are endless.

apps/finance/contracts/Finance.sol
bingen: Yes, I don't get it either. Assuming `paid == 0` back in `_executePaymentAtLeastOnce`, if `_nextPaymentTime(_paymentId) > getTimestamp64()` then it means that it didn't get into the loop of `_executePayment`, so `PaymentFailure` wasn't emitted; while if the opposite holds true, then it means it got into the loop and therefore it went through the `break` and so the event was emitted, right? Maybe I'm missing something too.

sohkai: Ohh yes, I see what @izqui is saying now. Yes, if `paid == 0` and a `PaymentFailure()` event is emitted then we will always be throwing `ERROR_EXECUTE_PAYMENT_NUM`.

I've reordered the `if`s in `_executePaymentAtLeastOnce()` to make this a bit more clear, but decided against the extra return value as it'd make successful calls more expensive.

PR #712, opened 2019-03-12
Token Manager: remove unused storage mapping

sohkai:
Seems like we forgot to remove this earlier. Shouldn't cause an issue with any upgrades since we weren't using the storage anyway.

### DIFF
--- a/apps/token-manager/contracts/TokenManager.sol
+++ b/apps/token-manager/contracts/TokenManager.sol
@@ -56,7 +56,6 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
     // We are mimicing an array in the inner mapping, we use a mapping instead to make app upgrade more graceful
     mapping (address => mapping (uint256 => TokenVesting)) internal vestings;
     mapping (address => uint256) public vestingsLengths;
-    mapping (address => bool) public everHeld;
 
     // Other token specific events can be watched on the token address directly (avoids duplication)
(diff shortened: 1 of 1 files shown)


izqui:
> Shouldn't cause an issue with any upgrades since we were using the storage anyway.

@sohkai what do you mean?

apps/token-manager/contracts/TokenManager.sol
```diff
@@ -56,7 +56,6 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
     // We are mimicing an array in the inner mapping, we use a mapping instead to make app upgrade more graceful
     mapping (address => mapping (uint256 => TokenVesting)) internal vestings;
     mapping (address => uint256) public vestingsLengths;
-    mapping (address => bool) public everHeld;
```
izqui: Can be safely removed because luckily it was the last storage mapping of the contract and it has never been used, although if we want to be really safe, we could leave a storage gap:

```
uint256 private everHeldDeprecationGap;
```

sohkai:
@izqui Oops, typo. But yes, luckily it was the last storage item and never used in the aOS 4 version of this contract.

I'd prefer not having the gap, and just pretending that it was never there since it's unnecessary 😄 .

izqui:
Makes sense, I was definitely thinking about how we should have done if it hadn't been in the last slot. LGTM

PR #715, opened 2019-03-12
New icons

luisivan:
New icons for:
- [x] Token Manager
- [x] Voting
- [x] Finance
- [ ] Survey
- [ ] Payroll


apps/finance/app/public/images/icon.svg
```diff
@@ -1 +1,28 @@
...
\ No newline at end of file
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
sohkai: One thing: could we put all of these svgs through [SVGOMG](https://jakearchibald.github.io/svgomg/) so they can be optimized?

sohkai:
Closing for #701.

PR #698, opened 2019-03-04
Upgrade rxjs to v6 in finance app

2color:
## What

- Add `rxjs` as a dependency
- Improve dev experience by watching the script in dev/watch mode too
- Make code changes where necessary

PR #716, opened 2019-03-18
Upgrade rxjs to v6 in token-manager app 

2color:
## What
- Add `rxjs` as a dependency
- Improve dev experience by watching the script in dev/watch mode too
- Make rxjs v6 code changes where necessary 

This PR is in the same vein as https://github.com/aragon/aragon-apps/pull/698

### DIFF
--- a/apps/token-manager/app/src/script.js
+++ b/apps/token-manager/app/src/script.js
@@ -1,5 +1,5 @@
 import Aragon from '@aragon/api'
-import { of } from './rxjs'
+import { of } from 'rxjs'
 import tokenSettings, { hasLoadedTokenSettings } from './token-settings'
 import { addressesEqual } from './web3-utils'
 import tokenAbi from './abi/minimeToken.json'
@@ -38,16 +38,13 @@ const retryEvery = (callback, initialRetryTimer = 1000, increaseFactor = 5) => {
 
 // Get the token address to initialize ourselves
 retryEvery(retry => {
-  app
-    .call('token')
-    .first()
-    .subscribe(initialize, err => {
-      console.error(
-        'Could not start background script execution due to the contract not loading the token:',
-        err
-      )
-      retry()
-    })
+  app.call('token').subscribe(initialize, err => {
+    console.error(
+      'Could not start background script execution due to the contract not loading the token:',
+      err
+    )
+    retry()
+  })
 })
 
 async function initialize(tokenAddr) {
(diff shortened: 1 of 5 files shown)


apps/token-manager/app/src/script.js
```diff
@@ -169,10 +166,7 @@ function updateHolders(holders, changed) {
 
 function loadMaxAccountTokens() {
   return new Promise((resolve, reject) =>
-    app
-      .call('maxAccountTokens')
-      .first()
-      .subscribe(resolve, reject)
+    app.call('maxAccountTokens').subscribe(resolve, reject)
```
bpierre: Maybe we could remove `loadMaxAccountTokens` entirely? Since we can get the value using:

```js
await app.call('maxAccountTokens').toPromise()
```

apps/token-manager/app/src/script.js
```diff
@@ -183,7 +177,6 @@ function loadNewBalances(token, ...addresses) {
         new Promise((resolve, reject) =>
           token
             .balanceOf(address)
-            .first()
             .subscribe(balance => resolve({ address, balance }), reject)
         )
```
bpierre: What about doing something like this?

```js
token
  .balanceOf(address)
  .toPromise()
  .then(address => ({ address, balance }))
```

apps/token-manager/app/src/script.js
```diff
@@ -200,10 +193,7 @@ function loadNewBalances(token, ...addresses) {
 
 function loadTokenSupply(token) {
   return new Promise((resolve, reject) =>
-    token
-      .totalSupply()
-      .first()
-      .subscribe(resolve, reject)
+    token.totalSupply().subscribe(resolve, reject)
   )
```
bpierre: We could probably get rid of `loadTokenSupply` entirely, since doing this is not much longer:

```js
token.totalSupply().toPromise()
```

apps/token-manager/app/src/script.js
```diff
@@ -212,11 +202,9 @@ function loadTokenSettings(token) {
...
-            .first()
-            .subscribe(value => {
-              resolve({ [key]: value })
-            }, reject)
+          token[name]().subscribe(value => {
+            resolve({ [key]: value })
+          }, reject)
         )
     )
   )
```
bpierre: Maybe we could do it like this?

```js
return Promise.all(
  tokenSettings.map(
    ([name, key]) => token[name]()
      .toPromise()
      .then(value => ({ [key]: value }))
  )
```

apps/token-manager/app/src/script.js
```diff
@@ -232,9 +220,6 @@ function loadTokenSettings(token) {
 
 function loadTokenSymbol(token) {
```
bpierre: And similar to the other comments, we could probably remove `loadTokenSymbol()`.

2color:
> Do you think it would make sense if we just changed all the single-emission observables in @aragon/api to return promises instead and release a new major version of it (v2.0.0) 

For sure.

Issue #510, opened 2018-10-19
Make vote descriptions more user-friendly


john-light:
Currently the vote descriptions have addresses where we could instead show app badges ("Voting App", hover over to see full address) or (once we have the identity app) ENS names for entities. Using badges and ENS names the descriptions become much more easy to read.

![aragon_vote_description](https://user-images.githubusercontent.com/9424721/47219781-1718b480-d3b0-11e8-919c-386a058314cf.png)

sohkai:
And we could also add a "role badge" for permissions bytes!

sohkai:
We now use IdentityBadges and role descriptions in voting descriptions. The only thing left is to use AppBadges in the descriptions: https://github.com/aragon/aragon-apps/issues/1040.

PR #717, opened 2019-03-18
Voting: use aragonAPI for React + React hooks

bpierre:
See https://github.com/aragon/aragon.js/pull/259


apps/voting/app/src/vote-hooks.js
```diff
@@ -0,0 +1,90 @@
+import { useState, useEffect } from 'react'
+import { first } from 'rxjs/operators'
+
+// transforms an observable into a promise
+function promisify(observable) {
```
2color: This is unnecessary for the calls that use it. Given how both `external` and normal `call`s use `sendAndObserveResponses` which returns an observable, you can just `toPromise` directly on the observables without the `pipe` and `first`

https://github.com/aragon/aragon.js/blob/f864f7e1c77de27c3b46cc16d38e8dc94d7da665/packages/aragon-api/src/index.js#L173

https://github.com/aragon/aragon.js/blob/e7be88b49bae33e0320efc9abb52abde8b04d906/packages/aragon-rpc-messenger/src/index.js#L90

apps/voting/app/src/script.js
```diff
@@ -262,19 +260,13 @@ function loadVoteSettings() {
...
   })
 }
 
 function loadTokenSymbol(tokenContract) {
   return new Promise((resolve, reject) => {
-    tokenContract
-      .symbol()
-      .first()
-      .subscribe(resolve, reject)
+    tokenContract.symbol().subscribe(resolve, reject)
```
2color: Instead of wrapping with a promise, we can call `.toPromise`, i.e.     `return tokenContract.symbol().toPromise()`

apps/voting/app/src/components/VotePanelContent.js
```diff
@@ -78,7 +79,7 @@ class VotePanelContent extends React.Component {
     if (tokenContract && user) {
       tokenContract
         .balanceOfAt(user, vote.data.snapshotBlock)
-        .first()
+        .pipe(first())
```
2color: This is an external call with a single response. `pipe` and `first` is redundant. If you like, you can also convert to a promise (just syntactic sugar here)

apps/voting/app/src/vote-hooks.js
bpierre: That’s what I was hesitating about, thanks! Just to confirm: does `toPromise()` complete the subscriber the same way `first()` does?

apps/voting/app/src/script.js
bpierre: Much better 👌

apps/voting/app/src/vote-hooks.js
2color: @bpierre See the edit in the original comment.

As for the question, I believe `toPromise` will return a promise that only resolves if the the observable `completes` (ends). If it doesn't end it will never complete.

2color: Confirmed: https://jsbin.com/munemiciyo/1/edit?js,console

apps/voting/app/package.json
```diff
@@ -4,16 +4,19 @@
...
-    "prop-types": "^15.6.0",
-    "react": "^16.5.2",
-    "react-dom": "^16.2.0",
+    "prop-types": "^15.7.2",
+    "react": "^16.8.4",
+    "react-dom": "^16.8.4",
     "react-linkify": "^0.2.2",
     "react-spring": "^5.7.2",
+    "rxjs": "^6.2.1",
+    "rxjs-compat": "^6.2.1",
```
2color: I don't think we need the `rxjs-compat` given the changes in this PR.

https://github.com/ReactiveX/rxjs/blob/master/docs_app/content/guide/v6/migration.md#drop-compat

apps/voting/app/src/vote-hooks.js
bpierre: Thanks!

apps/voting/app/src/components/VotePanelContent.js
```diff
@@ -115,7 +115,7 @@ class VotePanelContent extends React.Component {
 
     this.setState({ loadingCanExecute: true })
     const canExecute = await api.call('canExecute', vote.voteId).toPromise()
-    this.setState({ canExecute, loadingCanExecute: false })
+    this.setState({ loadingCanExecute: false, canExecute })
```
sohkai: Generally I try to keep the "shorthand" properties first (I think there was an ESLint rule on this at some point, but can't find it anymore).

Not sure how consistently we do this, but wondering if you have any thoughts @bpierre @AquiGorka

bpierre: Oh yes we could do that by default, expect maybe when we are dealing with a large number of properties and it makes more sense to group them semantically. In this specific case I wanted to be consistent with the two other `userCanVote`, but we can move the `loadingCanSomething` at the end 👍

AquiGorka: Never needed to sort variables in set state calls, I'll read up to see how much a difference it can make and in the meantime I can follow any convention we agree on.

apps/voting/app/src/components/VotePanelContent.js
```diff
@@ -73,54 +73,49 @@ class VotePanelContent extends React.Component {
...
-          this.setState({ userBalance: adjustedBalance })
-        })
+  loadUserBalance = async (user, vote, tokenContract) => {
+    if (!tokenContract || !user) {
+      return
     }
+
+    const { tokenDecimals } = this.props
+
+    const balance = await tokenContract
```
AquiGorka: What happens if this fails? I really like `async/await` as a clearer syntax but having to add `catch` statements is not as fun..

apps/voting/app/src/vote-hooks.js
```diff
@@ -0,0 +1,79 @@
...
+function usePromiseResult(fn, params, memoParams, defaultValue) {
+  const [result, setResult] = useState(defaultValue)
+  useEffect(() => {
+    let cancelled = false
+    fn(...params).then(value => {
+      if (!cancelled) {
+        setResult(value)
+      }
+    })
+    return () => {
```
AquiGorka: I like the idea of cancelling promises but this mechanism is not being used is it? Do we need it?

apps/voting/app/src/components/VotePanelContent.js
```diff
@@ -390,4 +385,8 @@ const VotingButton = styled(Button)`
   }
 `
 
-export default provideNetwork(VotePanelContent)
+export default props => (
+  <NetworkContext.Consumer>
```
AquiGorka: Is it too big a change to refactor from a `class` to `function` Component?

bpierre: Yes I didn’t want to touch it for now as it is quite self contained, so we can focus on `App.js` in this PR.

apps/voting/app/src/vote-hooks.js
bpierre: We are not cancelling the promise here (even though it could be a good idea to implement it this way), but it is important to prevent the state to be updated if it gets called with another `memoParams`. That way we don’t get into concurrency bugs if e.g. it gets called twice but the response of the first call comes after the second one.

We need it because in this case, because `.call()` is asynchronous.

apps/voting/app/src/components/VotePanelContent.js
bpierre: Yes I agree, `try / catch` are not fun 🙅‍♀️🙅‍♀️🙅‍♀️

I wonder if pattern matching is going to make it less common to throw errors, returning them instead, like [Reason does](https://reasonml.github.io/docs/en/exception) 🤔

But for now `@aragon/api` doesn’t emit errors, so calls never fail.

apps/voting/app/src/components/AppLayout.js
```diff
@@ -32,7 +32,7 @@ const AppLayout = ({
                   (below('medium') ? (
                     <ButtonIcon
                       onClick={mainButton.onClick}
-                      title={mainButton.label}
+                      label={mainButton.label}
```
AquiGorka: ❤️

apps/voting/app/src/components/MenuButton/MenuButton.js
```diff
@@ -11,6 +11,7 @@ export default props => (
       padding: 0 10px 0 20px;
       margin-right: 8px;
     `}
+    label="Open menu"
```
AquiGorka: ❤️ here too :)

apps/voting/app/src/utils-hooks.js
```diff
@@ -0,0 +1,14 @@
...
+export function useNow(updateEvery = 1000) {
+  const [now, setNow] = useState(new Date())
+  useEffect(() => {
+    const timer = setInterval(() => {
+      setNow(new Date())
+    }, updateEvery)
+    return () => {
+      clearInterval(timer)
+    }
+  }, [])
```
AquiGorka: `updateEvery` as dependency?
Love this hook btw.

apps/voting/app/src/components/VotePanelContent.js
sohkai: Probably makes almost 0 difference to the engine, but yes, let's go with:

1. Order semantically
1. Order by shorthand
1. Order by alphabetical

As "soft rules", in that priority 😄

apps/voting/app/src/vote-hooks.js
```diff
@@ -0,0 +1,80 @@
+import { useState, useEffect } from 'react'
+import { useAragonApi } from '@aragon/api-react'
+
+// when `params` are updated, call `fn` and pass the result
+function usePromiseResult(fn, params, memoParams, defaultValue) {
+  const [result, setResult] = useState(defaultValue)
+  useEffect(() => {
+    let cancelled = false
+    fn(...params).then(value => {
```
sohkai: Perhaps we could avoid passing params, and instead just invoke a no-arg function? Adding an extra set of params is quite confusing, and makes it inevitable we'll ask ourselves which one is the called params and which one is the memo set.

Users can just use an arrow function or `.bind()` to make an argument-less function for this.

apps/voting/app/src/vote-hooks.js
```diff
@@ -0,0 +1,80 @@
...
+    return () => {
+      cancelled = true
+    }
+  }, memoParams)
+  return result
+}
+
+async function getUserBalance(vote, userAccount, tokenContract, tokenDecimals) {
+  if (!tokenContract || !userAccount) {
+    return -1
```
sohkai: It seems like the checks in `VotePanelContent` usually check if this is `null` instead of `-1`. We should make sure to change those checks to use `-1`.

Should we also add a check for `!vote`?

apps/voting/app/src/vote-hooks.js
```diff
@@ -0,0 +1,80 @@
...
+
+  // If the account is not present, we assume the account is not connected.
+  if (!userAccount) {
+    return vote.data.open
+  }
+
+  return api.call('canVote', vote.voteId, userAccount).toPromise()
+}
+
+async function getCanExecute(vote, api) {
```
sohkai: Technically, there's no need for both `getCanVote()` and `getCanExecute()` to be declared `async`

apps/voting/app/src/vote-hooks.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+  return api.call('canExecute', vote.voteId).toPromise()
+}
+
+export const useCurrentVoteData = (
+  vote,
+  userAccount,
+  tokenContract,
+  tokenDecimals
+) => {
```
sohkai: Perhaps an options object with these keys would be easier to read and debug 😄

apps/voting/app/src/vote-hooks.js
```diff
@@ -0,0 +1,80 @@
...
+  const { api } = useAragonApi()
+  return {
```
sohkai: Perhaps we could bail here with the defaults if `!vote`, to simplify the logic afterwards?

apps/voting/app/src/App.js
```diff
@@ -1,352 +1,276 @@
...
+              // Render text fields
+              descriptionNode: voteTextNode(vote.data.description),
+              metadataNode: voteTextNode(vote.data.metadata),
+            },
+            userAccountVote: voteTypeFromContractEnum(
+              userAccountVotes.get(vote.voteId)
+            ),
+          }))
+        : appState.votes,
+    [appStateReady, cacheIdFromVotes(appState.votes), userAccountVotes]
```
sohkai: ⚠️ I think this caching for this effect is too aggressive and will result in confusing results.

Using `cacheIdFromVotes()` means `votes` will only get recomputed if a new vote is added. **But** it will not be recomputed if an existing vote's data was modified (e.g. its voting values change). Nor would it recompute whether the vote was still open or not, since `now`'s not in the dependencies list.

Perhaps we can instead apply caching for the rendered metadata and description nodes using a `Map`?

apps/voting/app/src/App.js
```diff
@@ -1,352 +1,276 @@
...
-  handleResolveLocalIdentity = address => {
-    return this.props.app.resolveAddressIdentity(address).toPromise()
-  }
-  handleShowLocalIdentityModal = address => {
-    return this.props.app
-      .requestAddressIdentityModification(address)
-      .toPromise()
-  }
+  // update user account votes
+  useEffect(() => {
```
sohkai: ⚠️ Unfortunately, the caching for this effect is also prone to failure.

Since we use `cacheIdFromVotes()` again, this will only update if we add a vote. It won't be re-evaluted if the user changes their vote.

Rather than the current dependencies list, we can improve the caching and also reduce the number of network requests by:

- Applying some throttling (most useful for initial loads)
- Once we've fetched the voter status for a vote that's executed or has ended (`!isVoteOpen()`), we know, _for sure_, that you can't change your vote any more and don't need to re-evaluate it again

------------

Question about `useEffect()`: does it let us know which of the dependencies changed?

In our case, if `connectedAccount()` changes, we should start with a completely new `Map`. It's probably not a problem to recreate the `Map` each on each computation, but it'd be nice to just keep reusing the same `Map`.

apps/voting/app/src/App.js
```diff
@@ -1,352 +1,276 @@
...
-            : [],
-        }
-      })
-    ),
-  {}
-)(App)
+export default () => (
+  <AragonApi reducer={appStateReducer}>
+    <App />
+  </AragonApi>
```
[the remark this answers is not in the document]
bpierre: Yes, I moved other providers at the end now: https://github.com/aragon/aragon-apps/blob/voting-react-api/apps/voting/app/src/App.js#L110

And I think it’s a nice way to see visualize it. Let’s move it here in the other apps too :+1:

apps/voting/app/src/components/VotePanelContent.js
```diff
@@ -1,20 +1,21 @@
...
   SidePanelSeparator,
-  Countdown,
+  SidePanelSplit,
   Text,
-  blockExplorerUrl,
   theme,
 } from '@aragon/ui'
+import { useAragonApi } from '@aragon/api-react'
+import { useCurrentVoteData } from '../vote-hooks'
+import { NetworkContext } from '../app-contexts'
```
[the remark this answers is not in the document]
bpierre: > See above, I think we can just use useNetwork from api-react?

:+1: 

> We could also move this down into the LocalIdentityBadge, so we don't need to manually handle the network when we use that component.

Yes nice idea! Done :+1:

apps/voting/app/src/components/VotePanelContent.js
```diff
@@ -27,123 +28,44 @@ import SummaryBar from './SummaryBar'
...
+    }, [])
+    const handleNoClick = useCallback(() => {
+      onVote(vote.voteId, VOTE_NAY)
+    }, [])
+    const handleYesClick = useCallback(() => {
+      onVote(vote.voteId, VOTE_YEA)
+    }, [])
+    const handleExecuteClick = useCallback(() => {
+      onExecute(vote.voteId)
+    }, [])
```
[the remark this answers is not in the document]
bpierre: Not it doesn’t get recreated, but `vote` comes from the current scope, so I thought there was no need… except React recommends to always list the dependencies. Adding them :+1:

apps/voting/app/src/App.js
```diff
@@ -1,352 +1,276 @@
...
 import Votes from './screens/Votes'
-import tokenAbi from './abi/token-balanceOfAt.json'
 import VotePanelContent from './components/VotePanelContent'
 import NewVotePanelContent from './components/NewVotePanelContent'
 import AutoLink from './components/AutoLink'
 import AppLayout from './components/AppLayout'
 import NewVoteIcon from './components/NewVoteIcon'
+
+import { NetworkContext, SettingsContext } from './app-contexts'
 import { IdentityProvider } from './components/IdentityManager/IdentityManager'
```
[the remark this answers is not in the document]
bpierre: I made it a normal JS module, as the component (the provider) is only a small part of this module: https://github.com/aragon/aragon-apps/blob/voting-react-api/apps/voting/app/src/App.js#L12-L13

apps/voting/app/src/App.js
```diff
@@ -1,352 +1,276 @@
...
-            pctBase: pctBaseNum,
-            tokenDecimals: tokenDecimalsNum,
-          },
+              <SidePanel
+                title="New Vote"
+                opened={createVoteVisible}
+                onClose={handleCreateVoteClose}
+              >
+                <NewVotePanelContent
+                  opened={createVoteVisible}
```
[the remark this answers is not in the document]
bpierre: Yes that could be nice, good idea!

apps/voting/app/src/App.js
```diff
@@ -1,352 +1,276 @@
...
-        voteTime,
-      },
-    }
-  }
+function voteTextNode(description) {
+  return description ? (
+    <AutoLink>
+      {description.split('\n').map((line, i) => (
+        <React.Fragment key={i}>
+          {shortenAddresses(line)}
```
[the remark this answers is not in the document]
bpierre: It looks a bit heavy, but I think this is an opportunity to introduce a small version of this badge, and keep it like this in the meantime for consistency. /cc @dizzypaty 

Screenshot from 2019-04-09 01-33-49
Screenshot from 2019-04-09 01-35-49

Also I wonder if there is anything we could do to address the problem of being wrong in some (probably rare) cases, e.g. if an organization on a network refers to an address on another network (in a question)?

Maybe we could have an optional, universal way of making addresses and transactions absolute, for advanced users?

e.g. an address could be written in these different ways, anywhere in the Aragon apps (inspired by Markdown [autolinks](https://github.github.com/gfm/#autolinks)):

```
# relative to the current network (automatically recognized in most cases)
My address is 0x87eF64AA97226cc21FC4C7e24C0D7546a1692024.
My address is p.aragonid.eth.

# same but explicit
My address is <0x87eF64AA97226cc21FC4C7e24C0D7546a1692024>.
My address is <p.aragonid.eth>.

# bound to a specific network, by name
My address is <mainnet:0x87eF64AA97226cc21FC4C7e24C0D7546a1692024>.
My address is <rinkeby:p.aragonid.eth>.

# by chainID too maybe
My address is <chain1:0x87eF64AA97226cc21FC4C7e24C0D7546a1692024>
My address is <chain4:p.aragonid.eth>
```

apps/voting/app/src/App.js
```diff
@@ -1,352 +1,276 @@
...
-    ) {
-      this.loadUserAccountVotes(nextProps.userAccount, nextProps.votes)
-    }
-  }
+  const [createVoteVisible, setCreateVoteVisible] = useState(false)
+  const [currentVoteId, setCurrentVoteId] = useState(-1)
+  const [tokenContract, setTokenContract] = useState(null)
+  const [voteVisible, setVoteVisible] = useState(false)
+  const [voteSidebarOpened, setVoteSidebarOpened] = useState(false)
+  const [userAccountVotes, setUserAccountVotes] = useState(new Map())
```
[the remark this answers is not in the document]
bpierre: It should be better now, see [`usePanelState()`](https://github.com/aragon/aragon-apps/blob/voting-react-api/apps/voting/app/src/voting-app.js#L195-L225) and [`useSelectedVotePanel()`](https://github.com/aragon/aragon-apps/blob/voting-react-api/apps/voting/app/src/voting-app.js#L227-L239)

apps/voting/app/src/vote-hooks.js
bpierre: Agreed, this hook has changed a little and it now lets users call the function themselves: https://github.com/aragon/aragon-apps/blob/voting-react-api/apps/voting/app/src/utils-hooks.js#L16-L32

apps/voting/app/src/vote-hooks.js
bpierre: They both return either promises or values, so `async` is needed to return a promise in every case, or am I missing something?

Also I think having `async` in front of a function returning a promise is a nice touch for readability (it doesn’t hurt performances since promises auto unwrap). It also ensures that no errors are thrown outside the returned promise.

apps/voting/app/src/vote-hooks.js
bpierre: This call has been simplified :laughing: https://github.com/aragon/aragon-apps/blob/voting-react-api/apps/voting/app/src/voting-app.js#L86

apps/voting/app/src/vote-hooks.js
bpierre: [Forbidden :no_good_man: :no_entry: :no_entry_sign: ](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level)

Doing so would prevent React to keep track of the hooks and their corresponding components.

apps/voting/app/src/App.js
bpierre: I removed the caching for now, let’s add a throttling mechanism to `@aragon/api` / `@aragon/api-react` as discussed offline :+1:

apps/voting/app/src/App.js
bpierre: Same here, the caching has been removed.

> Question about useEffect(): does it let us know which of the dependencies changed?

It doesn’t, but there are different ways to achieve this if needed (e.g. using several `useEffect()` with `useState()`).

> In our case, if connectedAccount() changes, we should start with a completely new Map. 

It is the case yes: if either `api`, `votes` or `connectedAccount` changes, a new `Map` gets created.

> It's probably not a problem to recreate the Map each on each computation, but it'd be nice to just keep reusing the same Map.

I’m not sure I understand what you meant here? Only one `Map` gets created every time we either have an update of `connectedAccount` or `votes`, and it only contains the information needed for the connected account: `vote ID => vote status`.

apps/voting/app/src/App.js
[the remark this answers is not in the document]
sohkai: Made https://github.com/aragon/aragon-ui/issues/347 for it :).

apps/voting/app/src/App.js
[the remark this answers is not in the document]
sohkai: Mmm that would be interesting, kind of like multiaddr.

Let's not worry about this too much for now, until multi-chain scenarios (e.g. with Polkadot) start showing up :).

---------

And on the `IdentityBadge`, I do agree it looks quite funny in the description. Let's relegate this to https://github.com/aragon/aragon-apps/issues/510 then.

apps/voting/app/src/vote-hooks.js
sohkai: Ahh good point, didn't realize they were returning normal values too for the default case. And I do agree adding the `async` is a nice syntactic touch, let's keep doing that 👍.

apps/voting/app/src/App.js
sohkai: Not having to recreate the map is an absolutely tiny optimization, since we could technically reuse it each time if the connected account hasn't changed.

But recreating it each time is also good for being more immutable / React-like :).

apps/voting/app/src/App.js
```diff
@@ -1,276 +1,115 @@
...
       <Main assetsUrl="./aragon-ui">
-        <IdentityProvider
-          onResolve={resolveLocalIdentity}
-          onShowLocalIdentityModal={showLocalIdentityModal}
+        <AppLayout
+          title="Voting"
+          mainButton={{
+            label: 'New vote',
+            icon: <NewVoteIcon />,
+            onClick: newVotePanel.open,
```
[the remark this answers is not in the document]
bpierre: Yes I think we can use `onX` for these cases, since it doesn’t conflict anymore :+1:

apps/voting/app/src/App.js
```diff
@@ -1,276 +1,115 @@
...
-import { NetworkContext, SettingsContext } from './app-contexts'
-import { IdentityProvider } from './components/IdentityManager/IdentityManager'
-import { isVoteOpen, voteTypeFromContractEnum } from './vote-utils'
-import { shortenAddress, transformAddresses } from './web3-utils'
-import { useNow } from './utils-hooks'
-import appStateReducer from './app-state-reducer'
+import { IdentityProvider } from './identity-manager'
+import { SettingsProvider } from './vote-settings-manager'
+import { transformAddresses } from './web3-utils'
+import { VotingAppProvider, useVotingApp } from './voting-app'
```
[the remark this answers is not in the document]
bpierre: Before I renamed it `VotingAppProvider` / `useVotingApp` / `voting-app.js`, I was using something really similar to what you suggest: `AppLogicProvider` / `useAppLogic` / `app-logic.js`. I changed it because I feared it was too generic, but the idea was to avoid adding another “appState”, and to have a name as generic as `App.js` so we can use the same in every app.

Let me know if you think that “voting” should stay in the name, otherwise I think we can use these names. :+1:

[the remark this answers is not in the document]
sohkai: I do like the simplicity and obviousness of `AppLogicProvider` / `useAppLogic` / `app-logic.js`.

My main goal is that it should be immediately obvious to someone new what a particular file is used for. `voting-app.js` is quite ambiguous, and also somewhat redundant.

apps/voting/app/src/App.js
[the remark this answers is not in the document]
bpierre: I ended up using `requestOpen()` / `requestClose()`: it was a bit weird to use “on” for a function that can be called from anywhere.

apps/voting/app/src/App.js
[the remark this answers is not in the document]
bpierre: Agreed, done :+1:

apps/voting/app/src/identity-manager.js
```diff
@@ -0,0 +1,70 @@
...
+  const { updates$ } = React.useContext(IdentityContext)
+
+  const handleNameChange = useCallback(metadata => {
+    setName(metadata ? metadata.name : null)
+  }, [])
+
+  const handleShowLocalIdentityModal = address => {
+    // Emit an event whenever the modal is closed (when the promise resolves)
+    return showLocalIdentityModal(address).then(() => updates$.next(address))
+  }
```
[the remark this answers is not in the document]
bpierre: Yes, I removed the references to a modal: https://github.com/aragon/aragon-apps/pull/717/commits/54f6f2a67953ac976d79d9e1db232e0e1346fb16

apps/voting/app/src/App.js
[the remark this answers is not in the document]
sohkai: Let's remember this naming rule for the future 😄

apps/voting/app/src/components/VotingCard/VotingCard.js
```diff
@@ -37,27 +63,69 @@ class VotingCard extends React.Component {
...
+      prevVote.voteId === nextVote.voteId &&
+      prevVote.connectedAccountVote === nextVote.connectedAccountVote &&
+      prevVoteData.endDate === nextVoteData.endDate &&
+      prevVoteData.open === nextVoteData.open &&
+      prevVoteData.metadata === nextVoteData.metadata &&
+      prevVoteData.description === nextVoteData.description &&
+      prevVote.numData.votingPower === nextVote.numData.votingPower &&
+      prevVote.numData.yea === nextVote.numData.yea &&
+      prevVote.numData.nay === nextVote.numData.nay
+    )
```
bpierre: @sohkai now 1000x faster :laughing:

apps/voting/app/src/components/VotingCard/VotingCard.js
```diff
@@ -37,27 +61,55 @@ class VotingCard extends React.Component {
...
+      prevVote.data.description === nextVote.data.description &&
+      prevVote.numData.votingPower === nextVote.numData.votingPower &&
+      prevVote.numData.yea === nextVote.numData.yea &&
+      prevVote.numData.nay === nextVote.numData.nay
+    )
   }
+)
+
+VotingCard.defaultProps = {
+  options: [],
```
sohkai: Don't think we need `options` here

PR #719, opened 2019-03-19
Token manager fixes

bpierre:
image

Changes:

- Layout fixes to remove the horizontal scrollbar on small viewports (320px).
- Fix the spacing between the “You” badge and the Identity badge.
- Add the “You” badge in the sidebar.
- Upgrade `@aragon/ui` to `0.32.0`.
- IdentityBadge: always shorten + set the `connectedAccount` prop.
- Table: use the `noSideBorders` prop.

### DIFF
--- a/apps/token-manager/app/src/components/SideBar.js
+++ b/apps/token-manager/app/src/components/SideBar.js
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from 'styled-components'
-import { IdentityBadge, Text, Viewport, breakpoint, theme } from '@aragon/ui'
+import { IdentityBadge, Text, breakpoint, theme } from '@aragon/ui'
 import provideNetwork from '../provide-network'
 import { formatBalance, stakesPercentages } from '../utils'
 import TokenBadge from './TokenBadge'
--- a/apps/token-manager/app/src/components/SideBar.js
+++ b/apps/token-manager/app/src/components/SideBar.js
@@ -4,6 +4,7 @@ import { IdentityBadge, Text, Viewport, breakpoint, theme } from '@aragon/ui'
 import provideNetwork from '../provide-network'
 import { formatBalance, stakesPercentages } from '../utils'
 import TokenBadge from './TokenBadge'
+import You from './You'
 
 const DISTRIBUTION_ITEMS_MAX = 7
 const DISTRIBUTION_COLORS = [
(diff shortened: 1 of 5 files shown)


apps/token-manager/app/src/components/SideBar.js
```diff
@@ -4,6 +4,7 @@ import { IdentityBadge, Text, Viewport, breakpoint, theme } from '@aragon/ui'
 import provideNetwork from '../provide-network'
 import { formatBalance, stakesPercentages } from '../utils'
 import TokenBadge from './TokenBadge'
+import You from './You'
```
sohkai: I think we can remove the `Viewport` import

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -62,11 +63,15 @@ class Holders extends React.Component {
                           groupmode={groupMode}
                         />
                         {!groupMode && (
-                          <StyledTableHeader title="Balance" align="right" />
+                          <StyledTableHeader
+                            title={<div css="margin-left: -20px">Balance</div>}
```
sohkai: This seems quite weird; did we modify the `TableHeader` to include internal padding? Could this be moved to the `StyledTableHeader` definition?

bpierre: I know 😩 it’s really specific to this app, and the only way I found to gain enough space to remove the horizontal scrollbar without compromising the layout too much, by letting the “Balance” header overlapping 20px on the “Holder” header:

image

Even at `320px` there is plenty of space in the “Holder” header, so it shouldn’t be a problem… but I should add a comment at the very least.

Also I don’t think we should add a prop for now, it’s more something that should be avoided when the screen gets designed.

sohkai: Maybe we could rename it to something else, like `BLN`. I think it's pretty clear what this column is.

bpierre: I agree it should stay pretty clear, but still not be as clear as having it entirely, and we have enough space to do it? Do you see a problem with using a negative margin here?

sohkai: Looks a bit weird with the "you" sticking half way into the balance header 😄.

Less of a problem if you're not the first token holder though.

bpierre: Oh I see, I thought it was about the code! I’d really like to avoid abbreviations if we have other options. What would you think about having it left aligned & fusioned with the next cell on small viewports?

image

sohkai: Maybe this is better, especially since the `HOLDER` header is left-aligned (and technically the action buttons _modify_ the balance).

bpierre: Done!

PR #721, opened 2019-03-20
Voting: Upgrade rxjs to v6

2color:
## What
- Add `rxjs` as a dependency
- Improve dev experience by watching the script in dev/watch mode too
- Make rxjs v6 code changes where necessary
- Instead of using rxjs pipe and operators, operate on resolved value with
  native javascript datatypes

This PR is in the same vein as:
 - https://github.com/aragon/aragon-apps/pull/698
 - https://github.com/aragon/aragon-apps/pull/716


apps/voting/app/package.json
```diff
@@ -41,8 +41,9 @@
   "scripts": {
```
sohkai: Let's also add rxjs as a dependency

apps/voting/app/src/script.js
```diff
@@ -195,12 +196,10 @@ async function loadVoteDescription(vote) {
 }
 
 function loadVoteData(voteId) {
-  return new Promise(resolve => {
-    app
-      .call('getVote', voteId)
-      .first()
-      .subscribe(vote => resolve(loadVoteDescription(marshallVote(vote))))
-  })
+  app
```
sohkai: This should return

apps/voting/package.json
```diff
@@ -56,6 +56,7 @@
   },
   "dependencies": {
     "@aragon/apps-shared-minime": "1.0.0",
-    "@aragon/os": "4.0.0"
+    "@aragon/os": "4.0.0",
+    "rxjs": "^6.2.1"
```
sohkai: Oops, this isn't the right place to add it :smile:

PR #722, opened 2019-03-20
Token manager: custom labels

AquiGorka:
Implements Custom Labels in Token Manager.

Needs:
- Aragon/ui with extended IdentityBadge
- https://github.com/aragon/aragon-ui/pull/337 for styling
- https://github.com/aragon/aragon/pull/656 for functionality

## Todo

- [x] Bump minor version of `@aragon/ui` once released with updated IdentityBadge
- [x] Bump version of `@aragon/api` once released with identities functionality


apps/token-manager/app/src/components/CustomLabelIdentityBadge/CustomLabelIdentityBadge.js
```diff
@@ -0,0 +1,86 @@
...
+      .then(handleResolve)
+      .catch(e => {
+        /* user cancelled modify intent */
+      })
+  }
+  React.useEffect(() => {
+    handleResolve(address)
+    const subscription = updates$.subscribe(addr => {
+      if (addr === address) {
+        handleResolve(address)
```
2color: I don't think there's a need to pass `address` to `handleResolve`

apps/token-manager/app/src/components/CustomLabelIdentityBadge/CustomLabelIdentityBadge.js
```diff
@@ -0,0 +1,86 @@
...
+      }
+    })
+    return () => subscription.unsubscribe()
+  }, [])
```
2color: Should we add `address` to the array passed in the second argument? This could prevent redundant resolutions.

AquiGorka: This effect only happens once:

> If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument.

[Hooks Effect](https://reactjs.org/docs/hooks-effect.html)

AquiGorka:
@2color what do you think about this approach: 7ebd442e32e902e131cc50e7cf833bb04dfffb7e

apps/token-manager/app/src/components/CustomLabelIdentityBadge/CustomLabelIdentityBadge.js
2color: Thanks that clarifies it

apps/token-manager/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+  React.useEffect(() => {
+    handleResolve()
+    const subscription = updates$.subscribe(updatedAddress => {
+      if (updatedAddress.toLowerCase() === address.toLowerCase()) {
+        handleResolve()
+      }
+    })
+    return () => subscription.unsubscribe()
+  }, [])
```
bpierre: Same suggestion than for Finance, but I notice now that we should also add `name` as `handleResolve` is using it:

```suggestion
  }, [address, name])
```

Maybe `handleResolve` could take `name` as a parameter, to make it more explicit?

2color: Are you sure? I think we just need to pass the address. The name – which in turn is set as the label – is the result of the async side-effect.

bpierre: Oh you’re right, I thought `name` was coming from somewhere else. What could change would be `resolve` though, but we might not want to react to `resolve` changes.

apps/token-manager/app/src/components/HolderRow.js
```diff
@@ -83,18 +84,27 @@ class HolderRow extends React.Component {
...
           </ContextMenu>
         </TableCell>
-      </TableRow>
+      </StyledTableRow>
     )
   }
 }
 
+const StyledTableRow = styled(TableRow)`
+  td:first-child {
```
sohkai: Perhaps we could change the first `TableCell` instead of adding a style for the entire row and then using css to select the first cell?

apps/token-manager/app/src/components/SideBar.js
```diff
@@ -109,7 +110,8 @@ class SideBar extends React.Component {
               <StakesListItem key={name}>
                 <span>
                   <StakesListBullet style={{ background: color }} />
-                  <IdentityBadge
+                  <LocalIdentityBadge
+                    address={name}
```
sohkai: It does seem quite weird that we use both the `entity` and `address` props; perhaps we could change the `LocalIdentityBadge` to use `entity` instead?

apps/token-manager/app/src/components/HolderRow.js
AquiGorka: I agree yet I'm sure that's where I started and was not able to override the style and had to do this - I'll double check.

apps/token-manager/app/src/components/SideBar.js
2color: I applied this change to voting https://github.com/aragon/aragon-apps/pull/728/commits/a8013469d2e8abc6cb19c1d690595e9c5820642e

apps/token-manager/app/src/components/HolderRow.js
AquiGorka: Updated :)

apps/token-manager/app/src/components/SideBar.js
AquiGorka: Applied here too :)

PR #723, opened 2019-03-20
Track @next contract changes (Aragon 0.7-part.2)

sohkai:
Completed security review.


tintinweb:
As a result of the change review we did not find any blocking issues. Please find a summary of our notes below. 

* comments directly related to the changes in this PR are discussed inline with the commit (see `comment-inline` tag below). 
* general observations not directly related to this change are provides as a comment

#### Commit
* [Token Manager: enforce all token controller methods to only be access… ](https://github.com/aragon/aragon-apps/pull/723/commits/4c9b720e8a887ed5e045872dc8ce7dd4fd1b7b45)
  * (ℹ) (`comment-inline`) token callbacks (caller is `minime`) can be set external
onTransfer
onApprove
proxyPayment
  * (ℹ) (`comment-inline`) proxyPayment - consider throwing an exception in proxyPayment already instead of relying on the caller (`minime-token`) to throw if the method returns false.
* [Finance: optimize SafeMath (#706)](https://github.com/aragon/aragon-apps/pull/723/commits/d5feb331a7ce5a452753c9ebc2f9edf1fe7d3714)
  * (ℹ) Unclear if it is worth the optimisation effort. Minor change. Security: minor risk for the current changes. More secure using Safemath (future changes)
* [Voting: mock time in tests rather than use timetravel (#737) ](https://github.com/aragon/aragon-apps/pull/723/commits/4e89224872a9b223d87e1ce92e3a6b7c0ec17d9d)
  * (ℹ) (in mocks implementing mockTime) does not test original `getTimestamp[64]()` anymore as it is mocked for testing arbitrary time. Problems in the underlying timehelpers might not be caught by this test. Make sure timehelpers is properly tested (also as part of apps) (also see changes in Token Manager: use TimeHelpers and mock time in tests (#738))
* [Finance, Survey, Token Manager, Voting: Clarify internal protection s… ](https://github.com/aragon/aragon-apps/pull/723/commits/4eefbb25c853694d5c115f0e730c6b2fd5d0f923)
  * Note - communicate the statement in commit message (e.g. documentation)
* [Token Manager: use TimeHelpers and mock time in tests (#738)](https://github.com/aragon/aragon-apps/pull/723/commits/8323e4c3da692c3c3dd8cbec4b4b74f4c9b2165f)
  * (ℹ) -  unused imports after this commit --> see general observations/TokenManager


#### General Observations

* Finance
  * (ℹ) (`comment-inline`) [finance#L690](https://github.com/aragon/aragon-apps/blob/next/apps/finance/contracts/Finance.sol#L690) could be `<=` to state that end-time cannot be start time (logically impossible as period > 1days)
  * (ℹ) getMaxPeriodTransitions can be restricted to pure (const)
* TokenManager
  * (ℹ) unused imports
    * uint256helpers
    * ERC20
* Voting
  * Note - `canforward` passes the `evmscript` to `canperform` but it is never checked. Note -  unnamed parameter in public method
[voting#L188-L196](https://github.com/aragon/aragon-apps/blob/next/apps/voting/contracts/Voting.sol#L188-L196)
* Code Complexity
  * Note - The recent changes increased the overall complexity of the codebase (tradeoff gas/security; making sure all paths are properly checked).
* Highlighting/Tagging unsafe functions
  * Note - Make sure to consistently mark unsafe functions that are not performing security checks and communicate how developers can quickly see whether functions are safe or not (e.g. add comments to function description)

* Future Migration Note - Compatibility with solidity >= 0.5.0
  * use of reserved keywords as identifiers (see [solidity-docs](https://solidity.readthedocs.io/en/latest/miscellaneous.html#reserved-keywords))
[finance#L120-L121](https://github.com/aragon/aragon-apps/blob/next/apps/finance/contracts/Finance.sol#L120-L121)
[voting#L62](https://github.com/aragon/aragon-apps/blob/next/apps/voting/contracts/Voting.sol#L62)
```
Compiling your contracts...
===========================
Error: CompileError: ParsedContract.sol:120:105: ParserError: Expected ',' but got reserved keyword 'reference'
    event NewPayment(uint256 indexed paymentId, address indexed recipient, uint64 maxExecutions, string reference);
```

apps/token-manager/contracts/TokenManager.sol
```diff
@@ -231,46 +235,38 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
...
+        return _isBalanceIncreaseAllowed(_to, _amount) && _transferableBalance(_from, getTimestamp()) >= _amount;
     }
 
     /**
     * @dev Notifies the controller about an approval allowing the controller to react if desired
+    *      Initialization check is implicitly provided by `onlyToken()`.
     * @return False if the controller does not authorize the approval
     */
-    function onApprove(address, address, uint) public returns (bool) {
+    function onApprove(address, address, uint) public onlyToken returns (bool) {
```
tintinweb: token callbacks (caller is minime) can be set external
onTransfer
onApprove
proxyPayment

apps/token-manager/contracts/TokenManager.sol
```diff
@@ -231,46 +235,38 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
...
+    * @dev Called when ether is sent to the MiniMe Token contract
+    *      Initialization check is implicitly provided by `onlyToken()`.
     * @return True if the ether is accepted, false for it to throw
     */
-    function proxyPayment(address) public payable returns (bool) {
-        // Sender check is required to avoid anyone sending ETH to the Token Manager through this method
-        // Even though it is tested, solidity-coverage doesnt get it because
-        // MiniMeToken is not instrumented and entire tx is reverted
-        require(msg.sender == address(token), ERROR_PROXY_PAYMENT_WRONG_SENDER);
+    function proxyPayment(address) public payable onlyToken returns (bool) {
```
tintinweb: consider throwing an exception in proxyPayment already instead of relying on the caller (minime-token) to throw if the method returns false (unless it is a defined interfaces intended to be used that way)

apps/finance/contracts/Finance.sol
```diff
@@ -504,132 +552,151 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
...
+        // There should be no way for this to overflow since each period is at least one day
+        uint64 newPeriodId = periodsLength++;
+
+        Period storage period = periods[newPeriodId];
+        period.startTime = _startTime;
+
+        // Be careful here to not overflow; if startTime + periodDuration overflows, we set endTime
+        // to MAX_UINT64 (let's assume that's the end of time for now).
+        uint64 endTime = _startTime + settings.periodDuration - 1;
+        if (endTime < _startTime) { // overflowed
```
tintinweb: could be <= to state that end-time cannot be start time (logically impossible as period > 1days as per check init/setPeriodDuration)

apps/token-manager/contracts/TokenManager.sol
sohkai: Not quite sure if I understood the exception:

> unless it is a defined interfaces intended to be used that way

We _are_ relying on the MiniMe token to revert on this false return, but this can only be called from the token (and should only be when it receives ETH).

Throwing an error ourselves seems a bit redundant, and we more or less completely trust the MiniMe implementation in the Token Manager.

sohkai:
> Voting: mock time in tests rather than use timetravel (#737) 

> Token Manager: use TimeHelpers and mock time in tests (#738)

Agreed, we are making changes to fix these mocks now :smile: : see https://github.com/aragon/aragon-apps/pull/766.

-------------

> Voting

> Note - canforward passes the evmscript to canperform but it is never checked.
> Note - unnamed parameter in public method voting#L188-L196

This is to conform with the `IForwarder`'s `canForward()` interface. Technically we could pass a new, empty bytes array, but that'd cost a bit more.

-------------

> Future Migration Note - Compatibility with solidity >= 0.5.0

Ouch :(. We'll make this change in the future, since it requires a frontend change to use the new variable name in the ABI.

apps/token-manager/contracts/TokenManager.sol
sohkai: See https://github.com/aragon/aragon-apps/pull/770.

apps/token-manager/contracts/TokenManager.sol
```diff
@@ -19,7 +19,6 @@ import "@aragon/apps-shared-minime/contracts/MiniMeToken.sol";
 
 contract TokenManager is ITokenController, IForwarder, AragonApp {
     using SafeMath for uint256;
-    using Uint256Helpers for uint256;
```
sohkai: See https://github.com/aragon/aragon-apps/pull/771 for removing unused imports.

apps/finance/contracts/Finance.sol
sohkai: Not super sure about this change; the `<` check is the "safe math" check, so it might be less obvious as `<=`.

apps/finance/contracts/Finance.sol
```diff
@@ -717,25 +786,52 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
...
+    // Syntax sugar
+
+    function _arr(address _a, address _b, uint256 _c, uint256 _d, uint256 _e, uint256 _f) internal pure returns (uint256[] r) {
+        r = new uint256[](6);
+        r[0] = uint256(_a);
+        r[1] = uint256(_b);
+        r[2] = _c;
+        r[3] = _d;
+        r[4] = _e;
+        r[5] = _f;
```
facuspagnuolo: Should we move this one to `ACLSyntaxSugar`?

sohkai: I had debated this, but this is such a super specific instance that I don't think it'd make sense to add back in `ACLSyntaxSugar`. Apps that have giant params lists like this will need their own syntax sugar :).

apps/finance/contracts/Finance.sol
```diff
@@ -504,132 +552,151 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
...
-        // If it is an external deposit, check that the assets are actually transferred
-        // External deposit will be false when the assets were already in the Finance app
-        // and just need to be transferred to the vault
-        if (_isExternalDeposit) {
-            if (_token != ETH) {
-                // Get the tokens to Finance
-                require(ERC20(_token).safeTransferFrom(msg.sender, this, _amount), ERROR_TOKEN_TRANSFER_FROM_REVERTED);
-            } else {
-                // Ensure that the ETH sent with the transaction equals the amount in the deposit
-                require(msg.value == _amount, ERROR_VALUE_MISMATCH);
```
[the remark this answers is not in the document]
sohkai: Yes, this was a result of "bubbling" the parameter checks to the "top" of the stack.

It's now covered by [`deposit()`](https://github.com/aragon/aragon-apps/blob/a66f8762589e27a0b0313a444c6f13fec26b5267/apps/finance/contracts/Finance.sol#L202) directly, and in the [fallback](https://github.com/aragon/aragon-apps/blob/a66f8762589e27a0b0313a444c6f13fec26b5267/apps/finance/contracts/Finance.sol#L158), no check is needed.

apps/finance/contracts/Finance.sol
```diff
@@ -172,9 +176,9 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
         require(_periodDuration >= 1 days, ERROR_INIT_PERIOD_TOO_SHORT);
         settings.periodDuration = _periodDuration;
```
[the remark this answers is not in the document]
sohkai: For sake of consistency with other `require()` checks, I've opted to keep this check duplicated, since it's a parameter check.

However, I've modified it so it uses a constant instead of a magic number in https://github.com/aragon/aragon-apps/pull/781.

apps/token-manager/contracts/TokenManager.sol
tintinweb: 👍 looks good.

apps/finance/contracts/Finance.sol
tintinweb: agree, might be better to keep this as a strict overflow check.

PR #727, opened 2019-03-21
Voting: rxjs upgrade fixes

2color:
## What 

Bug fixes that were introduced by a couple of critical changes not committed to #721.

This PR addresses thouse

### DIFF
--- a/apps/voting/app/src/App.js
+++ b/apps/voting/app/src/App.js
@@ -2,6 +2,7 @@ import React from 'react'
 import PropTypes from 'prop-types'
 import { Main, SidePanel, observe } from '@aragon/ui'
 import BN from 'bn.js'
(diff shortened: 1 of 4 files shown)


apps/voting/app/src/App.js
```diff
@@ -271,62 +272,66 @@ class App extends React.Component {
...
   observable =>
-    observable.map(state => {
-      const appStateReady = hasLoadedVoteSettings(state)
-      if (!appStateReady) {
-        return {
-          ...state,
-          appStateReady,
+    observable.pipe(
+      map(state => {
+        console.log(`state ${state}`)
```
AquiGorka: Console logs here

2color: Just cleaned those :)

PR #728, opened 2019-03-21
Voting: Custom Labels / Local Identities 

2color:
## What
- Local Identities (aka Custom Labels) in the Voting App.

## Dependencies in other repos
- [@aragon/ui](https://github.com/aragon/aragon-ui/commit/c281eefddf27b0650caa9f4a1d8d4874ffd69c08) latest master with updated IdentityBadge
- Aragon client - https://github.com/aragon/aragon/pull/656 
- Functionality in `@aragon/wrapper` and `@aragon/api` https://github.com/aragon/aragon.js/pull/253

## Nomenclature
- For consistency I tried to use the term **identity** 
- Only in the badge do I name things with **label** to reflect its actual use in the UI toolkit (the `IdentityBadge` component expects a `customLabel` prop)

## Todo before merging

- [x] Bump minor version of `@aragon/ui` once released with updated IdentityBadge
- [x] Bump version of `@aragon/api` once released with identities functionality


apps/voting/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
```diff
@@ -0,0 +1,86 @@
...
+  }
+  React.useEffect(() => {
+    handleResolve()
+    const subscription = updates$.subscribe(updatedAddress => {
+      if (updatedAddress.toLowerCase() === address.toLowerCase()) {
+        handleResolve()
+      }
+    })
+    return () => subscription.unsubscribe()
+  }, [])
```
bpierre: Same than https://github.com/aragon/aragon-apps/pull/722#discussion_r268723603:

```suggestion
  }, [name, address])
```

apps/voting/app/src/App.js
```diff
@@ -212,60 +222,67 @@ class App extends React.Component {
...
-                ready={voteSidebarOpened}
-                tokenContract={tokenContract}
-                tokenDecimals={tokenDecimals}
-                tokenSymbol={tokenSymbol}
-                onVote={this.handleVote}
-                onExecute={this.handleExecute}
-              />
-            )}
-          </SidePanel>
+      <IdentityProvider
```
sohkai: Wondering how you feel about this @bpierre; I think I'd generally recommend putting `<Main />` as the very first component that's rendered.

apps/voting/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
```diff
@@ -0,0 +1,89 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import styled from 'styled-components'
+import { Badge, IdentityBadge, font } from '@aragon/ui'
+import { IdentityContext } from '../IdentityManager/IdentityManager'
+
+function useIdentity(address) {
```
sohkai: Should we move this to its own file so it's easier to share?

This also seems like could be included with the `IdentityManager`.

apps/voting/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
```diff
@@ -0,0 +1,89 @@
...
+
+  const handleNameChange = metadata => {
+    setName(metadata ? metadata.name : null)
+  }
+
+  const handleShowLocalIdentityModal = address => {
+    // Emit an event whenever the modal is closed (when the promise resolves)
+    return showLocalIdentityModal(address)
+      .then(() => updates$.next(address))
+      .catch(e => null)
```
sohkai: Do we need this `catch()` here? Does it make sense to propagate this reject to the caller?

apps/voting/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+const Wrap = styled.div`
+  display: grid;
+  align-items: center;
+  grid-template-columns: auto 1fr;
+  padding-right: 24px;
+`
+
+const Address = styled.span`
```
sohkai: Perhaps this should be named `Label` instead (it would never actually be used if just an address was given)?

apps/voting/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
```diff
@@ -0,0 +1,89 @@
...
+    />
+  )
+}
+
+LocalIdentityBadge.propTypes = {
+  address: PropTypes.string.isRequired,
+}
+
+const Wrap = styled.div`
+  display: grid;
```
sohkai: Seems a _little_ bit unnecessary to use `grid` here 😄

apps/voting/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
AquiGorka: This is the only place where it is used and eventually we'll move this to the `aragonAPI` so I don't think it is necessary to create a new file right now.

apps/voting/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
AquiGorka: The linter will not allow us to push without it :(

sohkai: Aww shoot, we should disable this with the linter then :/

apps/voting/app/src/App.js
bpierre: Yes let’s do this 👍

The only one that really need to be outside is the `<div css="min-width: 320px">`, so it applies to the elements rendered at the root level.

2color: Two questions:
- Is `css` a valid prop? Why is it used over `style`
- Can we add `min-width: 320px` to the Main element?

bpierre: > Is css a valid prop? Why is it used over style

It is being transformed by the styled-components [babel plugin](https://www.styled-components.com/docs/api#css-prop).

It transforms this:

```jsx
import React from 'react'

const App = () => <div css="min-width: 320px" />
```

Into this:

```jsx
import React from 'react'

const App = () => <A />

const A = styled.div`
  min-width: 320px
`
```

And can it even works if you do this:


```jsx
import React from 'react'

const items = ['apple', 'orange', 'banana']

const App = () => (
  <ul>
    {items.map(item => (
      // Only one styled component will be extracted by the babel transform.
      <li css="color: tomato">{item}</li>
    ))}
  </ul>
)
```

It is basically `style`, but without the downsides :-)

> Can we add `min-width: 320px` to the `Main` element?

That’s something I keep wondering… We used to have `AragonApp`, which was defining some default styles, and it was causing too many issues. `Main` replaced it, and it was decided that  it would not set anything styles, and only providing context. We can imagine a website widget using aragonUI, and being smaller than `320px`, so it would be a problem to have it by default.

But on the other hand, developing a fullscreen app with aragonUI is quite a common use case… maybe we could have it as a prop, like `minWidth`, that would be either `0` or `320` by default?

apps/voting/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
AquiGorka: I don't see using `grid` as overkill over `flex` and I tend to use the first one before the latter by default. Let me know if you strongly think we'd need to use `flex` here.

apps/voting/app/src/App.js
2color: Thanks for the clarification.

> Can we add min-width: 320px to the Main element?

On that note, I actually meant to have it inline: `<Main assetsUrl="./aragon-ui" css="min-width: 320px">` rather than a separate element. 

> But on the other hand, developing a fullscreen app with aragonUI is quite a common use case… maybe we could have it as a prop, like minWidth, that would be either 0 or 320 by default?

I don't have an opinion on that.

bpierre: Oh I see, no it wouldn’t directly work because `Main` doesn’t render any DOM element. We could redirect styling props (`style` and `className`) to `Root`, but I would like to not encourage doing this, and prefer the `minWidth` solution for now.

@sohkai @AquiGorka any thoughts on that?

apps/voting/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
2color: Strangely the linter isn't giving me this error. But I do remember seeing it.

apps/voting/app/src/App.js
AquiGorka: `html, body { min-width: 320px}`?

apps/voting/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
sohkai: I'm not as familiar with the `grid` "language" as I am with the older layout ones, so that's partly why I'm not as comfortable using it everywhere.

But in general, my impression is that `grid` is so powerful you could probably build a spaceship with it, but it doesn't mean it's the _best_ or _easiest to maintain_ method of doing it :).

AquiGorka: Consider the verbosity levels here: if I were to use `flex`, I'd have to set the display on the container and then set `flex-grow` in one of the children with `grid` I get the same result with two rules on the container.

PR #729, opened 2019-03-21
Payroll: Organize internal/external functions

facuspagnuolo:
- Specify truffle dependency
- Apply external/internal functions schema (it was almost done)

### DIFF
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -1,11 +1,11 @@
 pragma solidity 0.4.24;
 
+
 import "@aragon/os/contracts/apps/AragonApp.sol";
 import "@aragon/os/contracts/common/EtherTokenConstant.sol";
 import "@aragon/os/contracts/common/IsContract.sol";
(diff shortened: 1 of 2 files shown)


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -369,6 +369,12 @@ contract Payroll is EtherTokenConstant, IsContract, AragonApp { //, IForwarder {
         emit ChangeAddressByEmployee(employeeId, oldAddress, _newAddress);
     }
 
+    function isForwarder() external pure returns (bool) {
```
sohkai: Can we keep the forwarding functions together, like we do in the other apps (e.g. [Token Manager](https://github.com/aragon/aragon-apps/blob/master/apps/token-manager/contracts/TokenManager.sol#L206))? Even if it breaks the flow a little bit, I think the logical grouping is nice :).

facuspagnuolo:
Thanks @sohkai! Will address your suggestions on a separate PR to avoid having conflicts :)

PR #730, opened 2019-03-22
Finance export data as csv

AquiGorka:
These changes implement a modified version of the DateRangeInput components from [here](https://github.com/aragon/aragon-ui/pull/272) and [here](https://github.com/aragon/aragon-apps/blob/master/future-apps/payroll/app/src/components/Input/DateInput.js), the idea is to update this code once the `aragon/ui` PR is merged.

Considerations:
- Sets the default range value to the user's current month.
- DateRange component update the filters as soon as the user changes start or end dates.
- Currently only downloads `csv` files

image


apps/finance/app/src/components/Transfers.js
```diff
@@ -58,6 +89,47 @@ class Transfers extends React.PureComponent {
...
+        )
+      )
+      .join('\n')
+    return window.encodeURI(csvContent)
+  }
+  getCsvFilename = () => {
+    const { selectedDateRange } = this.state
+    const start = format(selectedDateRange.start, 'MMM dd yyyy')
+    const end = format(selectedDateRange.end, 'MMM dd yyyy')
+    return `Finance (${start} to ${end}).csv`
```
2color: For consistency with https://github.com/aragon/aragon/pull/656#discussion_r269028909 

I'd suggest avoiding spaces and using the standard `yyyy-mm-dd` format

apps/finance/app/src/components/Transfers.js
```diff
@@ -58,6 +89,47 @@ class Transfers extends React.PureComponent {
...
+              amount,
+              isIncoming,
+              decimals,
+              true,
+              { rounding: 5 }
+            )
+            return `${format(
+              date,
+              'dd/MM/yy'
+            )},${entity},${reference},${`${formattedAmount} ${symbol}`}`
```
2color: I think it might make sense to put the symbol in a separate column. This will allow for easy aggregation and filtering in most spreadsheet software. 

This is with [tad](https://www.tadviewer.com)
Screen Shot 2019-03-28 at 1 03 17 pm

CC: @dizzypaty @luisivan

AquiGorka:
@bpierre :

image

and when picking dates (reviewed Hopper app and this is how they do it too):

image

AquiGorka:
@2color 

> One thing I noticed is that it's no possible is to set the date range as text in the input. Any thoughts on that? Or does it belong in the aragon-ui date range PR?

This is how the component I "gathered" works currently. We can always iterate if we see we need to input the dates via text.

apps/finance/app/src/components/Transfers.js
```diff
@@ -58,6 +89,47 @@ class Transfers extends React.PureComponent {
...
+        )
+      )
+      .join('\n')
+    return window.encodeURI(csvContent)
+  }
+  getCsvFilename = () => {
+    const { selectedDateRange } = this.state
+    const start = format(selectedDateRange.start, 'yyyy-MM-dd')
+    const end = format(selectedDateRange.end, 'yyyy-MM-dd')
+    return `Finance_(${start}_to_${end}).csv`
```
sohkai: Wondering we should add the proxy address to this file's name so it's unique across organizations cc @luisivan @dizzypaty

apps/finance/app/src/components/Transfers.js
```diff
@@ -58,6 +89,47 @@ class Transfers extends React.PureComponent {
...
+          ({
+            date,
+            numData: { amount },
+            reference,
+            isIncoming,
+            entity,
+            token,
+          }) => {
+            const { symbol, decimals } = tokenDetails[toChecksumAddress(token)]
+            const formattedAmount = formatTokenAmount(
```
sohkai: Wondering if we should use the `BN.js` numbers here instead; `formatTokenAmount()` is **NOT** meant to be used for accounting software due to precision issues.

apps/finance/app/src/components/Transfers.js
bpierre: We could also use the ENS name if available, rather than the proxy address.

sohkai: Ah yes, the org's ENS name will be hard for the Finance app to get at this point though. All we have is the Finance's app's proxy address.

bpierre: Oh true, I forgot we were in an app.

apps/finance/app/src/components/Transfers.js
2color: Created an issue https://github.com/aragon/aragon-apps/issues/756

Please add missing info :)

PR #734, opened 2019-03-25
 Token Manager: disallow minting and vesting tokens to itself

sohkai:
This change adds more checks to disallow usage of the `mint()` and `assignVested()` actions for the Token Manager itself:

- `mint()`: you should use `issue()` instead (separate action protected by a separate role)
- `assignVested()`: Assigning vestings to the Token Manager makes the accounting awkward when calculating the spendable balance of the Token Manager itself. The Token Manager itself never has a spending limit (see `onTransfer()`, so the concept of a vesting does not work very well.

This change also modifies the way `_isBalanceIncreaseAllowed()` works, by always allowing increases to the Token Manager's balance.

Previously, using `assign()` with the Token Manager as the recipient would fail if its balance was already over the limit (see [previously failing test](https://github.com/aragon/aragon-apps/pull/734/files#diff-17c8771ab9c99065348cbba1b3fba658R179)). Although this isn't something to really worry about, it does show this check wasn't fully aligned with the balance assumptions (of it not applying to the Token Manager itself).

### DIFF
--- a/apps/token-manager/contracts/TokenManager.sol
+++ b/apps/token-manager/contracts/TokenManager.sol
@@ -32,14 +32,15 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
     string private constant ERROR_CALLER_NOT_TOKEN = "TM_CALLER_NOT_TOKEN";
     string private constant ERROR_NO_VESTING = "TM_NO_VESTING";
     string private constant ERROR_TOKEN_CONTROLLER = "TM_TOKEN_CONTROLLER";
-    string private constant ERROR_MINT_BALANCE_INCREASE_NOT_ALLOWED = "TM_MINT_BAL_INC_NOT_ALLOWED";
-    string private constant ERROR_ASSIGN_BALANCE_INCREASE_NOT_ALLOWED = "TM_ASSIGN_BAL_INC_NOT_ALLOWED";
+    string private constant ERROR_MINT_RECEIVER_IS_TM = "TM_MINT_RECEIVER_IS_TM";
+    string private constant ERROR_VESTING_TO_TM = "TM_VESTING_TO_TM";
     string private constant ERROR_TOO_MANY_VESTINGS = "TM_TOO_MANY_VESTINGS";
     string private constant ERROR_WRONG_CLIFF_DATE = "TM_WRONG_CLIFF_DATE";
     string private constant ERROR_VESTING_NOT_REVOKABLE = "TM_VESTING_NOT_REVOKABLE";
     string private constant ERROR_REVOKE_TRANSFER_FROM_REVERTED = "TM_REVOKE_TRANSFER_FROM_REVERTED";
-    string private constant ERROR_ASSIGN_TRANSFER_FROM_REVERTED = "TM_ASSIGN_TRANSFER_FROM_REVERTED";
     string private constant ERROR_CAN_NOT_FORWARD = "TM_CAN_NOT_FORWARD";
+    string private constant ERROR_BALANCE_INCREASE_NOT_ALLOWED = "TM_BALANCE_INC_NOT_ALLOWED";
+    string private constant ERROR_ASSIGN_TRANSFER_FROM_REVERTED = "TM_ASSIGN_TRANSFER_FROM_REVERTED";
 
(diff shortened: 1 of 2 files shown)


apps/token-manager/contracts/TokenManager.sol
```diff
@@ -393,6 +393,10 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
         return tokens.sub(vestedTokens);
     }
 
+    function _spendableBalanceOf(address _holder) internal view returns (uint256) {
+        return _transferableBalance(_holder, now);
+    }
+
     function _transferableBalance(address _holder, uint256 _time) internal view returns (uint256) {
```
izqui: Wondering whether this should always return the TM's token balance without iterating over its vestings. Even if there are current instances that have vestings for the TM, we should probably return the full balance as the vesting restrictions won't apply in practice.

If we did this we could also simplify the logic on `onTransfer(...)` as we wouldn't need to check for `_from == address(this)`, but it would perform an extra call to the token that can be avoided.

I would update this function anyway so the `spendableBalanceOf ` getter returns the actual spendable balance of the TM.

apps/token-manager/contracts/TokenManager.sol
```diff
@@ -99,10 +100,11 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
 
     /**
     * @notice Mint `@tokenAmount(self.token(): address, _amount, false)` tokens for `_receiver`
-    * @param _receiver The address receiving the tokens
+    * @param _receiver The address receiving the tokens, cannot be the Token Manager itself (use `issue()` instead)
     * @param _amount Number of tokens minted
     */
     function mint(address _receiver, uint256 _amount) external authP(MINT_ROLE, arr(_receiver, _amount)) {
+        require(_receiver != address(this), ERROR_MINT_RECEIVER_IS_TM);
```
izqui: Totally support doing this. We should update the UI so it takes this change into account and calls the proper function when the TM is inputted in the 'Assign Tokens' panel.

apps/token-manager/contracts/TokenManager.sol
```diff
@@ -301,7 +300,7 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
     }
 
     function spendableBalanceOf(address _holder) public view isInitialized returns (uint256) {
-        return _transferableBalance(_holder, now);
+        return _spendableBalanceOf(_holder);
     }
```
bingen: If this is public (not external) why do we need the wrapper and the internal function here? Can't we just use one as before?

apps/token-manager/contracts/TokenManager.sol
sohkai: See https://github.com/aragon/aragon-apps/issues/748. We could also do a better job of actually showing the Token Manager's balances, if any.

apps/token-manager/contracts/TokenManager.sol
sohkai: This is used in `onTransfer()` as well, so I split it up into the public / internal functions.

apps/token-manager/contracts/TokenManager.sol
sohkai: > If we did this we could also simplify the logic on onTransfer(...) as we wouldn't need to check for _from == address(this), but it would perform an extra call to the token that can be avoided.

Even though this is a little bit more expensive for doing `assign()` and `assignVested()`, I think this makes sense as it cheapens other, arbitrary token transfers.

It also makes the branching simpler, which is always good 😄

apps/token-manager/contracts/TokenManager.sol
bingen: So the goal is to save checking `isInitialized`? Couldn't we just do `_transferableBalance(_holder, now)` from there? It seems like unnecessary complexity to me.

sohkai: It is a little bit more expensive with the extra jump, but I was thinking it'd be a lot more clear to reuse the functionality.

However, since this is getting used on every token transfer, it's probably worth it to be as optimized as possible 👍.

PR #735, opened 2019-03-25
Finance: rename RecurringPayments to ScheduledPayments

sohkai:
Builds on #711, but should be merged separately.

Now that the function name is called `newScheduledPayment()`, I thought it might make sense to rename the struct to also be `ScheduledPayment`.

### DIFF
--- a/apps/finance/contracts/Finance.sol
+++ b/apps/finance/contracts/Finance.sol
@@ -27,14 +27,14 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
     bytes32 public constant EXECUTE_PAYMENTS_ROLE = keccak256("EXECUTE_PAYMENTS_ROLE");
     bytes32 public constant MANAGE_PAYMENTS_ROLE = keccak256("MANAGE_PAYMENTS_ROLE");
 
-    uint256 internal constant NO_RECURRING_PAYMENT = 0;
+    uint256 internal constant NO_SCHEDULED_PAYMENT = 0;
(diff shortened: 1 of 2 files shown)


apps/finance/contracts/Finance.sol
```diff
@@ -211,6 +211,34 @@ contract Finance is EtherTokenConstant, IsContract, AragonApp {
...
+    *      to set a baseline requirement for the interval time), it does mean users can't grant a
+    *      permission that has a upperbound requirement for the interval time.
+    * @param _token Address of token for payment
+    * @param _receiver Address that will receive payment
+    * @param _amount Tokens that are paid every time the payment is due
+    * @param _reference String detailing payment reason
+    */
+    function newImmediatePayment(address _token, address _receiver, uint256 _amount, string _reference)
+        external
+        authP(CREATE_PAYMENTS_ROLE, _arr(_token, _receiver, _amount, uint256(getTimestamp64()), uint256(MAX_UINT64), uint256(1)))
```
bingen: We can use [`getTimestamp`](https://github.com/aragon/aragonOS/blob/dev/contracts/common/TimeHelpers.sol#L36) so we don't have to cast, right?

sohkai: Fixed in https://github.com/aragon/aragon-apps/pull/711.

bingen: Cool, just one little comment, I see this one builds on top of #711 , so let's make sure it's not overwritten when merging.

PR #736, opened 2019-03-26
Payroll: Avoid payout overflow

facuspagnuolo:
This PR revives what @bingen has been working on in order to avoid overflows while transferring the payroll + the accrued value ([see here](https://github.com/protofire/aragon-apps/pull/113)). However,  I'm proposing another alternative in order to deal with this edge case. The idea is to split the interface allowing the employees to request their salary or accrued value separately. 

Additionally, I used the verb `reimburse` to refer to the action of paying the accrued value to an employee, please let me know if there is a more suitable way to refer to it (cc @izqui @luisivan).

### DIFF
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -517,15 +517,7 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
     }
 
     function _addAccruedValue(uint256 _employeeId, uint256 _amount) internal {
-        Employee storage employee = employees[_employeeId];
-
-        // if the result would overflow, set it to max int
-        uint256 result = employee.accruedValue +_amount;
-        if (result >= _amount) {
-            employee.accruedValue = result;
-        } else {
-            employee.accruedValue = MAX_UINT256;
-        }
+        employees[_employeeId].accruedValue = employees[_employeeId].accruedValue.add(_amount);
 
         emit AddEmployeeAccruedValue(_employeeId, _amount);
     }
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -32,6 +32,7 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
     bytes32 constant public MODIFY_RATE_EXPIRY_ROLE = keccak256("MODIFY_RATE_EXPIRY_ROLE");
 
     uint128 internal constant ONE = 10 ** 18; // 10^18 is considered 1 in the price feed to allow for decimal calculations
+    uint256 internal constant MAX_UINT256 = uint256(-1);
     uint64 internal constant MAX_UINT64 = uint64(-1);
     uint8 internal constant MAX_ALLOWED_TOKENS = 20; // for loop in `payday()` uses ~260k gas per available token
     uint256 internal constant MAX_ACCRUED_VALUE = 2**128;
@@ -296,8 +297,6 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
         employeeActive(_employeeId)
         authP(ADD_ACCRUED_VALUE_ROLE, arr(_employeeId, _amount))
     {
-        require(_amount <= MAX_ACCRUED_VALUE, ERROR_ACCRUED_VALUE_TOO_BIG);
-
         _addAccruedValue(_employeeId, _amount);
     }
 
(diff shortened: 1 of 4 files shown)


shared/test-helpers/assertThrow.js
```diff
@@ -1,26 +1,28 @@
...
-  async assertInvalidOpcode(block, message = 'should have failed with invalid opcode') {
-    return assertThrows(block, message, 'invalid opcode')
+  async assertInvalidOpcode(blockOrPromise, message = '') {
+    return assertThrows(blockOrPromise, 'invalid opcode', message)
   },
 
-  async assertRevert(block, message = 'should have failed by reverting') {
-    return assertThrows(block, message, 'revert')
+  async assertRevert(blockOrPromise, message = '') {
+    return assertThrows(blockOrPromise, 'revert', message)
```
facuspagnuolo: I extended the functionality of this helper to support promises and error messages as well, making sure it is backwards compatible to what we have currently. It basically simplifies the way it can be used (see [here](https://github.com/aragon/aragon-apps/pull/736/commits/f2b94390ac954cf8ce94e0d93f529113fe5971fa#diff-60f24b1826110bb61491ab38fa09ecdcR206)). 

If you guys like it, I can take care and propagate it through the rest of the codebase. If you think this should be part of a separate PR, no problem at all as well :)

sohkai: Would be amazing to do so!

We should also add the ability to specify a specific revert message, but I can do that as part of the changes to Finance.

sohkai: See https://github.com/aragon/aragon-apps/pull/741/commits/ac7eb6a67b18792a06c5babfcddf96862780b505.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -634,4 +668,48 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+                emit SendPayment(employee.accountAddress, token, tokenAmount, _reference);
+                somethingPaid = true;
+            }
+        }
+    }
+
+    /**
+     * @dev Try removing employee if there are no pending payments and has reached employee's end date.
+     * @param _employeeId Employee's identifier
+     * @param _date Date timestamp used to evaluate if the employee can be removed from the payroll.
```
bingen: I don't think we should pass this in as a parameter, but always use `getTimestamp64()`.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -634,4 +668,48 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+     */
+    function _tryRemovingEmployee(uint256 _employeeId, uint64 _date) private {
+        Employee storage employee = employees[_employeeId];
+
+        bool hasReachedEndDate = employee.endDate <= _date;
+        bool isOwedSalary = _getOwedSalary(_employeeId, _date) > 0;
```
bingen: Wouldn't it be more straightforward to do something like: 
```
bool noPendingSalary = _getOwedSalary(_employeeId, _date) == 0;
bool noPendingAccruedValue = employees[_employeeId].accruedValue == 0;

if (hasReachedEndDate && noPendingSalary && noPendingAccruedValue) {
(...)
```
?

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: I see your point but I think it makes sense to use the min date between current timestamp and employee end-date, because we would get an invalid owed salary amount otherwise. That's why I parameterized it and make it `private`. 

I was also thinking that maybe we could move the `min(timestamp, enddate)` logic within the owed salary amount calc since we will need to use it always as is I think.

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: Sure 😅, I'll address this comment in this [polishing PR](https://github.com/aragon/aragon-apps/pull/744)

future-apps/payroll/contracts/Payroll.sol
bingen: Good point, but then I would do something like:
```
if (employee.endDate > getTimestamp64()) {
    return;
}
if (_getOwedSalary(_employeeId, employee.endDate) > 0) {
   return;
}
if (employee.accruedValue > 0) {
   return;
}

delete employeeIds[employee.accountAddress];
delete employees[_employeeId];
```
This way we are also saving some gas in most cases (where employee didn't reach the end date, which should be the regular case)

future-apps/payroll/contracts/Payroll.sol
bingen: Nice. Another thing I missed: you can use the pointer `employee` in a couple of places where you are using `employees[_employeeId]`.

facuspagnuolo: Yea! I already did that in #744 :)

bingen: Oh, sorry, as it was "WIP", I didn't check it out yet.

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: I finally solve it like [this](https://github.com/aragon/aragon-apps/pull/744/commits/8a86879ea0f9bccea8a3f682c41b9d09ba769726). As you can see I decided to move the `min(current, end date)` thing within `_getOwedSalary` since it was actually misleading other calculations for partial payrolls.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -334,14 +332,49 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
-        bool somethingPaid = _payTokens(employeeIds[msg.sender]);
+        bool somethingPaid = _payday(employeeIds[msg.sender], 0);
+        require(somethingPaid, ERROR_NOTHING_PAID);
+    }
+
+    /**
+     * @dev Withdraw accrued value by employee (the caller). The specified amount capped to the one owed will be transferred.
+     *      Initialization check is implicitly provided by `employeeMatches()` as new employees can
+     *      only be added via `addEmployee(),` which requires initialization.
+     * @notice Withdraw your own payroll.
```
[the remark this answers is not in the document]
facuspagnuolo: Addressed in #744

future-apps/payroll/contracts/Payroll.sol
bingen: Sounds good!

PR #737, opened 2019-03-26
Voting: mock time in tests rather than use timetravel

sohkai:
Must've forgot to do this back when we migrated to aragonOS@4 😅. `timeTravel()` isn't supported outside of Ganache, so we generally use the built-in time mocks in `AragonApp`.

### DIFF
--- a/apps/voting/contracts/test/mocks/VotingMock.sol
+++ b/apps/voting/contracts/test/mocks/VotingMock.sol
@@ -4,7 +4,7 @@ import "../../Voting.sol";
 
 
 contract VotingMock is Voting {
-    uint64 mockTime = uint64(now);
+    uint64 mockTime;
 
     function mock_setTimestamp(uint64 i) public { mockTime = i; }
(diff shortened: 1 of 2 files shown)


apps/voting/contracts/test/mocks/VotingMock.sol
```diff
@@ -4,6 +4,11 @@ import "../../Voting.sol";
 
 
 contract VotingMock is Voting {
+    uint64 mockTime = uint64(now);
```
izqui: When using this contract behind a proxy, this value will not be set and mockTime will be 0 until mock_setTimestamp is used.

sohkai: Turns out the tests are calling `VotingMock.mock_setTimestamp()` on creation of the proxy so this wasn't a problem, but I can't figure out a better way of doing this outside of making it good practice to set this immediately on mocks.

I was thinking about adding an `initialize()` wrapper that would set this, but it doesn't feel right.

PR #738, opened 2019-03-26
Token Manager: use TimeHelpers and mock time in tests

sohkai:
Builds on #734, but should be merged separately.

Similar to #737, we should use the built-in `TimeHelpers` and time mocking instead of relying on `now` and `timeTravel()`.

### DIFF
--- a/apps/token-manager/contracts/TokenManager.sol
+++ b/apps/token-manager/contracts/TokenManager.sol
@@ -190,7 +190,7 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
 
         uint256 nonVested = _calculateNonVestedTokens(
             v.amount,
-            getTimestamp64(),
+            getTimestamp(),
             v.start,
             v.cliff,
             v.vesting
@@ -249,7 +249,7 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
     * @return False if the controller does not authorize the transfer
     */
     function onTransfer(address _from, address _to, uint256 _amount) public onlyToken returns (bool) {
-        return _isBalanceIncreaseAllowed(_to, _amount) && _transferableBalance(_from, getTimestamp64()) >= _amount;
+        return _isBalanceIncreaseAllowed(_to, _amount) && _transferableBalance(_from, getTimestamp()) >= _amount;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


apps/token-manager/contracts/TokenManager.sol
```diff
@@ -303,7 +302,7 @@ contract TokenManager is ITokenController, IForwarder, AragonApp {
         return _spendableBalanceOf(_holder);
     }
 
-    function transferableBalance(address _holder, uint256 _time) public view isInitialized returns (uint256) {
+    function transferableBalance(address _holder, uint64 _time) public view isInitialized returns (uint256) {
```
izqui: I don't think there is a strong enough reason to break the interface (this function signature changes) of the TokenManager for this change

apps/token-manager/contracts/test/mocks/TokenManagerMock.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity 0.4.24;
+
+import "../../TokenManager.sol";
+
+
+contract TokenManagerMock is TokenManager {
+    uint64 mockTime = uint64(now);
```
izqui: When using this contract behind a proxy, this value will not be set and `mockTime` will be 0 until `mock_setTimestamp` is used

apps/token-manager/test/tokenmanager.js
```diff
@@ -384,25 +383,27 @@ contract('Token Manager', accounts => {
                 assert.equal(vRevokable, revokable)
             })
 
-            it('can start transfering on cliff', async () => {
-                await timetravel(cliff)
+            it('can start transferring on cliff', async () => {
```
izqui: The `beforeEach()` could be modified as it is setting dates depending on the actual timestamp of the block, which is not used when mocking time.

I would also set an initial time right after initializing the contract so it is not 0.

apps/token-manager/contracts/TokenManager.sol
sohkai: Ahh shoot, that's true :(. Will revert back to using `Uint256Helpers` for this.

sohkai: Ended up just leaving current time as uint256, as we never use it to store a value and only use it comparisons (where leaving it as uint256 is cheaper).

sohkai:
@bingen Ready for re-review!

PR #739, opened 2019-03-26
Finance: rename maxRepeats to maxExecutions

sohkai:
Builds on #711, but should be merged separately.

Renames `maxRepeats` to `maxExecutions` to better reflect the parameter and state.

A `maxRepeats` of 1 is quite confusing as to whether it means you can repeat the payment once (e.g. actually execute it twice), or only execute it once.


bingen:
> A `maxRepeats` of 1 is quite confusing as to whether it means you can repeat the payment once (e.g. actually execute it twice), or only execute it once.

So true!

sohkai:
> Ths looks good to me, but shouldn't repeats and paymentRepeatNumber be renamed to executions and paymentExecutionNumber too, for consistency?

Yes, good catch! Done in 7c70a0e

PR #741, opened 2019-03-26
Finance: test revert strings

sohkai:
Builds on #739, but should be merged separately.

Cherry-picks the change by @facuspagnuolo in https://github.com/aragon/aragon-apps/pull/736/commits/6964450e3d56159fd8049d6a3c7a4a4bf2de79a1 (to place it in `next`) and builds off of it.

Found a few previously broken tests with the revert strings 😄.

### DIFF
--- a/apps/finance/test/finance.js
+++ b/apps/finance/test/finance.js
@@ -1,5 +1,5 @@
 const assertEvent = require('@aragon/test-helpers/assertEvent')
-const { assertRevert, assertInvalidOpcode } = require('@aragon/test-helpers/assertThrow')
+const { assertRevert } = require('@aragon/test-helpers/assertThrow')
 const getBalance = require('@aragon/test-helpers/balance')(web3)
 
 const Finance = artifacts.require('FinanceMock')
--- a/apps/finance/test/finance.js
+++ b/apps/finance/test/finance.js
@@ -1,6 +1,7 @@
 const assertEvent = require('@aragon/test-helpers/assertEvent')
 const { assertRevert } = require('@aragon/test-helpers/assertThrow')
 const getBalance = require('@aragon/test-helpers/balance')(web3)
+const { makeErrorMappingProxy } = require('@aragon/test-helpers/utils')
 
 const Finance = artifacts.require('FinanceMock')
 const Vault = artifacts.require('Vault')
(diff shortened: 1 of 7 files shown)


shared/test-helpers/assertThrow.js
```diff
@@ -1,26 +1,37 @@
...
-    await block()
-  } catch (e) {
-    return assertError(e, errorCode, message)
+    (typeof blockOrPromise === 'function') ? await blockOrPromise() : await blockOrPromise
+  } catch (error) {
+    assertError(error, expectedErrorCode)
+    return error
   }
-  assert.fail('should have thrown before')
+  assert.fail(`Expected "${expectedErrorCode}" but it did not fail`)
```
izqui: It could be useful to have the expected revert reason as part of the failure message if it didn't revert

apps/finance/test/finance.js
```diff
@@ -798,37 +790,55 @@ contract('Finance App', accounts => {
...
-                        await finance.executePayment(paymentId, { from: recipient })
-                    })
+                    await assertRevert(finance.executePayment(paymentId), errors.FINANCE_PAYMENT_INACTIVE)
                 })
 
                 it('succeeds payment after setting payment status to active', async () => {
                     await finance.setPaymentStatus(paymentId, true)
                     await finance.mock_setTimestamp(time + 1)
 
-                    await finance.executePayment(paymentId, { from: recipient })
```
izqui: Can't really figure out how this test was working before if `recipient` didn't have the `EXECUTE_PAYMENT_ROLE`

apps/finance/test/finance.js
```diff
@@ -39,7 +40,41 @@ contract('Finance App', accounts => {
     let CREATE_PAYMENTS_ROLE, CHANGE_PERIOD_ROLE, CHANGE_BUDGETS_ROLE, EXECUTE_PAYMENTS_ROLE, MANAGE_PAYMENTS_ROLE
     let TRANSFER_ROLE
 
+    // Error strings
+    const errors = makeErrorMappingProxy({
```
izqui: Is the reason for creating the proxy just throwing if an error hasn't been defined?

apps/finance/test/finance.js
sohkai: I was really confused by this too, which is why I set up the roles more strictly now. Previously all the roles were set to allow anyone to use them.

apps/finance/test/finance.js
sohkai: Yup, was afraid it'd return `undefined` if you mis-typed and then the revert test wouldn't actually be using a reason.

PR #743, opened 2019-03-26
Abstract identity logic with a custom useIdentity react hook

2color:
## What
- Define a hook which abstracts resolution and `updates$` subscription logic
- Merge the two contexts into what. 
  - I feel somewhat conflicted about this. But I can't think of a situation where you only want the one. So perhaps we do want a gorilla whenever we want a banana... 

## Why

This partially addresses https://github.com/aragon/aragon/issues/661 and https://github.com/aragon/aragon/issues/660. Though it would probably look different in the aragon client implementation.


apps/voting/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
```diff
@@ -2,40 +2,33 @@ import React from 'react'
...
-      // address does not ressolve to identity
-    }
-  }
-  const handleClick = () => {
-    showLocalIdentityModal(address)
-      .then(handleResolve)
-      .catch(e => {
-        /* user cancelled modify intent */
-      })
+function useIdentity(address) {
```
AquiGorka: Since this is the only place it is being used it makes sense to define it here, eventually we could add this to the new [`aragonApi`](https://github.com/aragon/aragon.js/pull/259).

apps/voting/app/src/components/IdentityManager/IdentityManager.js
```diff
@@ -1,14 +1,33 @@
...
 
 const IdentityContext = React.createContext({
   resolve: () =>
     Promise.reject(Error('Please set resolve using IdentityProvider')),
 })
 
-const IdentityProvider = ({ onResolve, children }) => {
+const IdentityProvider = ({
+  onResolve,
+  onShowLocalIdentityModal,
```
AquiGorka: I think we are missing the `PropType` declaration for this prop.

PR #744, opened 2019-03-26
Payroll: Polish payroll app

facuspagnuolo:
**MERGE AFTER https://github.com/aragon/aragon-apps/pull/736**

Given that I've polishing and tweaking many parts of the payroll app, I tried to work submitting atomic commits to help reviewing my progress. The highlights of this PR are:
- Fix last payroll timestamp calculation for partial paydays (see https://github.com/aragon/aragon-apps/pull/744/commits/8a86879ea0f9bccea8a3f682c41b9d09ba769726)
- Polish contract interface and cosmetic enhancements
- Refactor and tune up test suite
- Add missing inline documentation and review wording
- Clean up unused mock contracts
- Organize test helper functions

 ___

### Interface changes

- The previous `addEmployee` function without a `startDate` was renamed to `addEmployeeNow`. 
- Removed `string name` argument from both `addEmployee` and `addEmployeeNow` functions.
- Function `determineAllocation` now receives an `uint256` array instead of an `uint8` one.
- Function `getAllocation` now returns an `uint256` instead of `uint8`
- Added new `partialPayday(uint256 requestedAmount)`, `reimburse()` and `partialReimburse(uint256 requestedAmount)` functions.
- Added `string reference` argument to `SendPayment` event.
- Removed `string name` argument from `AddEmployee` event.


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -686,17 +686,16 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+        bool hasReachedEndDate = employee.endDate <= getTimestamp64();
+        bool noPendingSalary = _getOwedSalary(_employeeId) == 0;
+        bool noPendingAccruedValue = employee.accruedValue == 0;
+        bool noPendingPayments = noPendingSalary && noPendingAccruedValue;
 
-        if (hasReachedEndDate && areNoPendingPayments) {
+        if (hasReachedEndDate && noPendingPayments) {
             delete employeeIds[employee.accountAddress];
             delete employees[_employeeId];
         }
```
bingen: I would still suggest using shortcuts, as explained [here](https://github.com/aragon/aragon-apps/pull/736#discussion_r269501064) to save some gas in the majority of the uses of this function.
```
if (employee.endDate > getTimestamp64()) {
    return;
}
if (_getOwedSalary(_employeeId) > 0) {
   return;
}
if (employee.accruedValue > 0) {
   return;
}

delete employeeIds[employee.accountAddress];
delete employees[_employeeId];
```
Don't you think?

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -612,25 +604,33 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
         emit TerminateEmployee(_employeeId, employee.accountAddress, _endDate);
     }
 
-    function _getOwedSalary(uint256 _employeeId, uint64 _date) internal view returns (uint256) {
+    function _getLastPayroll(uint256 _employeeId, uint256 _payedAmount) internal view returns (uint64) {
+        Employee storage employee = employees[_employeeId];
+        uint64 timeDiff = uint64(_payedAmount.div(employee.denominationTokenSalary));
+        uint date = uint(employee.lastPayroll.add(timeDiff));
```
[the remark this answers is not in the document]
facuspagnuolo: True, it was an old logging line 😛

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: Completely, I still think that boolean algebra was one of the easiest subjects I learned and one of those that I underestimated most at the same time 😂

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -612,25 +604,33 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
         emit TerminateEmployee(_employeeId, employee.accountAddress, _endDate);
     }
 
-    function _getOwedSalary(uint256 _employeeId, uint64 _date) internal view returns (uint256) {
+    function _getLastPayroll(uint256 _employeeId, uint256 _payedAmount) internal view returns (uint64) {
+        Employee storage employee = employees[_employeeId];
+        uint64 timeDiff = uint64(_payedAmount.div(employee.denominationTokenSalary));
```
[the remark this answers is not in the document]
bingen: Sorry, I meant `Uint256Helpers.toUint64`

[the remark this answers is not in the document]
facuspagnuolo: Discussed offline, we were referring to use `Uint256Helpers` actually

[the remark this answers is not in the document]
bingen: Going over it again, I see this is only called from `_payday`, so `_payedAmount` will always be <= `owedAmount`, which comes from `_getOwedSalary`, which does the opposite calculation but with a date `<= getTimestamp64()`, so this can not ever be greater than the  current date, which is fine. Anyway, as a sanity check, I would still use that helper, or, even better, check than the result is not greater than `getTimestamp64()`, as it doesn't seem to make sense otherwise.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -56,27 +58,33 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
     }
 
+    IFeed public feed;
     Finance public finance;
     address public denominationToken;
-    IFeed public feed;
     uint64 public rateExpiryTime;
 
-    mapping(uint256 => Employee) private employees;     // employee ID -> employee
-    // Employees start at index 1, to allow us to use employees[0] to check for non-existent address
```
bingen: This is still true, right? I would keep the comment then.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -9,19 +9,20 @@ import "@aragon/os/contracts/common/IForwarder.sol";
...
 import "@aragon/apps-finance/contracts/Finance.sol";
 
 
 /**
  * @title Payroll in multiple currencies
  */
 contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
-    using SafeMath for uint256;
-    using SafeMath64 for uint64;
     using SafeMath8 for uint8;
```
bingen: This is no longer needed, right?

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -435,55 +456,39 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+     * @param _name Employee's name.
+     * @param _role Employee's role.
+     * @param _startDate Employee's starting timestamp in seconds.
+     */
+    function _addEmployee(address _accountAddress, uint256 _initialDenominationSalary, string _name, string _role, uint64 _startDate) internal {
         // Check address isn't already being used
-        require(employeeIds[_accountAddress] == 0, ERROR_EMPLOYEE_ALREADY_EXIST);
+        require(!_employeeExists(_accountAddress), ERROR_EMPLOYEE_ALREADY_EXIST);
 
+        // Employees start at index 1, to allow us to use employees[0] to check for non-existent address
```
bingen: Oh, I see it here now, hehe. Anyway, I would keep it at the beginning, where it was, or in `initialize` function, where `nextEmployee` is set to 1.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -594,21 +598,44 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
-    function _getOwedSalary(uint256 _employeeId, uint64 _date) internal view returns (uint256) {
+    /**
+     * @dev Calculates the date timestamp corresponding to the requested paying amount based on the employee's last payroll date.
+     * @param _employeeId Employee's identifier.
+     * @param _payedAmount Amount payed to the employee to query the timestamp of.
+     * @return The date timestamp in seconds corresponding to the given paying amount based on the employee's last payroll date.
+     */
+    function _getLastPayroll(uint256 _employeeId, uint256 _payedAmount) internal view returns (uint64) {
+        Employee storage employee = employees[_employeeId];
+        uint64 timeDiff = _payedAmount.div(employee.denominationTokenSalary).toUint64();
```
bingen: As I said [here](https://github.com/aragon/aragon-apps/pull/744#discussion_r270079562), I think than better than this check, I would make sure that is not greater than `getTimestamp64()`. I guess that we could then remove the `Uint256Helpers` dependency.

future-apps/payroll/contracts/Payroll.sol
bingen: Hahaha, so true!

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: I think it makes much more sense to have it near to the LOC where we do it :)
I'll move it to the initialize function

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: Answered [below](https://github.com/aragon/aragon-apps/pull/744#discussion_r270369503)

facuspagnuolo:
Thanks for reviewing @bingen! Addressed all your comments if you'd like to look at my last commit. Regarding the test files unification, my rational was just to follow the rest of the apps where there is one single file per contract. However, it make sense splitting them into separate files, it actually helps a lot when running tests in parallel. Please let me know guys if you prefer having them split, I can roll it back easily.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -606,8 +603,18 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
-        uint64 timeDiff = _payedAmount.div(employee.denominationTokenSalary).toUint64();
-        return employee.lastPayroll.add(timeDiff);
+
+        uint64 timeDiff = uint64(_payedAmount.div(employee.denominationTokenSalary));
+
+        // This function is only called from _payday, where we make sure that _payedAmount is lower or equal to the
+        // total owed amount, that is obtained from _getOwedSalary, which does exactly the opposite calculation:
+        // multiplying the employee's salary by an uint64 number of seconds. Therefore, timeDiff will always fit in 64.
+        // Nevertheless, we are performing a sanity check at the end to ensure the computed last payroll timestamp
+        // is not greater than the current timestamp.
```
bingen: Really nice that you added this explanation here!

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -606,8 +603,18 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+
+        uint64 lastPayrollDate = employee.lastPayroll.add(timeDiff);
```
bingen: Nit pick: wouldn't it be better to do all the math on `uint256`, do the check, and then finally cast it to `uint64` at the end? As we have discussed, there's no way this can overflow as it is right now, but just in case, to be safer.

sohkai:
@facuspagnuolo On the subject of tests, I think it makes sense to split them up. As we get better about testing the other apps, we'll likely want to split their test files too :).

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -594,21 +595,54 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
-        // No need to use safe math as the underflow was covered by the previous check
-        uint64 time = _date - employee.lastPayroll;
+        // Get time diff in seconds, no need to use safe math as the underflow was covered by the previous check
+        uint64 timeDiff = date - employee.lastPayroll;
+        uint256 result = employee.denominationTokenSalary * uint256(timeDiff);
 
-        return employee.denominationTokenSalary.mul(time);
+        // Return max int if the result overflows
+        if (result / timeDiff != employee.denominationTokenSalary) {
+            return MAX_UINT256;
```
sohkai: I don't know if this would work that well in practice; if you use `_addAccruedValue()` afterwards (like we do in `setEmployeeSalary()`), you'd still revert there if there was any existing accrued value.

I think more consistent behaviour would be to just use safe math, or possibly even revert with better error strings for overflowing salary or accrued value.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -137,192 +139,145 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
-        employeeActive(_employeeId)
         authP(SET_EMPLOYEE_SALARY_ROLE, arr(_employeeId, _denominationSalary))
+        employeeActive(_employeeId)
     {
-        uint64 timestamp = getTimestamp64();
-
         // Add owed salary to employee's accrued value
-        uint256 owed = _getOwedSalary(_employeeId, timestamp);
+        uint256 owed = _getOwedSalary(_employeeId);
         _addAccruedValue(_employeeId, owed);
```
sohkai: ⚠️ This is the only bit that's not nice with changing accrued value to "reimbursements". Changing your salary should not make your previous accrued salary a reimbursement.

A way around this would be to add another storage slot, `employee.accruedSalary`, and keep tracked of accrued salary there instead of adjusting the last payroll date on partial payrolls.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -634,4 +667,79 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
 
         return xrt;
     }
+
+    /**
+     * @dev Loops over tokens to send requested amount to the employee
+     * @param _employeeId Employee's identifier
+     * @param _totalAmount Total amount to be transferred to the employee distributed through the setup tokens allocation.
```
```suggestion
     * @param _totalAmount Total amount to be transferred to the employee distributed in accordance to the employee's token allocation.
```

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -634,4 +667,79 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+     * @param _reference String detailing payment reason.
+     * @return True if there was at least one token transfer.
+     */
+    function _transferTokensAmount(uint256 _employeeId, uint256 _totalAmount, string _reference) private returns (bool somethingPaid) {
```
sohkai: Wondering if we should leave this as `internal`? Usually we leave everything as `internal` in case they're useful in subclasses.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -634,4 +667,79 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+            address token = allowedTokensArray[i];
+            if (employee.allocation[token] != uint256(0)) {
+                uint128 exchangeRate = _getExchangeRate(token);
+                require(exchangeRate > 0, ERROR_EXCHANGE_RATE_ZERO);
+                // Salary converted to token and applied allocation percentage
+                uint256 tokenAmount = _totalAmount.mul(exchangeRate).mul(employee.allocation[token]);
+                // Divide by 100 for the allocation and by ONE for the exchange rate
+                tokenAmount = tokenAmount / (100 * ONE);
+                finance.newPayment(token, employee.accountAddress, tokenAmount, 0, 0, 1, _reference);
+                emit SendPayment(employee.accountAddress, token, tokenAmount, _reference);
```
sohkai: Some more opportunities for a little bit more aggressive caching to avoid SLOADs.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -634,4 +667,79 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+                somethingPaid = true;
+            }
+        }
+    }
```
sohkai: ⚠️ We should add a reentrancy guard to this function and `Payroll.determineAllocation()` (or perhaps all the functions with the `employeeMatches` modifier).

Right now, with ETH, you can re-enter into `Payroll.determineAllocation()` and modify your allocation so that a later token becomes 100% of your allocation. The worse case of this is if you set ETH to 100%, and then upon paying ETH, your contract re-enters into `Payroll.determineAllocation()` and sets a later token to be 100%.

future-apps/payroll/contracts/test/mocks/PayrollMock.sol
```diff
@@ -6,33 +6,12 @@ import "../../Payroll.sol";
...
-            _initialDenominationSalary,
-            _name,
-            _role,
-            getTimestamp64()
-        );
-    }
-
     function mockUpdateTimestamp() public { _mockTime = uint64(now); }
     function mockSetTimestamp(uint64 i) public { _mockTime = i; }
     function mockAddTimestamp(uint64 i) public { _mockTime += i; require(_mockTime >= i); }
```
sohkai: In the other apps, we've used `mock_camelCase()` as the function names for the mock functions

future-apps/payroll/test/contracts/Payroll.test.js
```diff
@@ -0,0 +1,3362 @@
...
+const MAX_GAS_USED = 6.5e6
+const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000'
+
+contract('Payroll, initialization,', ([owner, employee, anotherEmployee, anyone]) => {
+  let dao, acl, payroll, payrollBase, finance, vault, priceFeed, denominationToken, anotherToken
+
+  const NOW = 1553703809 // random fixed timestamp in seconds
+  const ONE_MONTH = 60 * 60 * 24 * 31
+  const TWO_MONTHS = ONE_MONTH * 2
+  const RATE_EXPIRATION_TIME = TWO_MONTHS
```
sohkai: This is a bit unrealistic, since it's so long. Perhaps we could use something like thirty minutes instead?

future-apps/payroll/test/contracts/Payroll.test.js
```diff
@@ -0,0 +1,3362 @@
...
+    const contracts = await getDaoFinanceVault(owner)
+    dao = contracts.dao
+    finance = contracts.finance
+    vault = contracts.vault
+    acl = ACL.at(await dao.acl())
+
+    priceFeed = await PriceFeed.new()
+    payrollBase = await Payroll.new()
+
+    anotherToken = await deployErc20TokenAndDeposit(owner, finance, vault, 'Another token', 18)
```
sohkai:
```suggestion
    anotherToken = await deployErc20TokenAndDeposit(owner, finance, vault, 'Another token', DENOMINATION_TOKEN_DECIMALS)
```

future-apps/payroll/test/contracts/Payroll.test.js
```diff
@@ -0,0 +1,3362 @@
...
+  })
+
+  beforeEach('grant permissions', async () => {
+    const ADD_EMPLOYEE_ROLE = await payrollBase.ADD_EMPLOYEE_ROLE()
+    const ADD_ACCRUED_VALUE_ROLE = await payrollBase.ADD_ACCRUED_VALUE_ROLE()
+    const CHANGE_PRICE_FEED_ROLE = await payrollBase.CHANGE_PRICE_FEED_ROLE()
+    const MODIFY_RATE_EXPIRY_ROLE = await payrollBase.MODIFY_RATE_EXPIRY_ROLE()
+    const TERMINATE_EMPLOYEE_ROLE = await payrollBase.TERMINATE_EMPLOYEE_ROLE()
+    const SET_EMPLOYEE_SALARY_ROLE = await payrollBase.SET_EMPLOYEE_SALARY_ROLE()
+    const ALLOWED_TOKENS_MANAGER_ROLE = await payrollBase.ALLOWED_TOKENS_MANAGER_ROLE()
```
sohkai: Probably a super small thing, but since this gets called for every test and should be constant, could we cache their values in `before()`?

future-apps/payroll/test/helpers/setup.js
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  async function redistributeEth(finance) {
+    const split = 4
+    const amount = 10
+    const accounts = web3.eth.accounts
+
+    // transfer ETH to owner
+    for (let i = 1; i < split; i++)
+      await web3.eth.sendTransaction({ from: accounts[i], to: accounts[0], value: web3.toWei(amount, 'ether') })
```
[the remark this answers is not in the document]
bingen: Long time ago, it was to then fund the payroll app from that account, I think, and then have enough funds for all payments. But I guess this can be avoided, for instance lowering salaries in some orders of magnitude if needed.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -526,64 +534,57 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
-        } else {
-            toDate = timestamp;
-        }
-
-        // Compute owed amount
-        uint256 owed = employee.accruedValue.add(_getOwedSalary(_employeeId, toDate));
-        if (owed == 0) {
+        // Compute amount to be payed
+        uint256 owedAmount = _getOwedSalary(_employeeId);
+        if (owedAmount == 0 || owedAmount < _requestedAmount) {
```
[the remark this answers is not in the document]
bingen: Well, not really, if `owedAmount == 0` and `_requestedAmount == 0`, `owedAmount < _requestedAmount` will be false, while the original condition would be true.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -634,4 +667,79 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+                somethingPaid = true;
+            }
+        }
+    }
+
+    /**
+     * @dev Tries removing employee if there are no pending payments and has reached employee's end date.
+     * @param _employeeId Employee's identifier
+     */
+    function _tryRemovingEmployee(uint256 _employeeId) private {
```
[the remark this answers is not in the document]
bingen: I think we discussed this already, but of course I don't remember, haha. I think the reason was precisely because we don't have the "re-enabling" functionality and the address would be blocked. This way an employee can  be re-enabled by creating a new one and using the same address. Does it make sense?

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -594,21 +595,54 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+        uint256 timeDiff = _payedAmount.div(employee.denominationTokenSalary);
+
+        // This function is only called from _payday, where we make sure that _payedAmount is lower or equal to the
+        // total owed amount, that is obtained from _getOwedSalary, which does exactly the opposite calculation:
+        // multiplying the employee's salary by an uint64 number of seconds. Therefore, timeDiff will always fit in 64.
+        // Nevertheless, we are performing a sanity check at the end to ensure the computed last payroll timestamp
+        // is not greater than the current timestamp.
+
+        uint256 lastPayrollDate = employee.lastPayroll.add(timeDiff);
+        require(lastPayrollDate <= uint256(getTimestamp64()), ERROR_LAST_PAYROLL_DATE_TOO_BIG);
```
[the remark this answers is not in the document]
bingen: Actually if we use `getTimestamp` we don't need the cast, right?

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -594,21 +595,54 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+    /**
+     * @dev Calculates the date timestamp corresponding to the requested paying amount based on the employee's last payroll date.
+     * @param _employeeId Employee's identifier.
+     * @param _payedAmount Amount payed to the employee to query the timestamp of.
+     * @return The date timestamp in seconds corresponding to the given paying amount based on the employee's last payroll date.
+     */
+    function _getLastPayroll(uint256 _employeeId, uint256 _payedAmount) internal view returns (uint64) {
+        Employee storage employee = employees[_employeeId];
+
+        uint256 timeDiff = _payedAmount.div(employee.denominationTokenSalary);
```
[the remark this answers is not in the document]
bingen: I don't get it: doesn't Solidity always truncate?
Anyway, I agree this doesn't look good. I think we should try to avoid relying on division.

future-apps/payroll/contracts/Payroll.sol
bingen: Then I would just keep the old name ;-)

future-apps/payroll/contracts/test/mocks/PayrollMock.sol
bingen: That's true, but it's also true that it's kind of weird that we are mixing camel case and snake case ;-)

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -32,10 +31,11 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
     bytes32 constant public MODIFY_RATE_EXPIRY_ROLE = keccak256("MODIFY_RATE_EXPIRY_ROLE");
 
     uint128 internal constant ONE = 10 ** 18; // 10^18 is considered 1 in the price feed to allow for decimal calculations
+    uint256 internal constant MAX_ALLOWED_TOKENS = 20; // for loop in `payday()` uses ~270k gas per token
+    uint256 internal constant MAX_UINT256 = uint256(-1);
```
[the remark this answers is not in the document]
facuspagnuolo: If you agree, I'd prefer being explicit then :)

future-apps/payroll/contracts/Payroll.sol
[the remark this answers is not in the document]
sohkai: It's exactly the truncating that causes this behaviour.

If you had a salary per second of 100 USD, I could request 99 USD here to get a `timeDiff` of 0.

future-apps/payroll/contracts/Payroll.sol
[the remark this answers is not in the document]
sohkai: Ah, good point!

future-apps/payroll/contracts/Payroll.sol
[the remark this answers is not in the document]
sohkai: 👍 In that case, it seems like we have two paths:

1. Introduce "re-enabling" functionality for terminated employees in the future, and avoid storage gaps now
1. Use "re-adding" someone as a new employee as the way to "re-hire" someone, and never allow terminated employees to be re-enabled with the same ID

Not sure which one is the best real world analogy, but I am also starting to get the sense that assigning a new employee ID is not so bad for rehires.

future-apps/payroll/contracts/Payroll.sol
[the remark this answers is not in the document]
bingen: > It's exactly the truncating that causes this behaviour.

Yes, that's why I didn't understand this:

> We should check if the division was perfect, and if not, floor it.

Were you proposing to truncate that division? (As numbers are always positive I understand flooring and truncating are synonyms in our context)

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -53,30 +53,37 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
-    IFeed public feed;
     uint64 public rateExpiryTime;
 
-    mapping(uint256 => Employee) private employees;     // employee ID -> employee
-    // Employees start at index 1, to allow us to use employees[0] to check for non-existent address
-    // mappings with employeeIds
     uint256 public nextEmployee;
+    mapping(uint256 => Employee) private employees;     // employee ID -> employee
     mapping(address => uint256) private employeeIds;    // employee address -> employee ID
     mapping(address => bool) private allowedTokens;
```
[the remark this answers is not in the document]
facuspagnuolo: I sometimes think it's better to provide a useful internal interface rather than allowing subclasses to play directly with the storage structure, just in favor of encapsulation... but with smart contracts, as we cannot predict what we may need in a future, I prefer being more flexible, let's make them `internal` then :)

future-apps/payroll/contracts/Payroll.sol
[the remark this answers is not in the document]
facuspagnuolo: There is no problem on changing it at all, but I think we are losing information if we do. For example, I always try to use `uint256` instead of `uint`, but I also know that `SafeMath` has been out there for a while haha. Anyway, let's pick a convention and move forward. I'll let you both decide :)

future-apps/payroll/contracts/test/mocks/PayrollMock.sol
facuspagnuolo: Will take care of improving the way we mock time-related stuff in a separate PR :)

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: The idea of using the word "reimburse" was to name the action of cashing out your accrued value, no te rename accrued value, somehow an analogy of `payday` to `salary`. Therefore, avoiding the word "reimburse" won't hide us from the fact that we are mixing accounting concepts here.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -137,192 +139,145 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
-     * @param _amount Amount to add
+     * @notice Adds `_amount` to accrued value for employee #`_employeeId`.
+     * @param _employeeId Employee's identifier.
+     * @param _amount Amount be added to the employee's accrued value.
      */
-    function addAccruedValue(
-        uint256 _employeeId,
-        uint256 _amount
-    )
+    function addAccruedValue(uint256 _employeeId, uint256 _amount)
```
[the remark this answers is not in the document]
facuspagnuolo: Let's follow this conversation [here](https://github.com/aragon/aragon-apps/pull/744/files?file-filters%5B%5D=.sol&hide-deleted-files=true#r271881379) just to avoid multiple threads with the same topic :)

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -53,30 +53,37 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
     uint64 public rateExpiryTime;
 
-    mapping(uint256 => Employee) private employees;     // employee ID -> employee
-    // Employees start at index 1, to allow us to use employees[0] to check for non-existent address
-    // mappings with employeeIds
     uint256 public nextEmployee;
+    mapping(uint256 => Employee) private employees;     // employee ID -> employee
     mapping(address => uint256) private employeeIds;    // employee address -> employee ID
     mapping(address => bool) private allowedTokens;
     address[] internal allowedTokensArray;
```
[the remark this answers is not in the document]
facuspagnuolo: I'm not sure I follow, I thought we were using an array here simply because we need to iterate through that list.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -526,64 +534,57 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
-            toDate = timestamp;
-        }
-
-        // Compute owed amount
-        uint256 owed = employee.accruedValue.add(_getOwedSalary(_employeeId, toDate));
-        if (owed == 0) {
+        // Compute amount to be payed
+        uint256 owedAmount = _getOwedSalary(_employeeId);
+        if (owedAmount == 0 || owedAmount < _requestedAmount) {
             return false;
```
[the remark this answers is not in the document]
facuspagnuolo: I think we can improve this a bit :)

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -526,64 +534,57 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+     */
+    function _reimburse(uint256 _employeeId, uint256 _requestedAmount) internal returns (bool somethingPaid) {
+        Employee storage employee = employees[_employeeId];
 
-        // Try to remove employee
-        if (employee.endDate <= timestamp && employee.accruedValue == 0) {
-            delete employeeIds[employee.accountAddress];
-            delete employees[_employeeId];
+        // Compute amount to be payed
+        if (employee.accruedValue == 0 || employee.accruedValue < _requestedAmount) {
```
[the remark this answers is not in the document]
facuspagnuolo: Same as [here](https://github.com/aragon/aragon-apps/pull/744#discussion_r271429264)

future-apps/payroll/contracts/Payroll.sol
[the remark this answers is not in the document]
sohkai: Oops, amended to use "ceiling" instead, floor was the wrong direction 😄

future-apps/payroll/contracts/Payroll.sol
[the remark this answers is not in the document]
sohkai: Will move to `UINT256` in Finance (it's also easier to search for 😄)

future-apps/payroll/contracts/Payroll.sol
sohkai: I just realized bonuses would probably fall under accrued value as well, which would certainly not fall under reimbursements in the real world.

Wondering if it'd be too much to add another storage slot to keep track of these bonuses separately, just so we can use a different reference string. We could potentially use something like an enum argument to differentiate between the type of payment we'd like to do so the interface is streamlined into a single `payday()` and `partialPayday()` function.

I think salary + reimbursement + bonus _should_ cover all employee types of remuneration (anything related to a stock / token, e.g. options or dividends, would be a separate contract).

future-apps/payroll/contracts/Payroll.sol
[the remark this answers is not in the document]
bingen: Oh, ok, then it makes sense. Although it would be still inaccurate. I would keep the flooring and then adjust the real amount payed to that `timeDiff`, which in that example it would be zero and would always be less than requested.
Btw, if I had a salary of $100 per second, trust me, I wouldn't bother doing this trick, hahaha. This is of course a joke, but it's worth noting that the maximum error will be small, at most the amount of salary per second, which except for sport stars and other big salaries  would be less than a penny.
Another solution, easy but still slightly inaccurate, would be to define a minimum periodicity (i.e., `timeDiff`) for requesting salary. (I still like better the previous one)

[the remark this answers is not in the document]
sohkai: > The maximum error will be small, at most the amount of salary per second

Regardless of how small this is, you could do this infinitely many times as gas allowed.

If we go with our rough estimate that you can make 20 calls per block, you could fill entire blocks for as long as you could to "siphon" out money from the organization.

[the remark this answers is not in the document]
bingen: Yes, yes, I know, I never meant to leave it as it is (although I guess most of the times the cost of gas would be higher).

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: Will move this one to a separate issue

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: Moving this one to a separate issue

future-apps/payroll/test/contracts/Payroll.test.js
facuspagnuolo: I chose this value to make it simpler to deal with accruing salaries for the tests, but I also tested many different scenarios of expired rates :)

future-apps/payroll/contracts/Payroll.sol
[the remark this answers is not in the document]
facuspagnuolo: I think that's what we are doing currently, see [here](https://github.com/aragon/aragon-apps/pull/744/commits/c1b8831675d81158ab7072f2aa0c741619583043#diff-c669e10c65bc3c52e31525560eae501bR96).

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -594,21 +595,54 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+
+        // This function is only called from _payday, where we make sure that _payedAmount is lower or equal to the
+        // total owed amount, that is obtained from _getOwedSalary, which does exactly the opposite calculation:
+        // multiplying the employee's salary by an uint64 number of seconds. Therefore, timeDiff will always fit in 64.
+        // Nevertheless, we are performing a sanity check at the end to ensure the computed last payroll timestamp
+        // is not greater than the current timestamp.
+
+        uint256 lastPayrollDate = employee.lastPayroll.add(timeDiff);
+        require(lastPayrollDate <= uint256(getTimestamp64()), ERROR_LAST_PAYROLL_DATE_TOO_BIG);
+        return uint64(lastPayrollDate);
```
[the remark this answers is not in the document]
facuspagnuolo: I removed the obvious casts and I think it doesn't look cumbersome anymore, let me know your thoughts.

facuspagnuolo:
@sohkai I addressed all the minor comments in this PR and left the more important for further PRs, I'll merge this one in the meantime, thanks for the thorough review :)

future-apps/payroll/contracts/Payroll.sol
bingen: I still think here we can just use `getTimestamp()`.

facuspagnuolo: We are, see [here](https://github.com/aragon/aragon-apps/pull/744/files#diff-9d7adaffbcce19867f027a983a30bcc7R610). I think you were reviewing an old version :)

future-apps/payroll/contracts/Payroll.sol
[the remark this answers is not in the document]
facuspagnuolo: Moving this to a separate issue

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: Discussed offline: we will keep the `MAX_UINT256` capping here to avoid employees from cashing out their salary if they reached a huge owed amount.

sohkai:
@facuspagnuolo 

One note on

> Added string `reference` argument to `SendPayment` event.

Let's use a different variable name (e.g. `paymentReference`); I just learned that `reference` is a reserved keyword in pragma 0.5+.

future-apps/payroll/contracts/Payroll.sol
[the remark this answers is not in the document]
sohkai: Yep, although we could still do that with a manual index and a mapping. The main reason we don't need to make it a mapping is because the array's holding a constant-size structure.

See https://github.com/aragon/aragon-apps/issues/68 for some more context.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -64,15 +65,16 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
     address public denominationToken;
     uint64 public rateExpiryTime;
 
+    // Employees start at index 1, to allow us to use employees[0] to check for non-existent address
     uint256 public nextEmployee;
-    mapping(uint256 => Employee) private employees;     // employee ID -> employee
-    mapping(address => uint256) private employeeIds;    // employee address -> employee ID
-    mapping(address => bool) private allowedTokens;
+    mapping(address => uint256) internal employeeIds;    // employee address -> employee ID
+    mapping(uint256 => Employee) internal employees;     // employee ID -> employee
```
sohkai: Could we move `uint256 public nextEmployee` to be after this line (since it acts as its length)?

facuspagnuolo: Sure, will do it in the next payroll PR :)

PR #753, opened 2019-03-28
[DO NOT MERGE] GitHub actions

2color:
## What 
- Replicate travis logic in Github Actions In the same vein as https://github.com/aragon/aragon/pull/663


2color:
After many attempts, I keep on hitting different problems:
- The tests run in containers. They share the same repository workspace and when all run in parallel they seem to fail for an unknown reason (@bingen and I tried to find the root cause to no avail). 
- To run the tests locally I used [act](https://github.com/nektos/act).
- The workaround is to group the tests that work when run in parallel 
- The last failure is from the coverage test timing out. 
- I think the major difference to Travis is that Travis installs – albeit with better caching – for every step, whereas Github actions actually share a workspace for the whole workflow. 

Combination | result
-------------| ----
test:finance | passed
test:vault, test:survey, test:token-manager | passed
test:token-manager, test:vault, lint, test:shared, test:survey | passed
test:finance, test:voting  | passed
test:voting, test:finance, test:token-manager | passed
test:vault, test:finance | passed
test:finance, test:survey, test:token-manager, test:vault  | failed
test:survey, test:token-manager, test:vault, test:voting  | failed
lint, test:shared, test:survey, test:token-manager, test:vault, test:finance   | failed

sohkai:
> The tests run in containers. They share the same repository workspace and when all run in parallel they seem to fail for an unknown reason (@bingen and I tried to find the root cause to no avail).

Is this because they're using the same ganache instance? Or the ganache instances are colliding?

bingen:
Each process runs its own ganache instance in an isolated docker container, so we are pretty sure that was not the reason. Actually, withe less processes in parallel it works. it seems like it runs out of resources or something like that, because all of them work if ran only with ~3 concurrent tests, but they fail when trying to run them all.

2color:
> What were the failures?

In most cases it was the ganche process in a container that crashed without giving info why. 

My hunch is that is has to do with something in the shared workspace across the containers that causes it to fail. But as Bingen said, it could also be a resources problem. 

Since it's not happening when a single container is running, we could use [Docker's resource constraints](https://docs.docker.com/config/containers/resource_constraints/) to test a single container on meagre resources.

sohkai:
Going to close this for now, as recent improvements to Travis have made it bearable. We can revisit Github actions for CI when they release their new upgrades.

PR #754, opened 2019-03-28
[DO NOT MERGE] Add Payroll to travis

sohkai:


facuspagnuolo:
We are getting closer 🙌 
The coverage tool seems to get stuck with the Payroll contract. It may be related to the fact that I rolled back [this tweak](https://github.com/aragon/aragon-apps/pull/731/files#diff-9d7adaffbcce19867f027a983a30bcc7L21). It'd be sad if we do need to remove sth core like an inheritance statement to make this tool work :(

sohkai:
> The coverage tool seems to get stuck with the Payroll contract.

I think this is just `solidity-coverage` taking _forever_ to run, because it instruments everything recursively. You can see it timed out at 50min (so it shouldn't be a compile issue).

If you think about it, Payroll depends on Finance, which depends on Vault. Finance's coverage already takes ~45min, (probably ~35min of that is on instrumenting). I can only imagine how long it'll take Payroll to instrument 😄.

What do you think about allowing the Payroll's coverage test to fail, or not even including it?

PR #747, opened 2019-03-27
VotePanelContent: add no token cost to info card

AquiGorka:
Reference: https://github.com/aragonone/product/issues/30

PR #758, opened 2019-04-04
Survey: add no token cost info card

sohkai:
**NOT YET TESTED**

Replicates #747 for the Survey app.


sohkai:
Going to merge this now; we will move the Survey app to its own repo soon.

PR #764, opened 2019-04-05
Token Manager: use aragonAPI for React

bpierre:
Contains only the changes needed to support aragonAPI for React (not a full conversion to React Hooks).


apps/token-manager/app/src/components/AppLayout.js
```diff
@@ -21,18 +22,24 @@ const AppLayout = ({
           appBar={
             <AppBar>
               <AppBarContainer
-                style={{ padding: below('medium') ? '0' : '0 30px' }}
+                style={{ padding: below('medium') ? '0' : '0 30px 0 10px' }}
               >
                 <Title>
-                  {below('medium') && <MenuButton onClick={onMenuOpen} />}
-                  <TitleLabel>{title}</TitleLabel>
+                  {displayMenuButton && <MenuButton onClick={requestMenu} />}
```
sohkai: Could we clean up the `MenuButton` implementations a bit, since the icon is now exposed in aragonUI?

2color:
This is really nice!

PR #766, opened 2019-04-05
Payroll: Improve timestamps mocking

facuspagnuolo:
This PR improves timestamp mocking in a more consistent way and provides a new common mocking contract that can be easily re-used.


facuspagnuolo:
I wouldn't be finding why the coverage has decreased, any hints where should I look?

sohkai:
> I wouldn't be finding why the coverage has decreased, any hints where should I look?

Probably due to the number of coverage lines being decreased 🤷‍♂️.

---------

@facuspagnuolo Could you make this PR solely against Payroll, and then we'll open another one for the other apps? The reason is that the tests and contracts have changed in the `next` branch and we should work off of there instead.

PR #765, opened 2019-04-05
Finance: use aragonAPI for React

bpierre:
Contains only the changes needed to support aragonAPI for React (not a full conversion to React Hooks).

PR #767, opened 2019-04-05
Finance filters refresh

bpierre:
Builds on #765 (don’t merge before).

## Rationale

- The download button has been designed as a DropDown to support several export formats, but only one format (CSV) is supported at the moment. The issue is that the DropDown style has been kept while the component now behaves like a button. The new version uses the same style we have everywhere else for icon buttons.
- The row is now too crowded to be displayed properly at all sizes. The new version is moving the filters row below the title in large views, as it was already the case in smaller views.
- The labels have been updated: “Date range” becomes “Period”, and “Transfer type” becomes “Type” (we are in the “Transfers” filters so it is already implied).
- I think a proper redesign would still be welcomed, as it makes little sense to have the download button with the filters, and way the filters layout is not optimal.

## Before

filters-before-320
filters-before-600
filters-before-1000
filters-before-1450

## After
filters-after-320
filters-after-600
filters-after-1000
filters-after-1450


apps/finance/app/src/components/Transfers.js
```diff
@@ -128,7 +123,7 @@ class Transfers extends React.PureComponent {
     const { selectedDateRange } = this.state
     const start = format(selectedDateRange.start, 'yyyy-MM-dd')
     const end = format(selectedDateRange.end, 'yyyy-MM-dd')
-    return `Finance_(${start}_to_${end}).csv`
+    return `transfers_${start}_to_${end}.csv`
```
sohkai: Could we use the Finance app's proxy address here, like `0x1a0adf8909(Finance)_transfers_...`

apps/finance/app/src/components/Transfers.js
```diff
@@ -179,253 +174,110 @@ class Transfers extends React.PureComponent {
...
+          </Title>
+          <TransfersFilters
+            dateRangeFilter={selectedDateRange}
+            onDateRangeChange={this.handleDateRangeChange}
+            tokenFilter={selectedToken}
+            onTokenChange={this.handleTokenChange}
+            transferTypeFilter={selectedTransferType}
+            onTransferTypeChange={this.handleTransferTypeChange}
+            compactMode={compactMode}
+            opened={filtersOpened}
```
sohkai: Could we move `compactMode` and `opened` to the top of the props list?

apps/finance/app/src/components/Transfers.js
```diff
@@ -179,253 +174,110 @@ class Transfers extends React.PureComponent {
...
+            symbols={symbols}
+            transferTypes={TRANSFER_TYPES_STRING}
```
sohkai: Could we rename `symbols` to `tokenSymbols`, and move it together with the token filter and onchange handler?

Similarly, could we move `transferTypes` to be with the transfer type filter and onchange handler?

apps/finance/app/src/components/Transfers.js
```diff
@@ -179,253 +174,110 @@ class Transfers extends React.PureComponent {
...
+                  // Sort by date descending
+                  compareDesc(dateLeft, dateRight)
+                )
+                .slice(0, displayedTransfers)
+                .map(transfer => (
+                  <TransferRow
+                    key={transfer.transactionHash}
+                    token={tokenDetails[toChecksumAddress(transfer.token)]}
+                    transaction={transfer}
+                    smallViewMode={compactMode}
```
sohkai: Maybe we can also rename `smallViewMode` to `compactMode`?

apps/finance/app/src/components/TransfersFilters.js
```diff
@@ -0,0 +1,170 @@
+import React from 'react'
+import styled, { css } from 'styled-components'
+import { DropDown, springs, theme, unselectable } from '@aragon/ui'
+import { Spring, animated } from 'react-spring'
+import DownloadButton from './DownloadButton'
+import DateRange from './DateRange/DateRangeInput'
+
+const TransfersFilters = ({
```
sohkai: Small nitpick: what do you think about `TransferFilters` instead as the name?

apps/finance/app/src/components/TransfersFilters.js
```diff
@@ -0,0 +1,170 @@
...
+  }
+  ${p =>
+    p.compact
+      ? ''
+      : css`
+          width: 100%;
+          margin: 0;
+          justify-content: space-between;
+          margin-left: 0;
+          margin-right: 0;
```
sohkai: Seems like we don't need these two margin declarations

apps/finance/app/src/components/Transfers.js
```diff
@@ -179,253 +174,110 @@ class Transfers extends React.PureComponent {
...
+            symbols={symbols}
+            transferTypes={TRANSFER_TYPES_STRING}
+            downloadFileName={this.getCsvFilename()}
+            downloadUrl={this.encodeDataToCsv(filteredTransfers, tokenDetails)}
+          />
+        </Header>
+        {filteredTransfers.length === 0 ? (
+          <NoTransfers>
+            <p>
+              No transfers found.{' '}
```
sohkai: Could we also add the date period here? I think it's a bit confusing that we automatically apply the dates so a lot of users will just see "No transfers found." if they haven't made a transaction in the current month.

bpierre:
> @dizzypaty I do like having the "button" look to the download button; right now I get a bit confused if I can actually click on it or not.

If we decide to go with a button style, I think we should also align the “filters” button with it, and have a consistent button style for “icon buttons on the blue background” (assuming the confusion comes from having these buttons on the blue background?).

> On the mobile view as well, what do you think about centering the icon in the cell?

Everything else (the filters on the left, and the filter button above it) is either left or right aligned. Having this specific button slightly misaligned with the filters button above wouldn’t look good IMO.

But what might look a bit weird is that we are aligning the header icons on the illustrations directly, while the icon buttons are aligned on the button around it, with its padding:

<p align=center>

For these cases, I think we could shift these buttons so they look aligned, the same way we do it already with text buttons. The `IconButton` has `5px` padding on each side, so we could shift it `5px` on either side it is aligned:

<p align=center>

And after the icons internal alignment will be more consistent (see https://github.com/aragon/aragon-ui/pull/346), we could also move these buttons `2px` or `3px` more, so that they would look as aligned as possible with the gutter.

But if we decide to use another style for these (the filters toggle and the export button), then we could also move the label inside of it (to have an icon + label button), and maybe only have the icon in smaller views.

apps/finance/app/src/components/TransfersFilters.js
bpierre: I actually started by having it named `TransferFilters`, then I renamed it to `TransfersFilters`, not only because it applies to transfers (it’s ok to not be strictly correct if a name is nicer), but because it is a component that is meant to be used from the `Transfers` component. What do you think?

apps/finance/app/src/components/Transfers.js
bpierre: What about making it clearer that filters are applied in general?

> No transfers found matching the current filters.
> [Clear current filters]

/cc @dizzypaty

apps/finance/app/src/components/TransfersFilters.js
sohkai: Makes sense 👍

sohkai:
Mmm the left / right alignment makes sense 👍.

> assuming the confusion comes from having these buttons on the blue background

Yes, I think so. The other filters "pop out" with a box shadow and look like they're meant to be interacted with.

apps/finance/app/src/components/Transfers.js
dizzypaty: Yes, I agree with making it general – how about we say: 

"No transfers match your filter selection. [Clear filters](url)"

dizzypaty:
> @dizzypaty I do like having the "button" look to the download button; right now I get a bit confused if I can actually click on it or not.

I agree @sohkai and I think we need to explore a different visual style for these icon buttons in general, even more so if they're displayed on the blue background. We need to define a rationale for our **layer and depth logic** (mainBG, buttonsBG, popovers/modals, etc.) and keep it consistent so it's clearer when UI elements are interactive. Until we do this, my inclination would be to keep the way we present these icon buttons consistent, and we have 3 of them in this screen.

apps/finance/app/src/components/Transfers.js
sohkai: I still think it'd be really useful to show the date range being applied, because at first glance it'll appear immediately broken if there's no transactions and the user knows their finance app had transactions previously.

This is especially true on mobile, where the filters are hidden by default.

Another problem is that we'd need a definition for what "clear filters" mean for the date range. Should we just set it to Nov of last year (since it's when it launched)? Should we be smarter and set it based on the app's initialization block time (actions are only possible on the app after it's been initialized)?

bpierre: Oh right, I was assuming the date range was not applied until the user selects one!

@sohkai I think your solution would be an improvement in the current state of things, but it would still be quite confusing on mobile with the current design in some cases: if there is let’s say one item, and the others are hidden, nothing would indicate the current date range as it is only displayed in the empty state.

About the “clear filters”: what about the case where a specific date range has been selected, but also another filter? Do we reset the date range as well? Or do we consider it not being a filter?

I think it could be easier for the time being to only apply the date range once the user selects it (having the field empty by default), like the other filters, at least until this part gets redesigned.

Thoughts @dizzypaty @sohkai?

sohkai: > I think it could be easier for the time being to only apply the date range once the user selects it (having the field empty by default), like the other filters, at least until this part gets redesigned.

I think this would be preferable, but does the component allow us to do this? It might look weird being completely blank as its default state.

dizzypaty: @sohkai, @bpierre I also assumed none of filters were applied until users selection. 

Having a look at this with more detail now: 
- Date range should be deselected by default (until user makes a date selection)
- We should indicate today’s date with a slightly different visual style than the date selection – will do a quick mock-up!
- We should include a ‘Clear all’ button, that removes the current filter selection

dizzypaty: @sohkai , @bpierre Based on the previous comments, these are some changes to the date range filter:

### Default state (no filters applied, calendar displays today's date)
WebApp-1366px - Finance main view
### Date range selected 
WebApp-1366px - Finance main view-1
### Filters applied 
WebApp-1366px - Finance main view-2

**Note:** after trying a couple of things yesterday, I think we can leave the 'Clear filters' button out of these changes for the release and then rethink the layout properly. 

[Figma file](https://www.figma.com/file/JUMdeJefkK1TZtYmU8POnSd5/Finance-filters?node-id=0%3A1)

bpierre:
:shipit: :shipit: :shipit: :shipit: :shipit: :shipit:

PR #770, opened 2019-04-05
Token Manager: small improvements for ITokenController interface

sohkai:
Changes the `ITokenController` methods to be `external`, as they never need to be called from within the Token Manager itself.


tintinweb:
✓ minor changed func visibility -> pub->ext; lex. move func (no change)

PR #771, opened 2019-04-05
Token Manager: remove unused imports

sohkai:


tintinweb:
✓ minor cleanup as suggested in review #723

PR #775, opened 2019-04-06
Payroll: Support employees bonus

### LINKED ISSUE #760: Payroll: Support adding and cashing out bonuses
Add another storage slot to keep track of bonuses separately, and make sure we use a different reference string. We could potentially use something like an enum argument to differentiate between the type of payment we'd like to do so the interface is streamlined into a single `payday()` and `partialPayday()` function.

facuspagnuolo:
Fixes #760 

**Caveat**
I needed to refactor the external interface since we were reaching the bytecode max size limit. I removed `payday`, `partialPayday`, `reimbursement` and , `partialReimbursement` in favor of a single [`payday`](https://github.com/aragon/aragon-apps/pull/775/files#diff-9d7adaffbcce19867f027a983a30bcc7R320) method allowing the user to specify which payment type they want to cash out.


facuspagnuolo:
I accidentally closed it due to a bad PR link

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -751,7 +745,7 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
      * @param _employeeId Employee's identifier
      * @return True if the employee's end date has not been reached yet, false otherwise
      */
-    function _isEmployeeActive(uint256 _employeeId) internal returns (bool) {
+    function _isEmployeeActive(uint256 _employeeId) internal view returns (bool) {
```
sohkai: Oof, we should double check if anything else is missing these modifiers, I'm surprised we haven't caught the compiler warnings yet.

future-apps/payroll/test/contracts/Payroll_bonuses.test.js
```diff
@@ -0,0 +1,512 @@
...
+
+  const NOW = 1553703809 // random fixed timestamp in seconds
+  const ONE_MONTH = 60 * 60 * 24 * 31
+  const TWO_MONTHS = ONE_MONTH * 2
+  const RATE_EXPIRATION_TIME = TWO_MONTHS
+
+  const PCT_ONE = bigExp(1, 18)
+  const TOKEN_DECIMALS = 18
+  const BONUS_PAYMENT_TYPE = 2
+  const PAYROLL_PAYMENT_TYPE = 0
```
sohkai: Might be nice to map this out into a constant in the utils

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -245,6 +251,19 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
         _terminateEmployeeAt(_employeeId, _endDate);
     }
 
+    /**
+     * @notice Add `_amount` to bonus amount for employee #`_employeeId`
```
```suggestion
     * @notice Add `_amount` to bonus for employee #`_employeeId`
```

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -502,6 +521,17 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
     }
 
+    /**
+     * @dev Add a requested amount to the bonus amount for a given employee
+     * @param _employeeId Employee's identifier
+     * @param _amount Amount be added to the employee's bonus amount
+     */
+    function _addBonus(uint256 _employeeId, uint256 _amount) internal {
+        Employee storage employee = employees[_employeeId];
+        employee.bonus = employee.bonus.add(_amount);
```
sohkai: Wondering if we should max the bonus and accrued value at `MAX_UINT256` as well, so everything just uses that as a hard stop?

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -287,49 +307,40 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
-    }
+        if (_type == PaymentType.Payroll) {
+            (uint256 currentOwedSalary, uint256 totalOwedSalary) = _getOwedSalary(employeeId);
+            payingAmount = _ensurePayingAmount(totalOwedSalary, _requestedAmount);
+            _updateEmployeeStatusBasedOnPaidPayroll(employeeId, payingAmount, currentOwedSalary);
+        }
+        else if (_type == PaymentType.Reimbursement) {
+            uint256 owedAccruedValue = employee.accruedValue;
+            payingAmount = _ensurePayingAmount(owedAccruedValue, _requestedAmount);
+            employee.accruedValue = owedAccruedValue.sub(payingAmount);
```
sohkai: Maybe we should rename `accruedValue` to `reimbursement` now 😄

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: In this case it will be hold by SafeMath, the employer won't be able to overflow the employee's bonus amount basically. The same holds now for reimbursements (accrued value) since we split it from accrued salary. The only thing that could overflow is the owed salary, which is a concern for accrued salary and payroll payments :/

facuspagnuolo:
@sohkai addressed all your comments here https://github.com/aragon/aragon-apps/pull/776/commits/c3591b71ad3386f6cd39af3e73cf0af049d08459

PR #776, opened 2019-04-07
Payroll: Ceil last payroll date calculation

### LINKED ISSUE #763: Payroll: Ensure last payroll date accurate calculation
From @sohkai 

> We should check if the division was perfect, and if not, take its ceiling.
Right now, you could request a tiny amount of salary that results in `timeDiff == 0`, basically getting free money out of the contract as many times as you could reasonable call it within gas limits.
We should also add a warning that you may lose up to a second's worth of payroll if you use the "request partial amount" feature.

See [here](https://github.com/aragon/aragon-apps/pull/744#discussion_r271376553).

facuspagnuolo:
Fixes #763


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -602,6 +602,13 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
 
         uint256 timeDiff = _payedAmount.div(employee.denominationTokenSalary);
 
+        // We check if the division was perfect, and if not, taking its ceiling to avoid giving away tiny amounts of salary.
```
```suggestion
        // We check if the division was perfect, and if not, take its ceiling to avoid giving away tiny amounts of salary.
```

future-apps/payroll/test/contracts/Payroll_payday.test.js
```diff
@@ -282,9 +282,24 @@ contract('Payroll payday', ([owner, employee, anotherEmployee, anyone]) => {
                 })
 
                 context('when the requested amount is lower than the total owed salary', () => {
-                  const requestedAmount = currentOwedSalary - 10
+                  context('when the requested amount is represents less than a second of the earnings', () => {
```
sohkai:
```suggestion
                  context('when the requested amount represents less than a second of the earnings', () => {
```

future-apps/payroll/test/contracts/Payroll_payday.test.js
```diff
@@ -282,9 +282,24 @@ contract('Payroll payday', ([owner, employee, anotherEmployee, anyone]) => {
...
+                      const previousLastPayrollDate = (await payroll.getEmployee(employeeId))[5]
+
+                      await payroll.payday(PAYROLL_PAYMENT_TYPE, requestedAmount, { from })
+
+                      const currentLastPayrollDate = (await payroll.getEmployee(employeeId))[5]
+                      assert.equal(currentLastPayrollDate.toString(), previousLastPayrollDate.plus(1).toString(), 'last payroll date does not match')
+                    })
+                  })
+
+                  context('when the requested amount is represents more than a second of the earnings', () => {
```
sohkai:
```suggestion
                  context('when the requested amount represents more than a second of the earnings', () => {
```

PR #778, opened 2019-04-08
Payroll: Add reentrancy guard to payment transfers

### LINKED ISSUE #759: Payroll: Add reentrancy guard to transfers loop
From @sohkai (see [here](https://github.com/aragon/aragon-apps/pull/744#discussion_r271393365))

> We should add a reentrancy guard to this function and `Payroll.determineAllocation()` (or perhaps all the functions with the `employeeMatches` modifier).
Right now, with ETH, you can re-enter into `Payroll.determineAllocation()` and modify your allocation so that a later token becomes 100% of your allocation. The worse case of this is if you set ETH to 100%, and then upon paying ETH, your contract re-enters into `Payroll.determineAllocation()` and sets a later token to be 100%.

facuspagnuolo:
Fixes #759 

The CI failed because we reached the max size limit of a contract, I'm proposing to simplify the external interface removing some redundant functions. I'm dropping `addEmployeeNow` and `terminateEmployeeNow`, in favour of leaving simply `addEmployee` and `terminateEmployee`.

### DIFF
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -176,19 +176,6 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
     }
 
     /**
-     * @notice Add employee with address `_accountAddress` to Payroll with a salary of `_initialDenominationSalary` per second
-     * @param _accountAddress Employee's address to receive payroll
-     * @param _initialDenominationSalary Employee's salary, per second in denomination token
-     * @param _role Employee's role
-     */
-    function addEmployeeNow(address _accountAddress, uint256 _initialDenominationSalary, string _role)
-        external
-        authP(ADD_EMPLOYEE_ROLE, arr(_accountAddress, _initialDenominationSalary, getTimestamp()))
-    {
-        _addEmployee(_accountAddress, _initialDenominationSalary, _role, getTimestamp64());
-    }
-
-    /**
      * @notice Add employee with address `_accountAddress` to Payroll with a salary of `_initialDenominationSalary` per second, starting on `@transformTime(_startDate)`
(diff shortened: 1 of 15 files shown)


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -687,7 +687,7 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
      * @param _type Payment type to be performed
      * @return True if there was at least one token transfer
      */
-    function _transferTokensAmount(uint256 _employeeId, PaymentType _type, uint256 _totalAmount) internal returns (bool somethingPaid) {
+    function _transferTokensAmount(uint256 _employeeId, PaymentType _type, uint256 _totalAmount) internal nonReentrant returns (bool somethingPaid) {
```
sohkai: We should protect `determineAllocation()` as well, since that's the one that can change the later payments on an re-entrancy. It may be safest just to attach `nonReentrant` on all methods with the `employeeExists` modifier to prevent things like the address changing, etc.

facuspagnuolo:
@sohkai would be great if you can take another look at these changes. I updated the description and added some tests.

future-apps/payroll/contracts/test/mocks/MaliciousEmployee.sol
```diff
@@ -0,0 +1,105 @@
...
+        require(_to != address(0));
+
+        balances[_from] = balances[_from].sub(_value);
+        balances[_to] = balances[_to].add(_value);
+        allowed[_from][msg.sender] = allowed[_from][msg.sender].sub(_value);
+        emit Transfer(_from, _to, _value);
+
+        return true;
+    }
+}
```
sohkai: 😍 😍 😍

PR #780, opened 2019-04-08
Shared: Organize ganache cli script

facuspagnuolo:
Main idea was to avoid relying on `nc` in order to kill a process that is running on the port we want to use. Instead I'm adding adding a `cleanup` function with `trap` to ensure we always end the rpc process we have started. If there is a foreign process running on that port we shouldn't kill it, it's expected to fail rather than taking over a port while killing the process that is using it.

It may fix #775 since we are not relying on `nc` anymore.

### DIFF
--- a/shared/test-helpers/ganache-cli.sh
+++ b/shared/test-helpers/ganache-cli.sh
@@ -1,50 +1,58 @@
 #!/usr/bin/env bash
 
 # Exit script as soon as a command fails.
-set -o errexit
+set -o errexit -o pipefail
 
-if [ "$SOLIDITY_COVERAGE" = true ]; then
-  testrpc_port=8555
-else
-  testrpc_port=8545
-fi
+# Executes cleanup function at script exit.
+trap cleanup EXIT
 
-testrpc_running() {
-  nc -z localhost "$testrpc_port"
+cleanup() {
+  # Kill the RPC instance that we started (if we started one and if it's still running).
+  if [ -n "$rpc_pid" ] && ps -p $rpc_pid > /dev/null; then
+    kill -9 $rpc_pid
+  fi
+}
+
+start_ganache() {
+  echo "Starting ganache-cli..."
+  npx ganache-cli -i ${NETWORK_ID} -l ${GAS_LIMIT} -e ${BALANCE} -p ${PORT} > /dev/null &
+  rpc_pid=$!
+  sleep 3
+  echo "Running ganache-cli with pid ${rpc_pid}"
 }
 
 start_testrpc() {
-  echo "Starting our own testrpc instance at port $testrpc_port"
-  if [ "$SOLIDITY_COVERAGE" = true ]; then
-    npx testrpc-sc -i 16 --gasLimit 0xfffffffffff --port "$testrpc_port"  > /dev/null &
-  else
-    npx ganache-cli -i 15 --gasLimit 50000000 --port "$testrpc_port" > /dev/null &
-  fi
-
-  testrpc_pid=$!
+  echo "Starting testrpc-sc..."
+  npx testrpc-sc -i ${NETWORK_ID} -l ${GAS_LIMIT} -e ${BALANCE} -p ${PORT} > /dev/null &
+  rpc_pid=$!
+  sleep 3
+  echo "Running testrpc-sc with pid ${rpc_pid}"
 }
 
-if testrpc_running; then
-  echo "Killing testrpc instance at port $testrpc_port"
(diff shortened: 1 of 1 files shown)


shared/test-helpers/ganache-cli.sh
```diff
@@ -1,50 +1,58 @@
...
-result=0
 if [ "$SOLIDITY_COVERAGE" = true ]; then
-  npx solidity-coverage "$@"
-  result=$?
-elif [ "$TRUFFLE_TEST" = true ]; then
-  truffle test --network rpc "$@"
-  result=$?
-fi
+  PORT=8555
+  BALANCE=10000
```
sohkai: Perhaps we should increase the balance here to be the same? No reason for the coverage account balance to be lower, I'd imagine 😄

shared/test-helpers/ganache-cli.sh
```diff
@@ -1,50 +1,58 @@
...
+  GAS_LIMIT=0xfffffffffff
+  NETWORK_ID=16
 
-if [ ! -z "$testrpc_pid" ]; then
-  kill -9 $testrpc_pid
-fi
+  start_testrpc
+  measure_coverage
+else
+  PORT=8545
```
sohkai: We _could_ let this be parametrizable with an env var and use `8545`/`8555` as defaults if the env var isn't set. It should let us spin up multiple parallel instances in the future quite easily.

shared/test-helpers/ganache-cli.sh
```diff
@@ -1,50 +1,58 @@
...
+  BALANCE=100000
+  GAS_LIMIT=50000000
```
sohkai: We were setting a crazy amount before, but I wonder if this would work with just 8m so it's more similar to the real world gas limit?

facuspagnuolo: Not crazy at all, let's try if the tests pass :)

shared/test-helpers/ganache-cli.sh
```diff
@@ -1,50 +1,64 @@
 #!/usr/bin/env bash
 
 # Exit script as soon as a command fails.
-set -o errexit
+set -o errexit -o pipefail
```
bingen: Not too familiar with this. Why do we need it? I don't see any pipe script.

shared/test-helpers/ganache-cli.sh
```diff
@@ -1,50 +1,64 @@
...
 
-  testrpc_pid=$!
+setup_testing_variables() {
+  PORT=${PORT-8545}
+  BALANCE=${BALANCE-100000}
+  GAS_LIMIT=${GAS_LIMIT-8000000}
+  NETWORK_ID=${NETWORK_ID-15}
 }
 
-if testrpc_running; then
```
bingen: Aren't we killing previous instances of testrpc anymore? Is this on purpose?

facuspagnuolo: It was on purpose: the idea was to avoid relying on `nc` in order to kill a process that is running on the port we want to use. Instead I'm adding adding a `cleanup` function with `trap` to ensure we always end the rpc process we have started. If there is a foreign process running on that port we shouldn't kill it, it's expected to fail rather than taking over a port while killing the process that is using it.

shared/test-helpers/ganache-cli.sh
facuspagnuolo: True, it was instinctively

shared/test-helpers/ganache-cli.sh
bingen: Makes total sense!

shared/test-helpers/ganache-cli.sh
bingen: Haha, cool, I think it's ok to leave it anyway if you want (just in case), I just wanted to make sure I was not missing anything ;-)

shared/test-helpers/ganache-cli.sh
sohkai: @facuspagnuolo How does ganache error if it can't bind to the requested port? If it's not very good, would it still make sense to test it before trying to run ganache on the same port so the error message is more human-readable?

facuspagnuolo: I will fail with a classic `EADDRINUSE` error, see:
```
Ganache CLI v6.4.2 (ganache-core: 2.5.4)
listen EADDRINUSE 127.0.0.1:8545
```

sohkai: That looks pretty good 😄

PR #781, opened 2019-04-09
Finance: use constant for minimum period

sohkai:


tintinweb:
✓ minor lit to const. no sec issue.

PR #782, opened 2019-04-09
[DO NOT MERGE] Shared: PoC testing contracts versions

facuspagnuolo:
This is a proof of concept of how we could use previous versions of a contract in a test, it could be useful for testing upgrades for example. I proposed two alternatives here, either using a commit sha or a path to the build files of that version. 

Do not merge, it was just to show the idea.

### DIFF
--- a/future-apps/payroll/test/contracts/Payroll_upgrade.test.js
+++ b/future-apps/payroll/test/contracts/Payroll_upgrade.test.js
@@ -1,0 +1,16 @@
+const TruffleConfig = require('truffle-config')
+const TruffleContract = require('truffle-contract')
+const VersionedContractsProvider = require('@aragon/test-helpers/lib/VersionedContractsProvider')(web3, TruffleContract, TruffleConfig)
+
+contract('Payroll previous version', ([whatever]) => {
+  it.only('previous version should allow to call isAllowedToken when not initialized', async () => {
+    const commit = 'aff584ec78545cfb71415dcf2e56426e17a9f9cd'
+    const projectDir = 'future-apps/payroll'
+    const provider = VersionedContractsProvider.fromBranch({ commit, projectDir })
+
+    const PayrollOld = provider.getContract('Payroll')
+    const payroll = await PayrollOld.new()
+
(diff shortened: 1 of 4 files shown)


future-apps/payroll/test/contracts/Payroll_upgrade.test.js
```diff
@@ -0,0 +1,16 @@
+const TruffleConfig = require('truffle-config')
+const TruffleContract = require('truffle-contract')
+const VersionedContractsProvider = require('@aragon/test-helpers/lib/VersionedContractsProvider')(web3, TruffleContract, TruffleConfig)
+
+contract('Payroll previous version', ([whatever]) => {
+  it.only('previous version should allow to call isAllowedToken when not initialized', async () => {
+    const commit = 'aff584ec78545cfb71415dcf2e56426e17a9f9cd'
+    const projectDir = 'future-apps/payroll'
+    const provider = VersionedContractsProvider.fromBranch({ commit, projectDir })
```
sohkai: This is super, super cool!

shared/test-helpers/lib/VersionedContractsProvider.js
```diff
@@ -0,0 +1,83 @@
+const fs = require('fs')
+const path = require('path')
+const { spawnSync } = require('child_process')
+
+const APPS_GIT_URL = 'https://github.com/aragon/aragon-apps.git'
+
+module.exports = (web3, TruffleContract, TruffleConfig) => {
+  return class VersionedContractsProvider {
+    static fromBranch({ commit, source = APPS_GIT_URL, projectDir = '.', outputDir = 'tmp' }) {
```
sohkai: Wonder if we could be more generic than using "branch" terminology. I foresee us using tags most of the time, but I imagine there's a better abstraction for git references in general.

facuspagnuolo: I started it with a tag/branch abstraction, but I thought that maybe commit was more flexible. We can extend it to whatever we want actually :)

sohkai:
Closing for now; let's resurrect this (perhaps in [`contract-helpers`](https://github.com/aragon/contract-helpers)) when we are looking to do another contract upgrade.

PR #784, opened 2019-04-11
Update App Center copy

john-light:
- Finance app manifest + details
- Token Manager manifest + details
- Vault manifest + details
- Voting manifest + details


luisivan:
Won't get into 0.7, but will get into the immediate release after it, which should happen in a few days https://app.zenhub.com/workspaces/aragon-one-5c77a714eb692d24c2a8153e/issues/aragonone/product/56

sohkai:
@john-light This has been released on staging (e.g. https://aragon.staging.aragonpm.com/#/test.aragonid.eth/apps?p=installed.voting.aragonpm.eth).

john-light:
We might want to re-think the two-column layout:

Screenshot_2019-04-26 Aragon

bpierre:
Apart from the content size issue, it shouldn’t break after titles, that’s weird :thinking:  /cc @AquiGorka

sohkai:
@john-light We will move back to a one-column layout for the description in https://github.com/aragon/aragon/pull/750.

PR #786, opened 2019-04-12
Shared: Use time helpers mock in all apps

facuspagnuolo:


facuspagnuolo:
@sohkai I improved some of the tests to use `increasteTime` instead of mocking it every time. Feel free to take another look at it.

sohkai:
@facuspagnuolo Looks like the instrumentation during coverage fails due to the new mocks :/. Wonder if we have to add it to the skip list or etc.

facuspagnuolo:
@sohkai solcover is actually failing due to a compilation error that doesn't make sense, it is failing due to the functions being overwritten in the time helpers mock with the following error message:
```
TypeError: Overriding function changes state mutability from "nonpayable" to "view"
```

sohkai:
@facuspagnuolo That's why I think it's due to some instrumentation issue... usually that's the case if we have a `view` function declared in a contract whose instrumentation gets skipped but another contract that does not get skipped overrides it (since solidity-coverage strips away the `view`).

Let's not worry about this too much for now and look into it a bit later :).

PR #787, opened 2019-04-12
Payroll: Use new finance implementation

facuspagnuolo:
Updating Payroll contract to use new Finance interface


sohkai:
Ahh @facuspagnuolo I forgot to merge https://github.com/aragon/aragon-apps/pull/741 after it finished CI. It has this change :).

PR #795, opened 2019-04-15
Copy improvements

luisivan:
Mainly trying to clarify things for users and remove technical terms like *quorum* or *execution*


sohkai:
@luisivan This branch was created from quite an old commit (end of Feb); could you do a `git merge`, fix the conflicts, and push again?

apps/voting/app/src/components/VotePanelContent.js
```diff
@@ -187,7 +187,7 @@ class VotePanelContent extends React.Component {
           </div>
           <div>
             <h2>
-              <Label>Quorum progress</Label>
+              <Label>Total yes votes</Label>
```
john-light: I don't think `Total yes votes` really captures what is being shown (which is a % anyways)

Screenshot_2019-04-15 Aragon

Users _have_ to know what min. quorum is to have intelligent conversations about a vote e.g. 
Alice: "Why didn't this vote pass even though it has more yes than no votes?" 
Bob: "The minimum quorum threshold wasn't reached." 

So I would actually change it to `Min. Quorum progress` since this is how we refer to it in onboarding and the user guide.

apps/voting/app/src/components/VoteSuccess.js
```diff
@@ -19,7 +19,8 @@ const VoteSuccess = ({ settings, vote, ...props }) => (
             <Icon />
           </IconWrapper>
           <StatusLabel>
-            {success ? 'Vote will pass' : 'Vote won’t pass'}
+            Current outcome:
+            {success ? ' Yes' : ' No'}
```
john-light: I don't think a simple "Yes" or "No" is going to tell a user what's going on here:

Screenshot_2019-04-15 Aragon(1)

apps/voting/app/src/components/VoteSummary.js
```diff
@@ -21,7 +21,7 @@ const VoteSummary = ({ vote, tokenSymbol, tokenDecimals, ready }) => {
       <Header>
         <span>
           <Text color={theme.textSecondary} smallcaps>
-            Current support{' '}
+            Votes{' '}
```
john-light: Similar to the "Min. Quorum" comment above, I don't think "Votes" really captures what is being shown here. Yes, it shows how many votes there have been and which way they have voted. But it also says "(xx% needed)" and `Votes (xx% needed)` doesn't really tell me what I need to know. If you want this to say "Votes" then it should be followed by `(xx% support needed for approval)` or just remove this section in parentheses altogether (in which case users will have to look through the Permissions app to find the support level required on this voting app).

apps/voting/app/src/components/VoteSuccess.js
luisivan: I think Vote won't pass or Vote will pass infers that the vote will take one course or another, without explicitly saying that it's the current state, and not final

apps/voting/app/src/components/VotePanelContent.js
luisivan: I think quorum is a very technical term, and even I get really confused by it

apps/voting/app/src/components/VoteSummary.js
luisivan: Votes (xx% support needed for approval) sounds good to me

apps/voting/app/src/components/VotePanelContent.js
john-light: @luisivan then it could be explained, maybe with a tooltip next to the term. This is how it is explained in the User Guide:

> “Min. Quorum” is short for “Minimum Acceptance Quorum". Min. Quorum is the number for **what percent of the total outstanding supply of tokens must approve a proposal for the vote to be considered valid**. For example, if the Min. Quorum is set to 20%, then 20% of the outstanding token supply must vote to approve a proposal for the vote to be considered valid. If a vote does not make quorum, then it will fail, even if more tokens voted to approve the proposal than voted against it. For example, if the Min. Quorum is set to 20% and 10% of the outstanding token supply votes against the proposal but only 15% vote in support, then the proposal will fail because it has not reached the Min. Quorum threshold.

apps/voting/app/src/components/VoteSuccess.js
john-light: I see, it is adding a "Current outcome" header followed by Yes or No. Ok.

apps/voting/app/src/components/VotePanelContent.js
dizzypaty: @luisivan @john-light Last week I had a chat with @lkngtn about Aragon's quorum and how it differs from conventional quorum with the aim of adding clarity to the product on-boarding. He suggested that to rename 'Quorum' as ‘Minimum approval %’ because it’s more accurate. I personally like it because it might beeasier to understand by people that are confused by the definition of quorum in general, and because it's more precise as to how we use it.

I agree with John's idea, we should also offer further details/explanation by using a info popover or tooltip and should be consistent with the ones presented on the product on-boarding redesign. 

cc @bpierre @sohkai

luisivan: Retracted this change for now, proceeding to merge and will come back to the *quorum* term in the future

PR #797, opened 2019-04-16
Payroll: Fix exchange rates request call

### LINKED ISSUE #792: Payroll: denomination token should be passed as the 'quote' to PPF
The denomination token should always be the `quote` and the `base` should be the token that we are asking the price for (see [`PPF.get`](https://github.com/aragon/ppf/blob/master/packages/ppf-contracts/contracts/PPF.sol#L95)). 

This is currently swapped in the Payroll contract so the rates will be inverted.

facuspagnuolo:
Fixes #792 

Tests won't pass until we have a new version of PPF released including  https://github.com/aragon/ppf/pull/19, https://github.com/aragon/ppf/pull/20, https://github.com/aragon/ppf/pull/21, https://github.com/aragon/ppf/pull/22, and https://github.com/aragon/ppf/pull/23 changes

### DIFF
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -690,9 +690,7 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
             return ONE;
         }
 
-        uint128 xrt;
-        uint64 when;
-        (xrt, when) = feed.get(denominationToken, _token);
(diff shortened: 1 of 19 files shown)


facuspagnuolo:
@sohkai after changing the way we were mocking PPF, many tests started failing since they were expecting different payout amounts due to the wrong rates, fixing the contract fixed those as well :)

sohkai:
Haha that sounds reasonable 👍.

PR #798, opened 2019-04-16
Payroll: Fix accrued salary calculation

facuspagnuolo:
Accrued salary was being capped to max int if it there was an overflow. I'm splitting the way we calculate owed salary in order to:
- Revert in case the employer tries to set a new salary when the owed salary amount could not be represented as accrued salary due to an overflow
- Rescue capping to max int in case we are computing a payroll payment to allow employees withdrawing at least a portion of their owed salary

The rest of the scenarios (reimbursements and bonuses) will never overflow since those additions are being calculated with `SafeMath`.

### DIFF
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -9,7 +9,6 @@ import "@aragon/os/contracts/common/IForwarder.sol";
 import "@aragon/os/contracts/lib/math/SafeMath.sol";
 import "@aragon/os/contracts/lib/math/SafeMath64.sol";
 import "@aragon/os/contracts/lib/math/SafeMath8.sol";
-import "@aragon/os/contracts/common/Uint256Helpers.sol";
 
 import "@aragon/ppf-contracts/contracts/IFeed.sol";
 import "@aragon/apps-finance/contracts/Finance.sol";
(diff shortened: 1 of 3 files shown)


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -649,8 +650,8 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
      * @param _employeeId Employee's identifier
      * @return Total amounts of previous and current owed salaries for the requested employee since their last payroll
      */
-    function _getOwedSalary(uint256 _employeeId) internal view returns (uint256 currentOwedSalary, uint256 totalOwedSalary) {
-        currentOwedSalary = _getCurrentOwedSalary(_employeeId);
+    function _getOwedSalaries(uint256 _employeeId) internal view returns (uint256 currentOwedSalary, uint256 totalOwedSalary) {
+        currentOwedSalary = _getCurrentCappedOwedSalary(_employeeId);
         totalOwedSalary = currentOwedSalary + employees[_employeeId].accruedSalary;
```
sohkai: Hmm, but wouldn't this be possible of overflowing?

facuspagnuolo: Yes, and we are performing a check to cap it again if that happens in the next line (see [L657](https://github.com/aragon/aragon-apps/pull/798/files#diff-9d7adaffbcce19867f027a983a30bcc7R657))

sohkai: 🙈 missed that in the diff!

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -608,7 +607,7 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
         }
 
         // This function is only called from _payday, where we make sure that _payedAmount is lower than or equal to the
-        // total owed amount, that is obtained from _getCurrentOwedSalary, which does exactly the opposite calculation:
+        // total owed amount, that is obtained from _getCurrentCappedOwedSalary, which does exactly the opposite calculation:
         // multiplying the employee's salary by an uint64 number of seconds. Therefore, timeDiff will always fit in 64.
         // Nevertheless, we are performing a sanity check at the end to ensure the computed last payroll timestamp
         // is not greater than the current timestamp.
```
sohkai: I just reread this comment, and I think the sanity check has been removed now (it must've been `lastPayroolDate.toUint64()`?

facuspagnuolo: Added a last commit improving this casting

PR #800, opened 2019-04-16
Agent: allow tests to be shared

sohkai:
Similar to #668, but for subclasses of Agent (e.g. some of Apiary's contracts). cc @osarrouy


osarrouy:
That will be perfect to test the pool app :) Thanks a lot!

PR #806, opened 2019-04-16
voting: fetch active account votes in script

2color:
## What
- Move active account vote fetching to worker script

## Why
- Previous, every update to votes from the app reducer would fetch voter state for all votes, 
- With this change, we only fetch all the votes when the active account changes and fetch individual votes' state when a `CastVote` eth event is processed


apps/voting/app/src/script.js
```diff
@@ -137,17 +165,39 @@ async function createStore(token, tokenSettings) {
  *                     *
  ***********************/
 
-async function castVote(state, { voteId }) {
-  // Let's just reload the entire vote again,
-  // cause do we really want more than one source of truth with a blockchain?
+async function updateConnectedAccount(state, { account }) {
+  return {
+    ...state,
+    connectedAccount: account,
```
sohkai: Let's keep this outside of the cache, by moving this state to be outside of `createStore()`

apps/voting/app/src/script.js
```diff
@@ -137,17 +165,39 @@ async function createStore(token, tokenSettings) {
...
+    connectedAccountVotes: await getAccountVotes({
+      connectedAccount: account,
+      votes: state.votes,
+    }),
+  }
+}
+
+async function castVote(state, { voteId, voter }) {
+  const { connectedAccount, connectedAccountVotes } = state
+  // get the voter to see if
```
sohkai:
```suggestion
  // If the connected account was the one who made the vote, update their voter status
```

apps/voting/app/src/script.js
```diff
@@ -169,12 +219,39 @@ async function startVote(state, { creator, metadata, voteId }) {
...
+        states[voteId] = voteType
+        return states
+      }, {})
+    )
+    .catch(console.error)
+
+  return connectedAccountVotes
+}
+
+async function getAccountSpecificVote({ connectedAccount = '', voteId = '' }) {
```
sohkai: What do you think about renaming this to `getVoterState()` to mimic the contract call?

apps/voting/app/src/script.js
```diff
@@ -169,12 +219,39 @@ async function startVote(state, { creator, metadata, voteId }) {
...
+  return app
+    .call('getVoterState', voteId, connectedAccount)
+    .toPromise()
+    .then(voteTypeFromContractEnum)
+    .then(voteType => ({ voteId, voteType }))
+    .catch(console.error)
+}
```
sohkai: Could we move both of these new functions to be in the `helpers` section, since they're not directly called on events?

apps/voting/app/src/script.js
```diff
@@ -169,12 +219,39 @@ async function startVote(state, { creator, metadata, voteId }) {
   }))
 }
 
+async function getAccountVotes({ connectedAccount = '', votes = [] }) {
```
sohkai: Probably don't need the default args for both of these functions, I think web3 would complain if we tried to use an empty address too.

2color:
Tested again after applying changes. Looking good

Issue #732, opened 2019-03-24
Voting: don't show "execute" steps if no execution script is included in the vote


sohkai:
We could avoid showing the notion of "execution" in the frontend if no execution script is included, to avoid users being confused that a non-binding "yes/no" vote "requires execution"

cc @aragon/design

dizzypaty:
Yes, totally agree. Let's remove the 'Execute vote' button. Also, I think we should change the status to 'Passed' or 'Rejected' rather than 'Executed'.

sohkai:
Closed with #808.

PR #808, opened 2019-04-17
Voting app: clarify polls

bpierre:
Fixes https://github.com/aragon/aragon-apps/issues/582 and https://github.com/aragon/aragon-apps/issues/732

all-votes
panel-1
panel-2

### DIFF
--- a/apps/voting/app/src/components/NewVotePanelContent.js
+++ b/apps/voting/app/src/components/NewVotePanelContent.js
@@ -76,7 +76,6 @@ class NewVotePanelContent extends React.Component {
         )}
         {screenIndex === 1 && (
           <div>
-            <p css="margin: 30px 0 20px">Coming soon!</p>
             <Info.Action title="These votes are binding">
               <div
                 css={`
--- a/apps/voting/app/src/components/NewVotePanelContent.js
+++ b/apps/voting/app/src/components/NewVotePanelContent.js
@@ -1,9 +1,10 @@
 import React from 'react'
 import styled from 'styled-components'
-import { Button, Info, Text, TextInput, Field } from '@aragon/ui'
+import { TabBar, Button, Info, Text, TextInput, Field } from '@aragon/ui'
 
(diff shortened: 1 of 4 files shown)


apps/voting/app/src/components/NewVotePanelContent.js
```diff
@@ -31,30 +35,62 @@ class NewVotePanelContent extends React.Component {
...
+            items={['Question', 'Action']}
+            selected={screenIndex}
+            onChange={this.handleTabChange}
+          />
+        </div>
+        {screenIndex === 0 && (
+          <Form onSubmit={this.handleSubmit}>
+            <Field label="Question">
+              <TextInput
+                ref={question => (this.questionInput = question)}
```
sohkai: It would be nice if we focused this input on navigating back to the tab.

apps/voting/app/src/components/NewVotePanelContent.js
```diff
@@ -31,30 +35,62 @@ class NewVotePanelContent extends React.Component {
...
+                  Questions are used for signaling and don’t have any direct
+                  repercussions on the organization.
+                </div>
+              </Info.Action>
+            </div>
+          </Form>
+        )}
+        {screenIndex === 1 && (
+          <div>
+            <p css="margin: 30px 0 20px">Coming soon!</p>
```
sohkai: Not sure if we decided on keeping this @dizzypaty

dizzypaty: Mmm... we can keep it if we plan to eventually make it available from here (https://github.com/aragon/aragon-apps/issues/582#issuecomment-443433985). Otherwise I'm fine with removing it! I know this tab still feels a bit weird and empty but I'll definitely rethink this on the next iteration of the voting app.

luisivan: Yeah let's remove the Coming Soon part

PR #812, opened 2019-04-17
Voting: fix bug with votes not loading initially

2color:


apps/voting/app/src/app-state-reducer.js
```diff
@@ -26,6 +32,8 @@ function appStateReducer(state) {
       tokenDecimals: tokenDecimalsNum,
     },
 
+    connectedAccountVotes: connectedAccountVotes || {},
```
2color: This is to ensure there's an object in https://github.com/aragon/aragon-apps/blob/4afc2ad92c5633eeb7545f716c46703ff42193ab/apps/voting/app/src/vote-hooks.js#L29-L32

apps/voting/app/src/script.js
```diff
@@ -222,7 +224,7 @@ async function startVote(state, { creator, metadata, voteId }) {
  *                     *
  ***********************/
 
-async function getAccountVotes({ connectedAccount, votes }) {
+async function getAccountVotes({ connectedAccount, votes = [] }) {
```
sohkai: Can we leave a comment for why we need to default this?

2color:
> LGTM, I think I was the one that asked if we could remove the default for votes D:.

A comment in the first place would have prevented that :)

PR #813, opened 2019-04-17
Finance: fix token fallbacks and error handling

sohkai:
Fixes https://github.com/aragon/aragon/issues/734.

With https://github.com/aragon/aragon.js/pull/277 we now get the actual errors back from the RPC when a call goes wrong (e.g. naive `token.symbol()` on DAI) and we need to handle these explicitly.

This PR converts a lot of the token fallback-related bits into simpler Promise-based code and adds explicit handlers for their error cases.

### DIFF
--- a/apps/finance/app/src/components/NewTransfer/Deposit.js
+++ b/apps/finance/app/src/components/NewTransfer/Deposit.js
@@ -149,7 +149,11 @@ class Deposit extends React.Component {
     const token = api.external(address, tokenAbi)
 
     return new Promise(async (resolve, reject) => {
-      const userBalance = await token.balanceOf(connectedAccount).toPromise()
+      const userBalance = await token
+        .balanceOf(connectedAccount)
+        .toPromise()
+        // Assume no balance if the call failed
+        .catch(() => '0')
 
       const decimalsFallback =
         tokenDataFallback(address, 'decimals', network.type) || '0'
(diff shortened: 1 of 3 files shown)


apps/finance/app/src/components/NewTransfer/Deposit.js
```diff
@@ -167,16 +168,20 @@ class Deposit extends React.Component {
       }
 
       const [tokenSymbol, tokenDecimals] = await Promise.all([
-        getTokenSymbol(api, address),
-        token.decimals().toPromise(),
+        getTokenSymbol(api, address).catch(() => {}),
```
bpierre: Super nitpicky but should we return explicit values (`null`) rather than implicit `undefined` in these cases?

sohkai: Yeah... we should probably settle on a convention for handling errors and their "safe" default values.

In the case of the balance check above, I opted for "0" since it'll show "no balances" but we could probably use "-1" as "couldn't find balance".

sohkai: I've now changed it to use `-1` for the balance if the balance call failed and default the symbol and name values to `''` (since they're optional strings for the token to implement).

apps/finance/app/src/components/NewTransfer/Deposit.js
```diff
@@ -39,6 +39,19 @@ const TOKEN_ALLOWANCE_WEBSITE = 'https://tokenallowance.io/'
 
 const tokenAbi = [].concat(tokenBalanceOfAbi, tokenDecimalsAbi, tokenSymbolAbi)
 
+const renderBalanceForSelectedToken = selectedToken => {
+  const { decimals, loading, symbol, userBalance } = selectedToken.data
+  if (loading || !userBalance) {
+    return ''
+  }
+
+  return userBalance === -1
```
bpierre: I think this should be `'-1'`?

sohkai: 😅 this is why I need a second pair of eyes 😄

PR #817, opened 2019-04-18
chore: remove unnecessary dependencies

sohkai:
cc @facuspagnuolo 

Hopefully this will get travis to run a bit faster with a lighter cache. Might be some other cache improvements we can make... somehow the cache on travis is yuge at 16GB:

Screen Shot 2019-04-18 at 9 22 53 AM

### DIFF
--- a/apps/agent/package.json
+++ b/apps/agent/package.json
@@ -36,12 +36,10 @@
     "@aragon/test-helpers": "^1.1.0",
     "eth-gas-reporter": "^0.1.1",
     "ethereumjs-util": "^6.1.0",
-    "solidity-coverage": "0.5.11",
     "solium": "^1.2.3",
     "truffle": "4.1.14",
     "truffle-extract": "^1.2.1",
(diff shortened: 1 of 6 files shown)


apps/agent/package.json
```diff
@@ -36,12 +36,10 @@
     "@aragon/test-helpers": "^1.1.0",
     "eth-gas-reporter": "^0.1.1",
     "ethereumjs-util": "^6.1.0",
-    "solidity-coverage": "0.5.11",
     "solium": "^1.2.3",
     "truffle": "4.1.14",
     "truffle-extract": "^1.2.1",
-    "web3-eth-abi": "^1.0.0-beta.38",
-    "webpack": "3.10.0"
```
sohkai: We used to have to pin webpack because the frontends relied on it and somehow there were version clashes with ganache (or something weird), but we've since moved to parcel and nothing should rely on webpack now.

Issue #809, opened 2019-04-17
Finance: Date range filter selection shouldn't be applied by default 


dizzypaty:
As a user, I want to be able to select a date range to export my Finance app transfers. 

* Date range should be deselected by default displaying the following hint text: ‘Start date’ and ‘End date’, until user makes a date selection.
* We should indicate today’s date and open the calendar on the current month 
* We should include an Apply and Clear buttons, that removes the current date range selection

dizzypaty:
## Large form factor

<img width="1366" alt="WebApp-1366px - Finance main view" src="https://user-images.githubusercontent.com/7205369/56952372-824efa80-6b3a-11e9-9927-1fa64507a95a.png">

<img width="1366" alt="WebApp-1366px - Finance main view-2" src="https://user-images.githubusercontent.com/7205369/56952402-94c93400-6b3a-11e9-9498-71258dad788a.png">


<img width="1366" alt="WebApp-1366px - Finance main view-1" src="https://user-images.githubusercontent.com/7205369/56952410-9b57ab80-6b3a-11e9-8fa3-10630213566e.png">


<img width="1366" alt="WebApp-1366px - Finance main view-3" src="https://user-images.githubusercontent.com/7205369/56952416-9dba0580-6b3a-11e9-9482-edeede545df9.png">

### Small form factor 
<img width="320" alt="Mobile" src="https://user-images.githubusercontent.com/7205369/56952424-a0b4f600-6b3a-11e9-9ec7-e5caa51e02c1.png">
<img width="320" alt="Mobile-1" src="https://user-images.githubusercontent.com/7205369/56952423-a01c5f80-6b3a-11e9-8d03-efa6153741c9.png">


[Figma files](https://www.figma.com/file/Ffxzu9CDt8KWluyYOVELUnNW/0.7-Usability-testing-issues?node-id=18%3A200)

AquiGorka:
@dizzypaty for mobile/reponsive do we show two sets of `clear`/`apply` buttons?

dizzypaty:
@AquiGorka 

> For mobile/responsive do we show two sets of clear/apply buttons? 

No, only one

> And two titles? with two calendars? Or is it one calendar? 

Just one calendar, full screen 

> If it is one calendar is there a reference to which date (start or end) is the user choosing?

[...]

PR #819, opened 2019-04-22
Finance: date filters remove default selection

AquiGorka:
These changes:

- Placeholder for dates: `--/--/----`
- When no range is selected today's date is used as placeholder for `endDate`
- When `startDate` is selected but no `endDate`, then `endDate` uses the placeholder
- Date range filter is applied when the user clicks outside of the date range picker instead of when clicking on dates as before
- Modified today's border to main color
- Uses monospaced font so that placeholder uses same char width as actual dates and uses `ch` units to define container's size

Refs: #809 and https://github.com/aragon/aragon-apps/pull/767#discussion_r274326591

### DIFF
--- a/apps/finance/app/src/components/DateRange/DateRangeInput.js
+++ b/apps/finance/app/src/components/DateRange/DateRangeInput.js
@@ -15,6 +15,8 @@ import IconCalendar from './Calendar'
 import TextInput from './TextInput'
 import DatePicker from './DatePicker'
 
+const DATE_PLACEHOLDER = '--/--/----'
+
 class DateRangeInput extends React.PureComponent {
   state = {
     showPicker: false,
@@ -31,7 +33,7 @@ class DateRangeInput extends React.PureComponent {
 
     return isDate(startDate)
       ? formatDate(startDate, this.props.format)
-      : '--/--/----'
+      : DATE_PLACEHOLDER
   }
 
   get formattedEndDate() {
@@ -41,7 +43,7 @@ class DateRangeInput extends React.PureComponent {
     return isDate(endDate)
       ? formatDate(endDate, format)
       : startDate
-      ? '--/--/----'
+      ? DATE_PLACEHOLDER
       : formatDate(new Date(), format)
   }
 
@@ -116,6 +118,7 @@ class DateRangeInput extends React.PureComponent {
           adornment={icon}
           adornmentPosition="end"
           height={39}
+          placeholder=""
         />
         {this.state.showPicker && (
           <StyledDatePickersContainer>
--- a/apps/finance/app/src/components/DateRange/DateRangeInput.js
+++ b/apps/finance/app/src/components/DateRange/DateRangeInput.js
@@ -158,7 +158,7 @@ const StyledContainer = styled.div`
 
 const StyledTextInput = styled(TextInput)`
   width: 28ch;
-  fontfamily: ${font({ monospace: true })};
+  ${font({ monospace: true })};
 `
 
 const StyledDatePickersContainer = styled.div`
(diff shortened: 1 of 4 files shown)


apps/finance/app/src/components/DateRange/DateRangeInput.js
```diff
@@ -153,7 +157,8 @@ const StyledContainer = styled.div`
 `
 
 const StyledTextInput = styled(TextInput)`
-  width: 13rem;
+  width: 28ch;
+  fontfamily: ${font({ monospace: true })};
```
bpierre:
```suggestion
  ${font({ monospace: true })};
```

apps/finance/app/src/components/Transfers.js
```diff
@@ -200,12 +199,14 @@ class Transfers extends React.PureComponent {
...
-              {formatDate(selectedDateRange.end)}) selection.{' '}
+              No transfers match your filter{' '}
+              {selectedDateRange.start &&
+                `and period (${formatDate(
+                  selectedDateRange.start
+                )} to ${formatDate(selectedDateRange.end)}) selection. `}
               {filtersActive && (
                 <a role="button" onClick={this.handleResetFilters}>
-                  Clear filters
+                  Clear filters
```
bpierre: I think we lost the ` ` (non-breaking space) here :-)

apps/finance/app/src/components/DateRange/DateRangeInput.js
```diff
@@ -29,13 +29,20 @@ class DateRangeInput extends React.PureComponent {
...
-    const { endDate } = this.state
-
-    return isDate(endDate) ? formatDate(endDate, this.props.format) : ''
+    const { endDate, startDate } = this.state
+    const { format } = this.props
+
+    return isDate(endDate)
+      ? formatDate(endDate, format)
+      : startDate
+      ? '--/--/----'
```
bpierre: I wonder if we could do something so that screen readers don’t hear “dash-dash-slash-dash-dash…”? :thinking: Probably not a great solution, but maybe using `placeholder` with an empty value could help some readers to know that the field is empty, making it easy to skip the placeholder?

AquiGorka: I added this dashes to keep the input like this:
image

Instead of:
image

Added the empty placeholder as suggested.
The thing is that if only one date is selected then the dashes fill in to keep char width. How do screen readers behave with empty spaces? (Do they read "space-space-space...")

bpierre: Ah yes, I forgot it was the same field for both… empty spaces shouldn’t be read no, do you see a solution using spaces?

Otherwise I think we can keep it how it was for now (everything in `value`), and, when we’ll move it to aragonUI, we’ll use an element on top of the input to ensure that it’s only seen as a visual thing.

AquiGorka: `const DATE_PLACEHOLDER = '          '`

image

AquiGorka: I do think the dashes are more helpful and agree we can figure out a solution when implementing in `aragonUi`

AquiGorka: @bpierre the last commit (https://github.com/aragon/aragon-apps/pull/819/commits/96b23408f2cd62dd9d26c5402e39bd506d8bed7b) uses placeholder instead of value when either value is missing and value when it has both. Let me know if this ok and if so, can I get your approve to merge?

bpierre: Sorry I forgot to approve, done!

AquiGorka:
> I think this might get potentially confusing, but let's see how it is with users

@sohkai hmm, yeah, but if it is not done this way then the input appears to have 2 dates selected (start and end) but the period filter is not really being applied.

Let's see how it plays out and I can always come back and iterate on it.

PR #822, opened 2019-04-25
feat(Payroll): small SLOAD optimizations to _transferTokensAmount()

sohkai:
This should avoid a few unnecessary `SLOAD`s that result to the same state.

PR #823, opened 2019-04-25
Payroll: clarify how rates are handled and add unit tests

facuspagnuolo:
Adding unit tests to ensure how rates handling in payroll (cc https://github.com/aragon/aragon-apps/pull/821). I also rolled back the last change we did to the PPF request to confirm it was working properly.


future-apps/payroll/test/helpers/set_token_rates.js
```diff
@@ -11,8 +11,8 @@ module.exports = web3 => {
   return async function setTokenRates(feed, denominationToken, tokens, rates, when = undefined) {
     if (!when) when = await feed.getTimestampPublic()
 
-    const bases = tokens.map(token => typeof(token) === 'object' ? token.address : token)
-    const quotes = tokens.map(() => typeof(denominationToken) === 'object' ? denominationToken.address : denominationToken)
+    const quotes = tokens.map(token => typeof(token) === 'object' ? token.address : token)
+    const bases = tokens.map(() => typeof(denominationToken) === 'object' ? denominationToken.address : denominationToken)
```
izqui: We shouldn't be inverting the base and quote here as when setting the rates in PPF the `denominationToken` (USD) will be the quote.

izqui: Nvm, I thought this was being used in the `Payroll_rates` test

future-apps/payroll/test/contracts/Payroll_rates.test.js
```diff
@@ -0,0 +1,142 @@
...
+        const previousETH = await web3.eth.getBalance(employee)
+        const previousDAI = await DAI.balanceOf(employee)
+        const previousANT = await ANT.balanceOf(employee)
+
+        const { tx, receipt } = await payroll.payday(PAYMENT_TYPES.PAYROLL, 0, { from: employee })
+        const { gasPrice } = await web3.eth.getTransaction(tx)
+        const txCost = gasPrice.mul(receipt.gasUsed)
+
+        const currentETH = await web3.eth.getBalance(employee)
+        assert.equal(currentETH.minus(previousETH).plus(txCost).toString(), 6, 'expected current ETH amount does not match')
```
sohkai: Here and elsewhere, let's comment how we got to the expected payout number (e.g. `120s * 1usd/s / 20eth/usd == 6eth`)

future-apps/payroll/test/contracts/Payroll_rates.test.js
```diff
@@ -0,0 +1,142 @@
...
+
+    USD = '0x0000000000000000000000000000000000000001'
+    ETH = '0x0000000000000000000000000000000000000000'
+    await finance.deposit(ETH, bn(50, 18), 'Initial ETH deposit', { from: owner, value: bn(50, 18) })
+
+    DAI = await deployErc20TokenAndDeposit(owner, finance, 'DAI', TOKEN_DECIMALS)
+    ANT = await deployErc20TokenAndDeposit(owner, finance, 'ANT', TOKEN_DECIMALS)
+  })
+
+  beforeEach('initialize payroll app with USD as denomination token', async () => {
```
sohkai: Let's cover our bases and also add some tests for when the Payroll app is initialized with ETH as its denomination token (as the next most probable use case outside of fiat/DAI denomination token).

For this case, let's also use the denomination as the base pricefeed token to test the "non-inverse" case where the org's denomination token matches the pricefeed's base token.

facuspagnuolo:
@izqui @sohkai as you can see in https://github.com/aragon/aragon-apps/pull/823/commits/a9f3a0701900cbbd9b72faed4be48740b5b71583, I changed transfer amount [formula](https://github.com/aragon/aragon-apps/pull/823/commits/a9f3a0701900cbbd9b72faed4be48740b5b71583#diff-9d7adaffbcce19867f027a983a30bcc7R697) we were using in the contract. I think the previous one was wrong, and that's why using an inverse rate was working fine. The amount to be transferred must be in the allowed token units [AT], the owed amount is in denomination token units [DT], and the rate is fetched using the denomination token as the quote which is in [DT/AT], then we should divide the owed amount by the rate instead of multiplying it.

Besides that, I did some more tweaks to change all the test suite to use a real denomination token (USD) with real salaries amount.

future-apps/payroll/test/contracts/Payroll_rates.test.js
```diff
@@ -38,12 +37,338 @@ contract('Payroll rates handling,', ([owner, employee]) => {
...
+  })
+
+  context('when the denomination token is ETH', async () => {
+    beforeEach('initialize payroll app', async () => {
+      ({ payroll, priceFeed } = await createPayrollAndPriceFeed(dao, payrollBase, owner, NOW))
+      await payroll.initialize(finance.address, ETH, priceFeed.address, RATE_EXPIRATION_TIME, { from: owner })
+    })
+
+    beforeEach('set rates and allow tokens', async () => {
+      await setTokenRate(priceFeed, DAI, ETH, DAI_RATE.div(ETH_RATE)) // 0.050 ETH
```
sohkai: What do you think about keeping the base token in the price feed as `ETH`? I think this would be a better test since we've already covered the "inverse" case above with USD being the quote token in the price feed.

future-apps/payroll/test/contracts/Payroll_gas_costs.test.js
```diff
@@ -1,74 +1,59 @@
...
+        await payroll.determineAllocation([DAI.address], [100], { from: employee })
         const { receipt: { cumulativeGasUsed: employeePayoutGasUsed } } = await payroll.payday(PAYMENT_TYPES.PAYROLL, 0, { from: employee })
 
-        await payroll.determineAllocation([denominationToken.address, erc20Token1.address, erc20Token2.address], [65, 25, 10], { from: anotherEmployee })
+        await payroll.determineAllocation([DAI.address, ANT.address], [60, 40], { from: anotherEmployee })
         const { receipt: { cumulativeGasUsed: anotherEmployeePayoutGasUsed } } = await payroll.payday(PAYMENT_TYPES.PAYROLL, 0, { from: anotherEmployee })
 
         const gasPerAllowedToken = anotherEmployeePayoutGasUsed - employeePayoutGasUsed
-        assert.isBelow(gasPerAllowedToken, 280000, 'payout gas cost increment per allowed token should be ~270k')
+        assert.isBelow(gasPerAllowedToken, 295000, 'payout gas cost increment per allowed token should be ~295k')
```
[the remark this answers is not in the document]
sohkai: Ah no, just ran these tests with and without #822 and it looks like it's the formula change.

facuspagnuolo:
> It's subtle, but the first case with multiplication is a lot more clear to me.
It is and it was to me, but I thought we were also really sure that the PPF was always being used using the denomination token as quote.

future-apps/payroll/test/contracts/Payroll_gas_costs.test.js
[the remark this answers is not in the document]
facuspagnuolo: yea, I think so... it actually increased to 289, but I wanted to extend the threshold a bit to avoid changing these tests to much

future-apps/payroll/test/contracts/Payroll_rates.test.js
facuspagnuolo: sure, I think it make sense :)

PR #825, opened 2019-04-26
Fix voting displayed percent

AquiGorka:
Ref: fixes https://github.com/aragon/aragon/issues/742

The error came from [this line](https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/math-utils.js#L95): the check was not `break`ing because `(0.55 * 100 * 100) % 1  = 9.094947017729282e-13` (even though `(0.55 * (100 * 100)) % 1 = 0`).

Applied the function that is being used in the survey app and added local tests for voting.
As discussed with @bpierre we need to add a `shared/common` library for all the apps.

### DIFF
--- a/apps/voting/app/src/math-utils.js
+++ b/apps/voting/app/src/math-utils.js
@@ -1,3 +1,5 @@
+import BN from 'bn.js'
+
 /**
  * Format numbers for a given number of decimal places
  *
@@ -37,7 +39,7 @@ export function formatNumber(num, decimals = 2, { truncate = true } = {}) {
 }
 
 export function percentageList(values, digits = 0) {
-  return scaleValuesSet(values, digits)
+  return scaleBNValuesSet(values).map(value => value.toNumber())
 }
 
 /**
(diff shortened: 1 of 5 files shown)


apps/voting/app/src/math-utils.js
```diff
@@ -61,44 +63,99 @@ export function safeDiv(num, denom) {
...
-// will be exactly `total`.
-export function scaleValuesSet(values, digits = 0, total = 100) {
-  const digitsMultiplicator = Math.pow(10, digits)
+// Scale to `total` a set of values summing to 1.
+// Note that the accuracy of `values` is constrained to a set amount of decimals, as BN.js doesn't
+// support decimal operations
+export function scaleBNValuesSet(
+  values = [],
+  total = new BN(100),
+  correctionLimit = 0.001
```
sohkai: What do you think about keeping the inputs and outputs to a single number type (normal JS numbers)? Mixing them up is invariably going to lead to a really bad time at some point or another, and for the purposes of this function, I think it's enough to completely abstract away the BN usage.

AquiGorka: Makes sense, there is a tradeoff:

image

sohkai: Hmm, just checked what we need this for and indeed sometimes we do use this to scale balances (e.g. creating a distribution over multiple vote options), so we do need to expose `total` as a BN.

Let's backtrack a bit... perhaps we can do the `.toNumber()` in `percentageList()` (since that should never use a different `total` than 100)?

AquiGorka: Better approach yes, thanks for the suggestion. Done.

PR #829, opened 2019-04-28
Voting: skip unnecessary renderings

bpierre:
Noticeable improvement: the side panels transition should be smoother.

Changes:

- Move panels in their own components, so that the rendering of   SidePanel can be skipped entirely.
- Memoize the object returned by usePanel(), so that it can be compared   using a strict equality.
- Move the voting panel actions in their own component fetching their   data, so that the panel component doesn’t need to render that much   anymore.
- Don’t recreate `votes` every second, instead only do it if any   “opened” status has changed.

### DIFF
--- a/apps/voting/app/src/components/NewVotePanel.js
+++ b/apps/voting/app/src/components/NewVotePanel.js
@@ -1,13 +1,27 @@
 import React from 'react'
 import styled from 'styled-components'
-import { TabBar, Button, Info, Text, TextInput, Field } from '@aragon/ui'
+import { Button, Field, Info, SidePanel, TabBar, TextInput } from '@aragon/ui'
 
 const initialState = {
   question: '',
   screenIndex: 0,
 }
 
-class NewVotePanelContent extends React.Component {
+const NewVotePanel = React.memo(({ panelState, onCreateVote }) => (
+  <SidePanel
+    title="New Vote"
+    opened={panelState.visible}
+    onClose={panelState.requestClose}
+    onTransitionEnd={panelState.onTransitionEnd}
+  >
+    <NewVotePanelContent
+      onCreateVote={onCreateVote}
+      panelOpened={panelState.didOpen}
+    />
+  </SidePanel>
+))
+
+class NewVotePanelContent extends React.PureComponent {
   static defaultProps = {
     onCreateVote: () => {},
   }
@@ -99,4 +113,4 @@ const Form = styled.form`
   margin-top: 20px;
 `
 
-export default NewVotePanelContent
+export default NewVotePanel
--- a/apps/voting/app/src/App.js
+++ b/apps/voting/app/src/App.js
@@ -1,10 +1,10 @@
 import React from 'react'
-import { Main, SidePanel } from '@aragon/ui'
+import { Main } from '@aragon/ui'
 
 import EmptyState from './screens/EmptyState'
 import Votes from './screens/Votes'
-import VotePanelContent from './components/VotePanelContent'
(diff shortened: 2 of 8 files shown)


apps/voting/app/src/components/NewVotePanel.js
```diff
@@ -1,13 +1,27 @@
...
 const initialState = {
   question: '',
   screenIndex: 0,
 }
 
-class NewVotePanelContent extends React.Component {
+const NewVotePanel = React.memo(({ panelState, onCreateVote }) => (
+  <SidePanel
+    title="New Vote"
+    opened={panelState.visible}
```
AquiGorka: I followed `panelState` object to see if it was reliable for the `Object.is` comparison in `React.memo` - there is a lot of _memoization_ going upwards. A lot. I am glad everything has been optimized in such a way. Is it sustainable? What happens if a non-memoized object is added through the pipes? I'll keep this in mind to correctly implement new code.

bpierre: > Is it sustainable?

It’s still an experiment and I we should keep playing with this codebase to have an answer: I think it’s still a bit early to know if the things we don’t like are due to our lack of experience with hooks, or if they really are too limiting for some things (= big, stateful components like `App`).

> What happens if a non-memoized object is added through the pipes?

In the worst case, components would re-render too much, which should “only” impact performances until we optimize again.

If you take the `votes`, the entire array used to be recreated every second, because a key on every vote could *potentially* change (the vote being `open`ed or not). [It has now been updated](https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/components/VotingCard/VotingCard.js#L108), so that the `votes` array only gets recreated when the `open` status of one of the votes gets updated.

But it’s still imperfect: we should only recreate the array and the vote objects that need to be updated, not the others. That way, we could compare `prevProps.vote === nextProps.vote` (or `React.memo` / `React.PureComponent`) on the component directly.

In the meantime, and because it’s a critical component, I added this at the component level: https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/components/VotingCard/VotingCard.js#L108-L123

Another solution could be to make these all separate props, but I think it makes more sense to have a standard `vote` object in the app that can be passed directly to components, and ensuring that objects don’t get recreated if it’s not needed (à la Redux reducers).

bpierre:
@sohkai nice catch, thanks! :goal_net:

apps/voting/app/src/components/NewVotePanel.js
sohkai: > But it’s still imperfect: we should only recreate the array and the vote objects that need to be updated, not the others.

This is something that might be hard to do with how the background scripts work (and the caching involved). I think we get new objects back on every emission because it has to go through multiple frames (webworker -> client -> app). I guess eventually we'd use some sort of "change delta" API instead so you could declare in your reducer only the things that have changed.

bpierre: Ah yes this is another optimization we could have, but for now I was really talking about the ticker every second: at the moment, even if we don’t receive any updates from the worker, every vote object get updated if only one of them really need to be updated.

I was tempted to use a `WeakMap` to keep track of the previous vote object and reuse it if there is no update, but I preferred not to do it for now to not add even more complexity… maybe a more React-y solution would be to store them in a state, so we can access the previous vote objects and reuse them if needed.

Note: I realize I linked to the component skip function twice in the comment above, the first link was supposed to be this one: https://github.com/aragon/aragon-apps/pull/829/files#diff-604d07e63d04d87d1abce3ed54b927fdR18

PR #832, opened 2019-04-30
Voting: add block time to data fetching script

sohkai:


2color:
What is that script used for?

sohkai:
@2color It's helps us fetch some data for a particular vote (e.g. votes in ANV2) to analyze it (e.g in Tableau).

PR #834, opened 2019-05-04
Payroll: suggestions from review

sohkai:
Easiest to review commit by commit, mostly starting from less controversial to more controversial changes.

Also see inline annotations.


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -557,8 +559,7 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
      */
     function _setRateExpiryTime(uint64 _time) internal {
         // Require a sane minimum for the rate expiry time
-        // (1 min == ~4 block window to mine both a pricefeed update and a payout)
-        require(_time > 1 minutes, ERROR_EXPIRY_TIME_TOO_SHORT);
+        require(_time >= MIN_RATE_EXPIRY, ERROR_EXPIRY_TIME_TOO_SHORT);
```
sohkai: It seems more clear that "minimum" would be loose (you can choose a value greater than or equal to the minimum). This is also the definition we use for "minimum" for the Finance app's periods.

sohkai: It seems more clear that "minimum" would be loose (you can choose a value greater than or equal to the minimum). This is also the definition we use for "minimum" for the Finance app's periods.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -384,8 +384,8 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
      * @return True if the given address is an active employee, false otherwise
      */
     function canForward(address _sender, bytes) public view returns (bool) {
-        // Check employee exists (and matches)
-        return (employees[employeeIds[_sender]].accountAddress == _sender);
+        // Check sender is active employee
+        return _isEmployeeActive(employeeIds[_sender]);
```
sohkai: Even though terminated employees would eventually be removed, and therefore fail the previous check, we should also disallow employees who have been terminated but may be lingering as they are technically not part of payroll anymore.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -82,15 +82,15 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
     event AddEmployeeReimbursement(uint256 indexed employeeId, uint256 amount);
     event AddEmployeeAccruedSalary(uint256 indexed employeeId, uint256 amount);
-    event TerminateEmployee(uint256 indexed employeeId, address indexed accountAddress, uint64 endDate);
-    event ChangeAddressByEmployee(uint256 indexed employeeId, address indexed oldAddress, address indexed newAddress);
-    event DetermineAllocation(uint256 indexed employeeId, address indexed employee);
-    event SendPayment(address indexed employee, address indexed token, uint256 amount, uint128 exchangeRate, string paymentReference);
+    event TerminateEmployee(uint256 indexed employeeId, uint64 endDate);
+    event ChangeAddressByEmployee(uint256 indexed employeeId, address indexed newAccountAddress, address indexed oldAccountAddress);
+    event DetermineAllocation(uint256 indexed employeeId);
+    event SendPayment(uint256 indexed employeeId, address indexed employeeAccountAddress, address indexed token, uint256 amount, uint128 exchangeRate, string paymentReference);
```
sohkai: Generally it seems unnecessary to emit the employee's address as it's retrievable through `employeeId`.

However, I've kept it in `SendPayment` to allow us to search for payroll payments to particular addresses in the future.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -348,7 +348,7 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
         employeeIds[_newAddress] = employeeId;
         employee.accountAddress = _newAddress;
 
-        emit ChangeAddressByEmployee(employeeId, oldAddress, _newAddress);
+        emit ChangeAddressByEmployee(employeeId, _newAddress, oldAddress);
```
sohkai: I've flipped the ordering to be more similar to the order for roles (new value, old value)

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -737,18 +737,13 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
     function _removeEmployeeIfTerminatedAndPaidOut(uint256 _employeeId) internal {
         Employee storage employee = employees[_employeeId];
 
-        if (employee.endDate > getTimestamp64()) {
-            return;
-        }
-        if (_getCurrentCappedOwedSalary(_employeeId) > 0) {
-            return;
+        if (
+            employee.lastPayroll == employee.endDate &&
```
sohkai: Note that this change requires 3eab68a, to ensure that `lastPayroll` is set to `endDate` when drawing out all funds from a terminated employee.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -793,21 +789,24 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
-            employee.lastPayroll = (remainder == _currentOwedSalary) ? getTimestamp64() : _getLastPayroll(_employeeId, remainder);
-        } else {
-            // employee is only cashing out some current owed salary, and there is no previous owed salary,
-            // then we only need to update the last payroll date
-            employee.lastPayroll = (_paymentAmount == _currentOwedSalary) ? getTimestamp64() : _getLastPayroll(_employeeId, _paymentAmount);
+            // No need to use SafeMath here as we already know _paymentAmount > accruedSalary
+            currentSalaryPaid = _paymentAmount - accruedSalary;
         }
+        employee.lastPayroll = _getLastPayrollDate(_employeeId, currentSalaryPaid);
     }
```
sohkai: I found this reorganizing of logic a bit more clear than the previous version, since in the last two cases we want to update the employee's `lastPayroll`.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -793,21 +789,24 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
-            uint256 remainder = _paymentAmount - accruedSalary;
-            employee.lastPayroll = (remainder == _currentOwedSalary) ? getTimestamp64() : _getLastPayroll(_employeeId, remainder);
-        } else {
-            // employee is only cashing out some current owed salary, and there is no previous owed salary,
-            // then we only need to update the last payroll date
-            employee.lastPayroll = (_paymentAmount == _currentOwedSalary) ? getTimestamp64() : _getLastPayroll(_employeeId, _paymentAmount);
+            // No need to use SafeMath here as we already know _paymentAmount > accruedSalary
+            currentSalaryPaid = _paymentAmount - accruedSalary;
         }
+        employee.lastPayroll = _getLastPayrollDate(_employeeId, currentSalaryPaid);
```
sohkai: ⚠️ I've opted to remove the `_paymentAmount == _currentOwedSalary` check and just use `_getLastPayrollDate()` directly. It's a little bit more expensive in the case where you're drawing out all of your currently owed salary, but reduces the complexity.

In particular, I think it's much easier to reason about how an employee's `lastPayroll` value changes if it's only affected by how much currently owed salary is paid (through `_getLastPayrollDate()`).

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -600,11 +600,13 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
             timeDiff = timeDiff.add(1);
         }
 
-        uint64 lastPayrollDate = employee.lastPayroll.add(uint64(timeDiff));
```
sohkai: I was a little bit worried about this `uint64(timeDiff)` cast; although it's unlikely, at some point we may accidentally provide a `_paidAmount` that would cause `timeDiff` to be larger than MAX_UINT64.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -600,11 +600,13 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+        uint256 lastPayrollDate = uint256(employee.lastPayroll).add(timeDiff);
 
-        // Do a few sanity checks at the end to make sure the last payroll timestamp is clamped to our expected values
-        require(lastPayrollDate <= getTimestamp64(), ERROR_LAST_PAYROLL_DATE_TOO_BIG);
-        return lastPayrollDate > employee.endDate ? employee.endDate : lastPayrollDate;
+        // Even though this function should never receive a _paidAmount value that would result in
+        // the lastPayrollDate being higher than the current time, let's double check to be safe
+        require(lastPayrollDate <= uint256(getTimestamp64()), ERROR_LAST_PAYROLL_DATE_TOO_BIG);
+        // Already know lastPayrollDate must fit in uint64 from above
+        return uint64(lastPayrollDate);
```
sohkai: I took out the clamp to `endDate` as I'm pretty sure the `_paidAmount` should match `endDate` in circumstances where we're only paying up to an employee's `endDate`, but perhaps we may want a looser assumption.

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: I'm thinking of the other actions employees are allowed to perform even when they are terminated, like claiming their salaries. Could this still be useful in those cases when the calling employee is terminated?

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: I see and I like the idea

sohkai:
@facuspagnuolo I ended up reverting https://github.com/aragon/aragon-apps/pull/834/commits/7b43299c3d9e657b58033c16146f6f2d7ddef817 in https://github.com/aragon/aragon-apps/pull/834/commits/ce6359792c37bf77ca166c7b8742df4786b887fa because we were using the term "price feed" elsewhere, and `PriceFeed` is not the name of an actual contract (in Payroll's eyes).

Added a last few refactors / changes:

- https://github.com/aragon/aragon-apps/pull/834/commits/1b79e37195e86ef8eb9c5cc69363fcf357839f2f: I think inlining this bit is more clear as to how we use the capped value safely
- https://github.com/aragon/aragon-apps/pull/834/commits/1a471c00ec2b4e7392df5bdfd792fcf87733d23d: Pulls out some duplicated logic for setting the employee's address into a consolidated internal function

And added a few last touches for clarity / consistency:

- https://github.com/aragon/aragon-apps/pull/834/commits/bf5ce8b82c9be95073a499f028dc1fcfb42344f9: re-orders `role` to be the final parameter, as we generally leave the string metadata argument last in the other apps
- https://github.com/aragon/aragon-apps/pull/834/commits/094116b1a8684d4f5f31d4a4a4c198e33a036567: re-order `accruedSalary` to be closer to `denominationSalary` to group them together
- https://github.com/aragon/aragon-apps/pull/834/commits/d8dd5b1e7b01a2d70230a366a577fe956f5c9050: A little more explicit string for the Finance payment reference (note that these lengths are all still below 32 chars so there's no gas change :joy:)
- Rest: shuffling around some of the functions to logically group them by function (no hard rule, but generally: external -> public -> internal -> private, then state-changing -> view -> pure)

sohkai:
@facuspagnuolo I ended up reverting https://github.com/aragon/aragon-apps/pull/834/commits/7b43299c3d9e657b58033c16146f6f2d7ddef817 in https://github.com/aragon/aragon-apps/pull/834/commits/ce6359792c37bf77ca166c7b8742df4786b887fa because we were using the term "price feed" elsewhere, and `PriceFeed` is not the name of an actual contract (in Payroll's eyes).

Added a last few refactors / changes:

- https://github.com/aragon/aragon-apps/pull/834/commits/1b0e5f446dddb4a071afee9ae2add789b036982a: I think inlining this bit is more clear as to how we use the capped value safely
- https://github.com/aragon/aragon-apps/pull/834/commits/7bc705f1ca7f635655e08ac15dfed3e81283e39a: Pulls out some duplicated logic for setting the employee's address into a consolidated internal function

And added a few last touches for clarity / consistency:

- https://github.com/aragon/aragon-apps/pull/834/commits/bf5ce8b82c9be95073a499f028dc1fcfb42344f9: re-orders `role` to be the final parameter, as we generally leave the string metadata argument last in the other apps
- https://github.com/aragon/aragon-apps/pull/834/commits/094116b1a8684d4f5f31d4a4a4c198e33a036567: re-order `accruedSalary` to be closer to `denominationSalary` to group them together
- https://github.com/aragon/aragon-apps/pull/834/commits/d8dd5b1e7b01a2d70230a366a577fe956f5c9050: A little more explicit string for the Finance payment reference (note that these lengths are all still below 32 chars so there's no gas change :joy:)
- Rest: shuffling around some of the functions to logically group them by function (no hard rule, but generally: external -> public -> internal -> private, then state-changing -> view -> pure)

PR #835, opened 2019-05-06
Export labels with Finance export

AquiGorka:
How to test:

```sh
# one terminal 
cd into `aragon/aragon`
git fetch
git rebase origin/master
npm i
REACT_APP_ASSET_BRIDGE=local npm run start

# another terminal
cd into `aragon/aragon-apps`
git fetch
git checkout feature/export-labels
npm i
npm run start
```

- Open up browser to `localhost:3000`
- Click on `Finance` and test out export button setting and removing local identities
- Profit


apps/finance/app/src/components/Transfers.js
```diff
@@ -29,236 +30,220 @@ const TRANSFER_TYPES = [
...
   details[toChecksumAddress(address)] = {
     decimals,
     symbol,
   }
   return details
 }
+// Filter transfer based on the selected filters
+const getFilteredTransfers = ({
+  tokens,
+  transactions,
```
sohkai: Since we're moving this out, perhaps we can also modify its arguments.

What do you think about this signature instead?

```js
Transaction = { ... } // current Transaction object
Token = { ... } // current Token object
Range = {
  start: Date,
  end: Date,
}

const getFilteredTransfers = ({
  transactions: Array<Transaction>,
  selectedToken: Token || null,
  selectedTransferType: oneOf(TRANSFER_TYPES),
  selectedDateRange: Range
})
```

apps/finance/app/src/components/Transfers.js
```diff
@@ -29,236 +30,220 @@ const TRANSFER_TYPES = [
 ]
 const TRANSFERS_PER_PAGE = 10
```
sohkai: Perhaps we could rename this to `INITIAL_TRANSFERS_PER_PAGE`

apps/finance/app/src/components/Transfers.js
```diff
@@ -29,236 +30,220 @@ const TRANSFER_TYPES = [
...
+    setSelectedTransferType(index)
+    setDisplayedTransfers(TRANSFERS_PER_PAGE)
+  })
+  const handleResetFilters = React.useCallback(() => {
+    setDisplayedTransfers(TRANSFERS_PER_PAGE)
+    setSelectedToken(0)
+    setSelectedTransferType(0)
+  })
+  const getCsvFilename = React.useCallback(() => {
+    let filename = 'transfers.csv'
```
sohkai: What do you think about adding the current export date to the default filename?

apps/finance/app/src/components/Transfers.js
```diff
@@ -29,236 +30,220 @@ const TRANSFER_TYPES = [
...
+    const { start, end } = selectedDateRange
+    if (start && end) {
+      const formattedStart = format(selectedDateRange.start, 'yyyy-MM-dd')
+      const formattedEnd = format(selectedDateRange.end, 'yyyy-MM-dd')
+      filename = `transfers_${formattedStart}_to_${formattedEnd}.csv`
```
sohkai: Perhaps this is a good opportunity to add the `finance(<proxy address>)_` prefix, to make these files unique 😄

apps/finance/app/src/components/Transfers.js
```diff
@@ -29,236 +30,220 @@ const TRANSFER_TYPES = [
...
-            end: endOfDay(selectedDateRange.end),
-          })) &&
-        (selectedToken === 0 ||
-          addressesEqual(token, tokens[selectedToken - 1].address)) &&
-        (transferType === TransferTypes.All ||
-          (transferType === TransferTypes.Incoming && isIncoming) ||
-          (transferType === TransferTypes.Outgoing && !isIncoming))
+  })
+  const symbols = tokens.map(({ symbol }) => symbol)
+  const tokenDetails = tokens.reduce(reduceTokenDetails, {})
```
sohkai: Perhaps also a good chance to move this bit entirely, so we replace `reduceTokenDetails()` with `getTokenDetails()`

apps/finance/app/src/components/Transfers.js
```diff
@@ -29,236 +30,220 @@ const TRANSFER_TYPES = [
...
+  const filtersActive = selectedToken !== 0 || selectedTransferType !== 0
+  const { resolve } = React.useContext(IdentityContext)
```
sohkai: What do you think about renaming `resolve` to `resolveAddress` or etc? Looks really similar to a `Promise.resolve()`, and we even use it in the context of a Promise later.

apps/finance/app/src/components/Transfers.js
```diff
@@ -29,236 +30,220 @@ const TRANSFER_TYPES = [
...
+          tokenFilter={selectedToken}
+          onTokenChange={handleTokenChange}
+          transferTypeFilter={selectedTransferType}
+          onTransferTypeChange={handleTransferTypeChange}
+          compactMode={compactMode}
+          opened={filtersOpened}
+          symbols={symbols}
+          transferTypes={TRANSFER_TYPES_STRING}
+          downloadFileName={getCsvFilename()}
+          onDownloadData={getDownloadData}
```
sohkai: Perhaps we could just hoist all of the download logic (name + data) into this component on an `handleDownloadTransactions()` callback?

apps/finance/app/src/components/Transfers.js
```diff
@@ -29,236 +30,220 @@ const TRANSFER_TYPES = [
...
+          const formattedAmount = formatTokenAmount(
+            amount,
+            isIncoming,
+            decimals,
+            true,
+            { rounding: 5 }
+          )
+          const { name = null } = (await resolve(entity)) || {}
+          return `${formatDate(date)},${
+            name ? `${name} (${entity})` : entity
```
sohkai: Perhaps we should use a different column for the `name`, so the data import / manipulation is easier (much easier to annotate existing columns with a name when it's a separate column than for the tool to support text extraction from a field).

apps/finance/app/package.json
```diff
@@ -7,10 +7,11 @@
     "@aragon/api": "^1.0.0",
     "@aragon/api-react": "^1.0.0-beta.2",
     "@aragon/templates-tokens": "^1.2.0",
-    "@aragon/ui": "^0.34.0",
+    "@aragon/ui": "^0.37.0",
     "@babel/polyfill": "^7.0.0",
     "bn.js": "^4.11.8",
     "date-fns": "2.0.0-alpha.22",
+    "downloadjs": "^1.4.7",
```
sohkai: Looking at this dependency, is it mainly to smooth over some iOS incompatibilities (I think we were having some issues before)?

apps/finance/app/src/components/Transfers.js
AquiGorka: This is the filename when a date range has not been selected.

apps/finance/app/package.json
AquiGorka: I added this dep to be able to download the file dynamically. Before the dep it was being downloaded via the `download` + `href` props of a link - but this is rendered with the render context, if the user changes any custom label that does not trigger a re-render and a download will not have all the custom labels updated. With this dep, the file contents get created on demand to make sure all the custom labels exported are up to date.

apps/finance/app/src/components/Transfers.js
AquiGorka: Added this change as it makes sense to me but cc @luisivan to double check

apps/finance/app/src/components/Transfers.js
AquiGorka: Done 😄

apps/finance/app/src/components/Transfers.js
luisivan: Makes a lot of sense

apps/finance/app/src/components/Transfers.js
```diff
@@ -93,64 +136,28 @@ const Transfers = React.memo(({ tokens, transactions }) => {
...
-      'data:text/csv;charset=utf-8,Date,Name,Source/Recipient,Reference,Amount',
-    ]
-      .concat(mappedData)
-      .join('\n')
-
-    return window.encodeURI(csvContent)
-  }
+    const filename = getDownloadFilename(dao, selectedDateRange)
+    download(data, filename, 'data:text/csv')
+  })
```
sohkai: Should we add `filteredTransfers`, `tokenDetails`, and `resolveAddress` to this hook's cached list?

apps/finance/app/src/components/Transfers.js
```diff
@@ -55,15 +56,58 @@ const getFilteredTransfers = ({
...
+  return window.encodeURI(csvContent)
+}
+const getDownloadFilename = (dao, { start, end }) => {
+  const today = format(Date.now(), 'yyyy-MM-dd')
+  let filename = `finance_${dao}_${today}.csv`
+  if (start && end) {
+    const formattedStart = format(selectedDateRange.start, 'yyyy-MM-dd')
+    const formattedEnd = format(selectedDateRange.end, 'yyyy-MM-dd')
+    filename = `finance_${dao}_${formattedStart}_to_${formattedEnd}.csv`
+  }
```
sohkai: Nitpick: since we're changing this now, should we handle the case where you have a start date but no end date (so end defaults to today)?

apps/finance/app/src/components/Transfers.js
```diff
@@ -12,252 +13,248 @@ import {
...
+    setDisplayedTransfers(INITIAL_TRANSFERS_PER_PAGE)
+  })
+  const handleResetFilters = React.useCallback(() => {
+    setDisplayedTransfers(INITIAL_TRANSFERS_PER_PAGE)
+    setSelectedToken(0)
+    setSelectedTransferType(0)
+  })
+  const showMoreTransfers = React.useCallback(() => {
+    setDisplayedTransfers(displayedTransfers + INITIAL_TRANSFERS_PER_PAGE)
+  })
```
sohkai: This is probably starting to get annoying... but I just noticed these callbacks also need to have their dependencies listed :smile:

PR #845, opened 2019-05-13
Token manager - App state caching and init

2color:
## What

- Use update API to pass an async init function for loading state

## Related Issues/PRs
- Requires https://github.com/aragon/aragon.js/pull/297


apps/token-manager/app/src/script.js
```diff
@@ -47,67 +44,66 @@ retryEvery(retry => {
...
-      token.events(),
-    ]
+    {
+      externals: [{ contract: token }],
+      init: initState({ token, tokenAddress }),
+    }
   )
 }
 
+const initState = ({ token, tokenAddress }) => async cachedState => {
```
sohkai: Nitpick: we might as well use the normal function declaration since there's no need for the lexical binding, and all the other functions are declared using the function syntax.

apps/token-manager/app/src/script.js
```diff
@@ -47,67 +44,66 @@ retryEvery(retry => {
...
+  const maxAccountTokens = await app.call('maxAccountTokens').toPromise()
+
+  const inititalState = {
+    ...cachedState,
+    tokenAddress,
+    maxAccountTokens,
+    ...tokenSettings,
+  }
+
+  return inititalState
```
sohkai: We should leave a note here explaining why it's safe to not refresh the balances of all token holders (we'd process any event that could change their balances, even with block caching).

apps/token-manager/app/src/script.js
```diff
@@ -47,67 +44,66 @@ retryEvery(retry => {
...
-
-  return createStore(token, tokenAddr)
-}
+async function initialize(tokenAddress) {
+  const token = app.external(tokenAddress, tokenAbi)
 
-// Hook up the script as an aragon.js store
-async function createStore(token, tokenAddr) {
   return app.store(
     async (state, { address, event, returnValues }) => {
```
sohkai: We can do this later, but it may be nice to extract the `reducer` out into its own function from now on, now that there's a distinction between the `init` and the `reducer`. It may make it more clear at a glance what the `createStore()` is hooked up to.

PR #846, opened 2019-05-13
Update dynamic download dep

AquiGorka:
Updates to `file-saver` and fixes issue when exporting date range

### DIFF
--- a/apps/finance/app/src/components/Transfers.js
+++ b/apps/finance/app/src/components/Transfers.js
@@ -89,7 +89,9 @@ const getDownloadData = async (transfers, tokenDetails, resolveAddress) => {
       }
     )
   )
-  return ['Date,Name,Source/Recipient,Reference,Amount'].concat(mappedData)
+  return ['Date,Name,Source/Recipient,Reference,Amount']
+    .concat(mappedData)
+    .join('\n')
 }
 const getDownloadFilename = (dao, { start, end }) => {
   const today = format(Date.now(), 'yyyy-MM-dd')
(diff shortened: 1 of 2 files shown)


apps/finance/app/src/components/Transfers.js
```diff
@@ -89,19 +89,14 @@ const getDownloadData = async (transfers, tokenDetails, resolveAddress) => {
       }
     )
   )
-  const csvContent = [
-    'data:text/csv;charset=utf-8,Date,Name,Source/Recipient,Reference,Amount',
-  ]
-    .concat(mappedData)
-    .join('\n')
-  return window.encodeURI(csvContent)
+  return ['Date,Name,Source/Recipient,Reference,Amount'].concat(mappedData)
```
sohkai: I think the `.join()` being here is a bit more clear

apps/finance/app/src/components/Transfers.js
```diff
@@ -162,7 +157,7 @@ const Transfers = React.memo(({ dao, tokens, transactions }) => {
       resolveAddress
     )
     const filename = getDownloadFilename(dao, selectedDateRange)
-    download(data, filename, 'data:text/csv')
+    saveAs(new Blob([data.join('\n')]), filename)
```
bpierre: Should we provide the MIME type?

```suggestion
    saveAs(new Blob([data.join('\n')], {type: "text/csv;charset=utf-8"}), filename)
```

apps/finance/app/src/components/Transfers.js
AquiGorka: That's where it was originally but it made more sense to me to keep the function as a function that returns an array instead of a string, what do you think? should I change it?

sohkai: Depends what data format you think "data" should be; I felt like a string made more sense here (as the export "data") but if an array feels better to you let's leave it :).

PR #847, opened 2019-05-14
UX improvements date range picker

AquiGorka:
Reference: https://github.com/aragonone/product/issues/88

These changes:

- [x] Date range should be deselected by default displaying the following hint text: ‘Start date’ and ‘End date’, until user makes a date selection.
- [x] We should indicate today’s date and open the calendar on the current month
- [x] We should include an Apply and Clear buttons, that removes the current date range selection

**Screenshots:**

**How to test**:

```sh
# execute Aragon client with local apps:
REACT_APP_ASSET_BRIDGE=local npm run start

# in other terminal into /aragon-apps/apps/finance/app
git checkout feature/ux-improvements-date-range-picker
npm i
npm start 
```

Open up browser to http://localhost:3000 and head to Finance app

### DIFF
--- a/apps/finance/app/src/components/DateRange/DateRangeInput.js
+++ b/apps/finance/app/src/components/DateRange/DateRangeInput.js
@@ -100,11 +100,15 @@ class DateRangeInput extends React.PureComponent {
     })
   }
 
-  handleCancel = e => {
+  handleClear = e => {
     e.preventDefault()
     e.stopPropagation()
-    const { startDate, endDate } = this.props
-    this.setState({ showPicker: false, startDate, endDate })
+    this.setState({ showPicker: false, startDate: null, endDate: null }, () =>
+      this.props.onChange({
+        start: null,
+        end: null,
+      })
+    )
   }
 
   render() {
@@ -186,9 +190,9 @@ class DateRangeInput extends React.PureComponent {
               <Button
                 css={'width: 124px;'}
                 mode="secondary"
-                onClick={this.handleCancel}
+                onClick={this.handleClear}
               >
-                Cancel
+                Clear
               </Button>
               <Button
                 css={`
--- a/apps/finance/app/src/components/DateRange/DateRangeInput.js
+++ b/apps/finance/app/src/components/DateRange/DateRangeInput.js
@@ -204,7 +204,7 @@ class DateRangeInput extends React.PureComponent {
 
             <Controls>
               <Button
-                css={'width: 124px;'}
+                css="width: 124px"
                 mode="outline"
                 onClick={this.handleClear}
               >
--- a/apps/finance/app/src/components/DateRange/DateRangeInput.js
+++ b/apps/finance/app/src/components/DateRange/DateRangeInput.js
@@ -185,7 +185,7 @@ class DateRangeInput extends React.PureComponent {
               {(!compactMode || !startDateSelected) && (
                 <DatePicker
                   key={`start-picker-${startDate}`}
-                  name="start-date-picker"
+                  name="Start date"
                   currentDate={startDate}
                   onSelect={this.handleSelectStartDate}
                   overlay={false}
@@ -194,7 +194,7 @@ class DateRangeInput extends React.PureComponent {
               {(!compactMode || startDateSelected) && (
                 <DatePicker
                   key={`end-picker-${endDate}`}
-                  name="end-date-picker"
+                  name="End date"
                   currentDate={endDate}
                   onSelect={this.handleSelectEndDate}
                   overlay={false}
--- a/apps/finance/app/src/components/DateRange/DateRangeInput.js
+++ b/apps/finance/app/src/components/DateRange/DateRangeInput.js
@@ -189,7 +189,7 @@ class DateRangeInput extends React.PureComponent {
             <Controls>
               <Button
                 css={'width: 124px;'}
-                mode="secondary"
+                mode="outline"
                 onClick={this.handleClear}
               >
                 Clear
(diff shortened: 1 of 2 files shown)


dizzypaty:
Hey @AquiGorka, a couple of comments: 

* The start and end date hint text inside the date range component shouldn't use mono typeface, it should be Overpass Regular (15px #99A0A2)

* When users open the date range filter (default state), can we display the previous & current month instead of the 2 months being the current one? See mockups here [#809](https://github.com/aragon/aragon-apps/issues/809#issuecomment-487888505).

AquiGorka:
> The start and end date hint text inside the date range component shouldn't use mono typeface, it should be Overpass Regular (15px #99A0A2)

No prob. This is how it would look then:

image

apps/finance/app/src/components/DateRange/DateRangeInput.js
```diff
@@ -92,27 +86,69 @@ class DateRangeInput extends React.PureComponent {
...
       <IconCalendar />
     )
-    const value =
-      this.formattedStartDate && this.formattedEndDate
-        ? `${this.formattedStartDate} - ${this.formattedEndDate}`
+    // closed: shows props, if props null then placeholder
+    // opened:
+    //  compact: shows constants, till dates selected
+    //  not compact: shows props, changes with selection
+    const value = !showPicker
```
bpierre: What would you think about making this a function using early returns?

dizzypaty:
> No prob. This is how it would look then:
> image
 
> And when dates are selected there is not enough room: 
> image

@AquiGorka Can you elaborate on the reasons why it couldn't be done like on the designs? Do you see any other way than using a mono space font?

AquiGorka:
@dizzypaty the main reason is the text is being rendered in an input element where space-tweaking is limited. We could render the text on top of the input via another tag and then tweak the spacing there.

Having said that, and reading through some of @bpierre 's comments, this PR addresses the 3 items on the product issue, I have been patching this component to get it were we want it. Spending time re-implementing it from half-scratch to get the new style specs is on the list /cc @luisivan .

dizzypaty:
Hey @AquiGorka, I totally agree with you in that we shouldn’t be re-implementing it from scratch now – we haven't even redesign it yet with the new style and interaction behavior. My focus is solely to get it to a more usable state as it’s something that came a lot during the user testing sessions. 

I think that a low hanging fruit could be to display the previous and current months (i.e.: April – May), and when users selects a date such as [05-01-2019 — 05-15-2019], allow them to do so in the current calendar displayed. Otherwise, date selection feels unintuitive. But perhaps this is a bigger technical change than I imagine. 

Also, if the solution you suggested above to make the right alignment (“We could render the text on top of the input via another tag and then tweak the spacing there.”) is completely out of scope, let’s go back to monospace, as you first suggested. I think in this case, it’s more important to fix the interaction with the calendar than visual consistency in this case.

apps/finance/app/src/components/DateRange/DateRangeInput.js
```diff
@@ -92,27 +86,69 @@ class DateRangeInput extends React.PureComponent {
     }
   }
 
+  handleApply = e => {
+    e.preventDefault()
+    e.stopPropagation()
```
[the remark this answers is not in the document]
AquiGorka: Because of how [handleClickOutside](https://github.com/aragon/aragon-apps/pull/847/files/f7248ae5d6a29a133dbcd4211d52398e3ac5c288#diff-9b38b9af944da3e99573cf03d3e19786R65) is implemented - without the `stopPropagation` the popover would not close.

apps/finance/app/src/components/DateRange/DateRangeInput.js
AquiGorka: Let's make things for readable 😄 !

AquiGorka:
@bpierre 

## Visual:

> I can’t select a starting date after the ending date, but nothing indicates it.
> Are we not supposed to see the selection range with a specific background, like on the mockups?
> Isn’t the mobile version supposed to look different (fullscreen with a close button)?
> The weight and color of the current year doesn’t seem correct.
> The spacing of the calendar icon is a bit different than from the Figma file.
> The separator is different too (color and weight).

Discussed with @dizzypaty  and pushed back til component re-design.

> The left button should use mode="outline".

👍 

## Behavior:

> I think we should reset it when it gets closed without applying a range?

It keeps the previous selection, think of a user opening it by mistake, if clicking outside resets their previous choice it might not be intuitive (intuitive, odd word to use in the context of this component).

> The apply button doesn’t always become active. For example: open the picker, select 14 and 16, apply. Reopen the picker, select 13: the apply button stays inactive.

👍

sohkai:
I'd agree that the current component is very unintuitive, because you don't know that the left one is for the start date and the right one is for the end date. Once you know this, it becomes a lot more intuitive (especially on mobile).

It is quite a bit of work to get the calendar to behave well in a 2-panel set up, as we're using it here, since it was originally developed for a single date (not an interval). We'd likely have to start from scratch to be able to handle the 2-panel design well.

Perhaps something super small that could make it more usable now is to add headers for the corresponding input fields? E.g.

Screen Shot 2019-05-16 at 9 50 22 PM

It does make the calendar a bit big, but in this case UX > UI :smile:

AquiGorka:
cc @dizzypaty ☝️ thoughts? I can implement it easily and right away.

dizzypaty:
@sohkai @AquiGorka I think it's a good idea. It's a small change that can make things a bit clearer. We'll redesign and rebuild this date range component with the new design system so for now, it's a go ahead from my side : )

AquiGorka:
@sohkai 
> I think the only thing we could look into is setting a width property on the input boxes so we don't have to use the monospace font.

That wouldn't solve the issue, with input boxes we cannot set rules for how texts uses the available space; even with a set width the content would just use the font's width. A solution we proposed but will not work on for the time being is using a `div` on top of the input to display the data.
We will figure all this out with the redesign of the component.

PR #854, opened 2019-05-16
Finance - use updated app API with caching and init

2color:
## What
- Replace the `INITIALIZATION_TRIGGER` with an init function (part of the [new API](https://github.com/aragon/aragon.js/pull/297))
- Pass the external vault contract and its `initializationBlock` to the api to leverage caching
- Loading/Syncing indicator

## Dependency PR
- https://github.com/aragon/aragon.js/pull/297

### DIFF
--- a/apps/finance/app/src/script.js
+++ b/apps/finance/app/src/script.js
@@ -187,15 +187,13 @@ async function loadTokenBalances(state, settings) {
     return newState
   }
 
-  for (let i = 0; i < newState.balances.length; i++) {
-    const balance = newState.balances[i]
-    const amount = await loadTokenBalance(balance.address, settings)
-    newState.balances[i] = {
-      ...balance,
-      amount,
+  const addresses = newState.balances.map(({ address }) => address)
+  for (const address of addresses) {
+    newState = {
+      ...newState,
+      balances: await updateBalances(newState, address, settings),
     }
   }
-
   return newState
 }
 
--- a/apps/finance/app/src/script.js
+++ b/apps/finance/app/src/script.js
@@ -121,6 +121,7 @@ async function initialize(vaultAddress, ethAddress) {
 
       if (addressesEqual(eventAddress, vault.address)) {
         // Vault event
+        // TODO: Load balances for known token only once syncing past state has completed
         nextState = await vaultLoadBalance(nextState, event, settings)
       } else {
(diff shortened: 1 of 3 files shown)


apps/finance/app/src/script.js
```diff
@@ -169,16 +166,15 @@ async function createStore(settings) {
...
+  const newState = {
     periodDuration: marshallDate(
       await app.call('getPeriodDuration').toPromise()
     ),
     vaultAddress: settings.vault.address,
   }
 
-  const withTestnetState = await loadTestnetState(nextState, settings)
+  const withTestnetState = await loadTestnetState(newState, settings)
   const withEthBalance = await loadEthBalance(withTestnetState, settings)
```
sohkai: ⚠️ We should also refresh any existing token balances we know of here (could just loop through the balances and call `updateBalances()` on each token). If a user transfers some tokens to the Vault without using its `deposit()` functionality (by interacting with the token contract itself), and we don't process another event related to this token, we wouldn't know about it.

apps/finance/app/src/script.js
```diff
@@ -124,10 +119,9 @@ async function createStore(settings) {
         ...state,
       }
 
-      if (eventName === INITIALIZATION_TRIGGER) {
-        nextState = await initializeState(nextState, settings)
-      } else if (addressesEqual(eventAddress, vault.address)) {
+      if (addressesEqual(eventAddress, vault.address)) {
         // Vault event
+        // TODO: Load balances for known token only once syncing past state has completed
```
sohkai: I think we can remove this comment now :).

apps/finance/app/src/script.js
```diff
@@ -124,14 +119,18 @@ async function createStore(settings) {
...
         nextState = await vaultLoadBalance(nextState, event, settings)
       } else {
         // Finance event
         switch (eventName) {
+          case events.SYNC_STATUS_SYNCING:
+            nextState.isSyncing = true
+            break
+          case events.SYNC_STATUS_SYNCED:
+            nextState.isSyncing = false
+            break
```
sohkai: Small thing, could we lift these events out of the address check and do something similar to the [Voting app](https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/script.js#L69)?

PR #857, opened 2019-05-17
Use TokenBadge from aragonUI

bpierre:


apps/token-manager/app/src/components/SideBar.js
```diff
@@ -79,6 +78,7 @@ class SideBar extends React.Component {
                 address={tokenAddress}
                 name={tokenName}
                 symbol={tokenSymbol}
+                networkType={network.type}
```
AquiGorka: Is this no longer needed with the new badge?

bpierre: It is needed with the new badge yes! But at some point we’ll have providers for the network the tokens in aragonUI, which will make the address the only thing needed on `TokenBadge`.

PR #858, opened 2019-05-17
Voting countdowns: limit the units displayed

bpierre:
Requires https://github.com/aragon/aragon-ui/pull/377

Fixes https://github.com/aragon/aragon-apps/issues/638

image


bpierre:
No just testing data :laughing:

PR #862, opened 2019-05-21
Tokens: Local identities autocomplete

AquiGorka:
Needs before merging: https://github.com/aragon/aragon.js/pull/308

Referece: https://github.com/aragonone/product/issues/59

https://www.loom.com/share/2dd8032b97214967b78955b0f959ace8

Update: no longer uses mock data and video update

```
cd aragon-apps
git fetch
git checkout feature/local-identities-autocomplete
Go to Tokens -> Add Tokens -> Start typing something (name or address) that you know will have a hit
```


apps/token-manager/app/src/components/AutoComplete/AutoComplete.js
```diff
@@ -0,0 +1,161 @@
...
+      onSelect,
+      onChange,
+      renderItem,
+      required,
+      value,
+      wide,
+    },
+    ref
+  ) {
+    const [opened, setOpened] = useState(true)
```
2color: Is it opened by default?

apps/token-manager/app/src/components/AutoComplete/AutoComplete.js
```diff
@@ -0,0 +1,161 @@
...
+    const handleSelect = useCallback(
+      item => e => {
+        e.preventDefault()
+        onSelect(item)
+      },
+      [onSelect]
+    )
+    const handleChange = useCallback(
+      ({ target: { value } }) => onChange(value),
+      [onChange, value]
```
2color: Why is `value` a dependency for `useCallback`?

apps/token-manager/app/src/components/AutoComplete/AutoCompleteSelected.js
```diff
@@ -0,0 +1,108 @@
...
+    {
+      itemButtonStyles,
+      items,
+      onChange,
+      onSelect,
+      onSelectedClick = noop,
+      renderItem,
+      required,
+      renderSelected = identity,
+      selected,
```
2color: What is the role of `selected` and how does it differ from `value`?

2color: I saw below: `value` is the search term(s) and selected is the selected identity (or any other type) object

apps/token-manager/app/src/components/LocalIdentitiesAutoComplete/LocalIdentitiesAutoComplete.js
```diff
@@ -0,0 +1,171 @@
...
+        handleSearch(value)
+        onChange(value)
+      },
+      [onChange]
+    )
+    const handleSelect = useCallback(
+      selected => {
+        const { name, address } = selected
+        setSearchTerm(name)
+        handleSearch(name)
```
2color: Why do we search here upon selection?

apps/token-manager/app/src/components/AutoComplete/AutoCompleteSelected.js
```diff
@@ -0,0 +1,108 @@
...
+const noop = () => null
+
+const AutoCompleteSelected = React.memo(
+  React.forwardRef(function AutoCompleteSelected(
+    {
+      itemButtonStyles,
+      items,
+      onChange,
+      onSelect,
+      onSelectedClick = noop,
```
2color: Can you add a comment about the difference between `onSelect` and `onSelectedClick`. 

I'm not sure I got it based on the usage below.

apps/token-manager/app/src/hooks.js
```diff
@@ -0,0 +1,36 @@
+import { useRef, useEffect, useCallback } from 'react'
+
+export function useClickOutside(cb, existingRef) {
+  const ref = existingRef || useRef()
```
2color: When would one not pass an `existingRef`? 

I'm a bit confused about the utility of `ref` when not passing an `existingRef`

2color:
Looks really nice.

I noticed it's not possible to scroll down the options with the arrow keys. Any thoughts on that?

AquiGorka:
> I noticed it's not possible to scroll down the options with the arrow keys. Any thoughts on that?

Uses tab for navigation, I've considered using arrow keys but maybe for a next iteration

apps/token-manager/app/src/components/AutoComplete/AutoComplete.js
AquiGorka: Yes, but it won't show anything if there are no items and there will only be items when the user starts searching.

apps/token-manager/app/src/components/AutoComplete/AutoComplete.js
AquiGorka: It shouldn't, good catch.

apps/token-manager/app/src/components/AutoComplete/AutoCompleteSelected.js
AquiGorka: > I saw below: value is the search term(s) and selected is the selected identity (or any other type) object

Yes. This autocomplete components allow for rendering on complexer objects, but in the end it works on an input element so had to keep value and used selected (which for simple elements will probably just be some text, or the dev can use `AutoComplete` without `AutoCompleteSelected`.

apps/token-manager/app/src/components/LocalIdentitiesAutoComplete/LocalIdentitiesAutoComplete.js
AquiGorka: If the user clicks on the selected item the input box receives focus and the list of options is rendered, without the list might show more items than it should (number of items when they clicked on the item)

apps/token-manager/app/src/components/AutoComplete/AutoCompleteSelected.js
AquiGorka: Sure.

apps/token-manager/app/src/hooks.js
AquiGorka: `useClickOutside` can use an existing `ref` or return a new one:

```JavaScript
// creates new ref
const {ref}=useClickOutside(handleCb)
<input type="text" ref={ref} />

// existing
const {ref}=useClickOutside(handleCb, existingRef)
// ref and existingRef would be the same, but no need to consume ref anymore
useClickOutside(handleCb, existingRef)
```

apps/token-manager/app/src/components/AutoComplete/AutoCompleteSelected.js
AquiGorka: Reference:
`onSelect` happens when the user clicks on an item (selecting the item).
`onSelectedClick` happens when the user clicks on the rendered selected item.

apps/token-manager/app/src/components/AutoComplete/AutoComplete.js
AquiGorka: Thee main reason for this is to close (`!open`) the items box when a user selects and item or looses focus without setting a null search.

dizzypaty:
> > I noticed it's not possible to scroll down the options with the arrow keys. Any thoughts on that?
> 
> Uses tab for navigation, I've considered using arrow keys but maybe for a next iteration

Hey @AquiGorka, I've specified in the design specs for this issue [#59](https://github.com/aragonone/product/issues/59) that the autocomplete component should support keyboard navigation in the following manner: Users should be able to select an item from the autocomplete list navigating with the Up and Down arrows and hitting the enter key to execute the item in focus. The Tab key should be used for navigation between the different screen widgets/components. This is a standard and commonly learnt behaviour for any digital product when using keyboard navigation : )

AquiGorka:
@dizzypaty updated the PR, now you can cycle through the list using up/down arrows

apps/token-manager/app/src/hooks.js
```diff
@@ -0,0 +1,82 @@
...
+      }
+    },
+    [cb, ref]
+  )
+
+  return { ref, handleBlur }
+}
+
+export function useArrowKeysFocus(query, containerRef = useRef()) {
+  const [index, setIndex] = useState(-1)
```
2color: I'd suggest adding a comment or renaming `index` to something like `highlightedIndex`.

apps/token-manager/app/src/hooks.js
```diff
@@ -0,0 +1,82 @@
...
+      if (!elements[next]) {
+        next = -1
+      }
+      setIndex(next)
+    },
+    [index]
+  )
+  const handleKeyDown = useCallback(
+    ({ keyCode }) =>
+      keyCode === 38 ? cycleFocus(-1) : keyCode === 40 ? cycleFocus(1) : null,
```
2color: What do you think about switching the "magic numbers" with constants:
```suggestion
      keyCode === KEYCODE_UP ? cycleFocus(-1) : keyCode === KEYCODE_DOWN ? cycleFocus(1) : null,
```

PR #864, opened 2019-05-23
Tests: Enhance events assertions

facuspagnuolo:
Borrowing events helpers and assert methods from aragonOS to improve the way we are dealing with those in the apps tests, at least until we release a new version exporting those new libraries.


future-apps/payroll/test/contracts/Payroll_add_employee.test.js
```diff
@@ -1,5 +1,5 @@
 const { assertRevert } = require('@aragon/test-helpers/assertThrow')
-const { getEvents, getEventArgument } = require('../helpers/events')
+const { getEvents, getEventArgument } = require('../../../../shared/test-helpers/events')
```
sohkai: Could we use the `@aragon/test-helpers/` imports in the Payroll tests too?

facuspagnuolo: Sure, I completely ignored this one :)

PR #867, opened 2019-05-24
Finance - app loading state

2color:
## What

- Reduce sync state events to render the app loading state

## Related issues/PRs

- Dependent in aragon.js: https://github.com/aragon/aragon.js/pull/297
- Related issue: https://github.com/aragonone/product/issues/98

### DIFF
--- a/apps/finance/app/src/App.js
+++ b/apps/finance/app/src/App.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import styled from 'styled-components'
-import { Badge, EmptyStateCard, Main, SidePanel } from '@aragon/ui'
+import { SyncIndicator, EmptyStateCard, Main, SidePanel } from '@aragon/ui'
 import { useAragonApi } from '@aragon/api-react'
 import Balances from './components/Balances'
 import NewTransferPanelContent from './components/NewTransfer/PanelContent'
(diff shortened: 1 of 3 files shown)


apps/finance/app/src/script.js
```diff
@@ -1,4 +1,4 @@
-import Aragon from '@aragon/api'
+import Aragon, { SYNC_STATUS_SYNCING, SYNC_STATUS_SYNCED } from '@aragon/api'
```
sohkai: Perhaps an improvement in the API itself; what do you think about namespacing these constants so that we would instead do

```js
import { syncStatus } from '@aragon/api'

case syncStatus.SYNCING
```

Or similar?

sohkai:
cc @bpierre A bit of context in https://github.com/aragonone/product/issues/98#issuecomment-495547847 on if we should keep the `SYNC_STATUS_INITIALIZING` state.

apps/finance/app/src/App.js
```diff
@@ -106,6 +107,7 @@ class App extends React.Component {
               }}
               smallViewPadding={0}
             >
+              {isSyncing && <Badge>Syncing...</Badge>}
```
sohkai: @bpierre @dizzypaty Have we finalized the designs for how the Finance app should show its loading indicator? I imagine we'd want the `FloatIndicator`, but what about the balances and transactions?

dizzypaty: @sohkai After discussing with @bpierre about the different possibilities here, we arrived to the conclusion that the most practical option for now is to use the `FloatIndicator` for both the initial loading data state (removing the loading table skeleton) and for syncing data after the initial load. 

It's not a good idea to have a custom/different loading screen for each app and it's best to take this opportunity to redesign this component with the new style, layout and find a pattern that can be reused across all apps.

apps/finance/app/src/script.js
2color: Sounds good!

PR #861, opened 2019-05-18
Token-manager: Add syncing states (faked for now)

bpierre:
Requires https://github.com/aragon/aragon-ui/pull/381

PR #868, opened 2019-05-24
Token manager - loading state (with real states)

2color:
## What
- Builds on https://github.com/aragon/aragon-apps/pull/861
- Uses sync status reduced in the API

## Required PRs:
- `@aragon/api` https://github.com/aragon/aragon.js/pull/297
- `@aragon/ui` https://github.com/aragon/aragon-ui/pull/381


apps/token-manager/app/src/App.js
```diff
@@ -1,8 +1,13 @@
-import React from 'react'
+import React, { useState } from 'react'
```
AquiGorka: Is this used here?

apps/token-manager/app/src/script.js
```diff
@@ -65,8 +69,16 @@ async function initialize(tokenAddress) {
           default:
             return nextState
         }
+      } else {
+        // Finance event
```
sohkai:
```suggestion
        // Token Manager event
```

apps/token-manager/app/src/script.js
```diff
@@ -65,8 +69,16 @@ async function initialize(tokenAddress) {
...
+        switch (event) {
+          case SYNC_STATUS_SYNCING:
+          case SYNC_STATUS_SYNCED:
+            nextState.syncStatus = event
```
sohkai: Perhaps we could use normal `if` statements for this, and move this to be above the the token event handlers, since they're "special" events?

PR #869, opened 2019-05-24
[WIP] Voting: Single delegation (aka proxy voting)

facuspagnuolo:
*This PR aims to show some simple alternatives to tackle single delegation voting based on the requirements described in [the forum post](https://forum.aragon.org/t/voting-v1-single-delegation/440).*

There are two main problems we are trying to solve here:
- Having a way to vote with a pair of hot and cold wallets
- Having a way to delegate my vote to a representative that will vote on my behalf

That said, I've been exploring some simple alternatives that don't require changing the voting app at all <sup>(1)</sup>, as opposed in https://github.com/aragon/aragon-apps/pull/881.

## Terminology

*The following terminology is not set in stone, it was based on this [article](https://en.wikipedia.org/wiki/Proxy_voting) and followed to keep some consistency between the implementations showed.* 

**Principal:** The person assigning someone to vote on his/her behalf.

**Representative:** The person assigned to vote on behalf of someone.

### Principal's proxy
[[Implementation]](https://github.com/aragon/aragon-apps/pull/869/commits/120fc726cdb57da80b241bc72f451d95437faddf)

One option could be to have a particular smart contract that token holders (principals) can deploy, while transferring their tokens to this contract, that will basically allow them to control their representatives to vote on their behalf.

**PROS:**
- Principals have full control of their tokens
- Principals don't need to rely on someone else’s code
- Easy to have representatives for different domains
- Same vote gas costs for the principal or the representative

**CONS:**
- Representatives voting on behalf of many principals need to perform one transaction for each of them

### Representative's proxy
[[Implementation]](https://github.com/aragon/aragon-apps/pull/869/commits/661469ade01e00fa88541caba2974aa61db79be2)

Another option could be to have a particular smart contract that representatives can deploy, that will basically allow them to receive tokens from the principals being represented to vote on their behalf.

**PROS:**
- Representatives can vote on behalf of multiple principals with a single transaction

**CONS:**
- Principals need to rely on someone else’s code
- Principals may not have full control of their tokens and in which domains they are used for
- Cannot have multiple representatives per principal 
- Cannot overrule vote for a single principal, all or none

### Hybrid
[[Implementation]](https://github.com/aragon/aragon-apps/pull/869/commits/c2febdae772f6eb08483f3cedb84aabdb960b311)

A third alternative could be to try an implementation that allows having both ways, full control for principals while optimized gas costs for representatives.

----- 
<sup>(1)</sup> One of the requirements mentioned in the forum post for vote delegation is to allow the principal to overrule their representative's decision. With the current voting app this is actually not possible. Note that votes can be executed as soon as they reached the support and quorum expected.

### DIFF
--- a/apps/agent/package.json
+++ b/apps/agent/package.json
@@ -34,6 +34,7 @@
     "@aragon/apps-shared-migrations": "1.0.0",
     "@aragon/cli": "~5.6.0",
     "@aragon/test-helpers": "^1.1.0",
+    "eth-ens-namehash": "^2.0.8",
     "eth-gas-reporter": "^0.2.0",
     "ethereumjs-testrpc-sc": "^6.1.6",
     "ethereumjs-util": "^6.1.0",
--- a/apps/voting/contracts/HybridProxyVoting.sol
+++ b/apps/voting/contracts/HybridProxyVoting.sol
@@ -1,0 +1,226 @@
+pragma solidity ^0.4.24;
+
+import "./Voting.sol";
+import "./ProxyVoting.sol";
+import "./VotingHelpers.sol";
+import "@aragon/os/contracts/lib/token/ERC20.sol";
+import "@aragon/os/contracts/lib/math/SafeMath.sol";
+import "@aragon/os/contracts/common/IsContract.sol";
+import "@aragon/os/contracts/common/TimeHelpers.sol";
+
+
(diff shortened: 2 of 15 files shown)


facuspagnuolo:
Closing in favour of https://github.com/aragon/aragon-apps/pull/881, we decided to move forward with supporting voting delegation from the voting app itself.

PR #870, opened 2019-05-24
Rename Token Manager to Tokens

delfipolito:


apps/token-manager/app/src/components/Panels/AssignVotePanelContent.js
```diff
@@ -137,7 +137,7 @@ class AssignVotePanelContent extends React.Component {
       <div>
         <form onSubmit={this.handleSubmit}>
           <InfoMessage
-            title="Token Manager action"
+            title="Tokens action"
```
sohkai: This sounds a bit weird

Screen Shot 2019-05-24 at 4 22 59 PM

cc @dizzypaty

luisivan:
LGTM

apps/token-manager/app/src/components/Panels/AssignVotePanelContent.js
luisivan: Yeah, just Action is fine

PR #876, opened 2019-05-29
Apps: use consistent style across WebWorker scripts

sohkai:
Tries to apply a consistent style across the WebWorker scripts.

I liked leaving the `return nextState` at the end previously, because it was easy to add debugging logs, but this is still fairly easy to do by piping a `tap()` operator.

### DIFF
--- a/apps/finance/app/src/script.js
+++ b/apps/finance/app/src/script.js
@@ -115,43 +115,41 @@ async function initialize(vaultAddress, ethAddress) {
     async (state, event) => {
       const { vault } = settings
       const { address: eventAddress, event: eventName } = event
-      let nextState = {
+      const nextState = {
         ...state,
       }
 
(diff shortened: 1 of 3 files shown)


apps/finance/app/src/script.js
```diff
@@ -115,43 +115,41 @@ async function initialize(vaultAddress, ethAddress) {
...
+      // Finance event
+      switch (eventName) {
+        case 'ChangePeriodDuration':
+          nextState.periodDuration = marshallDate(
+            event.returnValues.newDuration
+          )
+          return nextState
+        case 'NewPeriod':
+          return {
+            ...(await newPeriod(nextState, event, settings)),
```
2color: Spread new state here too:

```suggestion
            ...newState
            ...(await newPeriod(nextState, event, settings)),
```

PR #881, opened 2019-06-03
Voting: Single delegation voting app

facuspagnuolo:
*This PR aims to explore a possible approach to tackle single delegation voting based on the requirements described in [the forum post](https://forum.aragon.org/t/voting-v1-single-delegation/440).*

There are two main problems we are trying to solve here:
- Having a way to vote with a pair of hot and cold wallets
- Having a way to delegate my vote to a representative that will vote on my behalf

That said, I've been exploring some alternatives to implement such features within the voting app itself as opposed in https://github.com/aragon/aragon-apps/pull/869.

## Terminology

*The following terminology is not set in stone, it was based on this [article](https://en.wikipedia.org/wiki/Proxy_voting) and followed to keep some consistency between the implementations showed.* 

**Principal:** The person assigning someone to vote on his/her behalf.

**Representative:** The person assigned to vote on behalf of someone.

## Approach

The basic idea here is to allow principals to designate representatives in the voting app itself. This means that if a representative was allowed by a principal, then they will be allowed to vote on their  behalf if certain conditions are met.

**PROS:**
- Principals have full control of their tokens
- Principals don't need to rely on someone else’s code
- Supports having multiple representatives per principal
- Supports casting a vote for multiple principals without requiring 1 tx for each of them

**CONS:**
- Different gas costs for the principal and the representative
- Extra storage required to allow representatives changing their vote
- Casting a vote for multiple principals requires a for loop


apps/voting/contracts/Voting.sol
```diff
@@ -361,14 +425,39 @@ contract Voting is IForwarder, AragonApp {
...
+               _isRepresentativeOf(_principal, _representative) &&
+               _hasNotVoteYet(_voteId, _principal) &&
+               !_withinOverruleWindow(_voteId);
+    }
+
+    function _isRepresentativeOf(address _principal, address _representative) internal view returns (bool) {
+        return representatives[_principal][_representative];
+    }
+
+    function _hasNotVoteYet(uint256 _voteId, address _principal) internal view returns (bool) {
```
bingen: Should it be `hasNotVotedYet`?

apps/voting/contracts/Voting.sol
```diff
@@ -361,14 +425,39 @@ contract Voting is IForwarder, AragonApp {
 
     function _canVote(uint256 _voteId, address _voter) internal view returns (bool) {
         Vote storage vote_ = votes[_voteId];
-
         return _isVoteOpen(vote_) && token.balanceOfAt(_voter, vote_.snapshotBlock) > 0;
     }
 
+    function _canVoteOnBehalfOf(uint256 _voteId, address _principal, address _representative) internal view returns (bool) {
```
bingen: Wouldn't it make sense to have `_voteExists` as acondition inside this funcion too?

apps/voting/contracts/Voting.sol
```diff
@@ -213,10 +269,27 @@ contract Voting is IForwarder, AragonApp {
...
 
-    function getVote(uint256 _voteId)
-        public
-        view
-        voteExists(_voteId)
+    /**
+    * @dev Initialization check is implicitly provided by `voteExists()` as new votes can only be
+    *      created via `newVote(),` which requires initialization
+    */
+    function canVoteOnBehalfOf(uint256 _voteId, address _principal, address _representative) public view voteExists(_voteId) returns (bool) {
```
bingen: As we already have the internal one, wouldn't it make sense to make this one external?

apps/voting/contracts/Voting.sol
```diff
@@ -361,14 +425,39 @@ contract Voting is IForwarder, AragonApp {
...
+    function _hasNotVoteYet(uint256 _voteId, address _principal) internal view returns (bool) {
+        VoterState state = getVoterState(_voteId, _principal);
+        return state == VoterState.Absent;
+    }
+
     function _isVoteOpen(Vote storage vote_) internal view returns (bool) {
         return getTimestamp64() < vote_.startDate.add(voteTime) && !vote_.executed;
     }
 
+    function _withinOverruleWindow(uint256 _voteId) internal view returns (bool) {
```
bingen: I would make this one private, as it doesn't have the `_voteExists` check.

apps/voting/contracts/Voting.sol
```diff
@@ -164,6 +179,36 @@ contract Voting is IForwarder, AragonApp {
...
+    * @param _principals Addresses of the principals voting on behalf of
+    * @param _voteIds Ids for each of the proxied votes
+    * @param _supports Whether the representative supports each of the proxied votes
+    */
+    function voteOnBehalfOfMany(address[] _principals, uint256[] _voteIds, bool[] _supports) external isInitialized {
+        require(_principals.length == _voteIds.length && _voteIds.length == _supports.length, ERROR_INVALID_DELEGATES_INPUT_LEN);
+        require(_principals.length <= MAX_VOTES_DELEGATION_SET_LENGTH, ERROR_DELEGATES_EXCEEDS_MAX_LEN);
+
+        for (uint256 i = 0; i < _principals.length; i++) {
+            if (_voteExists(_voteIds[i]) && _canVoteOnBehalfOf(_voteIds[i], _principals[i], msg.sender)) {
```
bingen: For consistency with the single vote version above, shouldn't this revert if the condition is not met?

apps/voting/contracts/Voting.sol
```diff
@@ -58,14 +62,19 @@ contract Voting is IForwarder, AragonApp {
     mapping (uint256 => Vote) internal votes;
     uint256 public votesLength;
 
+    uint64 public overruleWindow;
+    mapping (address => mapping (address => bool)) internal representatives;
```
bingen: If we finally decide not to allow having multiple representatives, I guess we could shorten this double mapping.

apps/voting/contracts/Voting.sol
```diff
@@ -58,14 +62,19 @@ contract Voting is IForwarder, AragonApp {
     mapping (uint256 => Vote) internal votes;
     uint256 public votesLength;
 
+    uint64 public overruleWindow;
```
bingen: Shouldn't this be added to the `initialize` function?

apps/voting/contracts/Voting.sol
```diff
@@ -125,13 +126,27 @@ contract Voting is IForwarder, AragonApp {
...
+    }
+
     /**
     * @notice Create a new vote about "`_metadata`"
     * @param _executionScript EVM script to be executed on approval
     * @param _metadata Vote metadata
     * @return voteId Id for newly created vote
     */
     function newVote(bytes _executionScript, string _metadata) external auth(CREATE_VOTES_ROLE) returns (uint256 voteId) {
+        // TODO: should we allow representatives to create votes on behalf of their principals?
```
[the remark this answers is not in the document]
izqui: Agreed

apps/voting/contracts/Voting.sol
```diff
@@ -164,6 +179,36 @@ contract Voting is IForwarder, AragonApp {
...
+    */
+    function voteOnBehalfOf(address _principal, uint256 _voteId, bool _supports) external voteExists(_voteId) {
+        require(_canVoteOnBehalfOf(_voteId, _principal, msg.sender), ERROR_REPRESENTATIVE_CANT_VOTE);
+        _vote(_voteId, _supports, _principal, false);
+    }
+
+    /**
+    * @notice Vote on behalf of many principals
+    * @param _principals Addresses of the principals voting on behalf of
+    * @param _voteIds Ids for each of the proxied votes
```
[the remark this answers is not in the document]
izqui: Regardless of how the UI works, there could be some gas savings (less calldata) if the only parameter that is an array in this function is `_principals` and both `_voteId` and `_supports` are the same for all.

apps/voting/contracts/Voting.sol
izqui: I don't think it is necessary since all the consumers of this function have already checked it, but we should make a comment about this

apps/voting/contracts/Voting.sol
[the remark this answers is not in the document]
bingen: Yes, forgot to mention in my comment but I was thinking that too.

apps/voting/contracts/Voting.sol
izqui: That's a good point, although it would be pretty wasteful to revert a batch of votes if just one cannot be processed (the principal votes at the same time). I would at least add an event when a vote cannot be processed though so we have some visibility

apps/voting/contracts/Voting.sol
bingen: But the function is internal, not private, so it could be misleading/dangerous. 
If I'm not wrong, the way it is right now it could happen that this function is called for a vote that doesn't exists and returns true, which would be formally correct from a logical point of view but wouldn't make any semantic sense in my opinion. (And if I'm wrong, then we should remove the `voteExists` check from the public wrapper).
Besides, generally speaking I like better having these kind of checks here in the inner part, although we already talked about that.

apps/voting/contracts/Voting.sol
facuspagnuolo: Sure we can have simply `mapping (address => address)`

apps/voting/contracts/Voting.sol
facuspagnuolo: I did this way thinking of current deployed voting apps, assuming those are already initialized, it won't work, they should use the setter instead. But now that you mention it, it may be necessary for new instances of the voting app :+1:

apps/voting/contracts/Voting.sol
```diff
@@ -174,6 +219,17 @@ contract Voting is IForwarder, AragonApp {
         _executeVote(_voteId);
     }
 
+    /**
+    * @notice Configure representative `representative` for the sender
```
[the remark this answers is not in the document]
facuspagnuolo: 🙌

apps/voting/contracts/Voting.sol
facuspagnuolo: I was just following the visibility pattern we are using for the rest of these "kind of getter" functions, thoughts?

apps/voting/contracts/Voting.sol
facuspagnuolo: I was thinking of returning a list here, or maybe two, with the successful and unsuccessful votes, what do you think?

apps/voting/contracts/Voting.sol
```diff
@@ -164,6 +179,36 @@ contract Voting is IForwarder, AragonApp {
...
+
+    /**
+    * @notice Vote on behalf of many principals
+    * @param _principals Addresses of the principals voting on behalf of
+    * @param _voteIds Ids for each of the proxied votes
+    * @param _supports Whether the representative supports each of the proxied votes
+    */
+    function voteOnBehalfOfMany(address[] _principals, uint256[] _voteIds, bool[] _supports) external isInitialized {
+        require(_principals.length == _voteIds.length && _voteIds.length == _supports.length, ERROR_INVALID_DELEGATES_INPUT_LEN);
+        require(_principals.length <= MAX_VOTES_DELEGATION_SET_LENGTH, ERROR_DELEGATES_EXCEEDS_MAX_LEN);
```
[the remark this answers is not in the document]
facuspagnuolo: Just to prevent that from the contract itself, we can work it out with a nice UX tho

apps/voting/contracts/Voting.sol
```diff
@@ -174,6 +219,17 @@ contract Voting is IForwarder, AragonApp {
         _executeVote(_voteId);
     }
 
+    /**
+    * @notice Configure representative `representative` for the sender
+    * @param _representative Address of the representative to configure for the sender
+    * @param _allowed Boolean to tell whether the representative is being allowed or not
+    */
+    function setRepresentative(address _representative, bool _allowed) external isInitialized {
+        // TODO: can we validate sth here to avoid malicious people from spamming events?
```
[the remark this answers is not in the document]
facuspagnuolo: Yea, checking balances could drive to some undesired scenarios, for example, you wouldn't able to allow a representative if you transfer your tokens right after a vote was created, weird scenario tho. Since it doesn't seem harmful, I'd leave it as is.

apps/voting/contracts/Voting.sol
```diff
@@ -174,6 +219,17 @@ contract Voting is IForwarder, AragonApp {
...
+        representatives[msg.sender][_representative] = _allowed;
```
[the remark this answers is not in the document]
facuspagnuolo: I agree we can go with that simplified version for now :)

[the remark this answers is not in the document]
facuspagnuolo: Good point, I think it is not harmful to allow principals to do that, it is actually helpful in terms of gas costs. Unless there is side effect, I think we can leave it as is.

apps/voting/contracts/Voting.sol
facuspagnuolo: I fixed it to receive a storage pointer instead

apps/voting/contracts/Voting.sol
facuspagnuolo: There are more than one internal function receiving a vote id without checking the vote actually exists. However, I see your point but I also think we should find a way of performing that check only once. In that case, we can turn the internal functions to receive a `Vote` per se instead of an id, assuming the consumer already checked the given storage pointer is valid.

apps/voting/contracts/Voting.sol
bingen: But I guess we need the event for the frontend.

facuspagnuolo: Sure, I was thinking of doing both

facuspagnuolo:
@bingen @izqui thanks for your review guys! I addressed most of your comments and answered others in their threads. I finally simplified the implementation to have only one representative per principal and fixed `voteOnBehalfOfMany` to expect only a list of principals, and now I think most of the weird scenarios have been cleared out. 

A  few things that we still need to discuss are:
- Should we allow representatives to change their vote outside the overrule window?
- Do we want to store who casted the vote or at least if it was proxied?
- Do we want to change the voting app to turn the early execution aspect configurable?

facuspagnuolo:
@bingen @izqui I finally decided to:
- Store the issuer of a vote, 
- Allow representatives to change their vote, and
- Drop `principal` terminology in favor of `voter` since that was the one we were using before. 

I also proposed a way to make early execution something configurable [here](https://github.com/aragon/aragon-apps/pull/886). It may be nice to have that as well to ensure principals have always a chance to overrule their representative's decision.

facuspagnuolo:
@izqui @bingen based on what we discussed offline, I found a more optimized way to track the issuer of a vote (see https://github.com/aragon/aragon-apps/pull/881/commits/d4b0a9d35fc15745c48242a568a82a8873b1bc2c). 

Having in mind that with the current voting app it costs `85366` gas to cast a vote, I arrived to that solution after comparing the following alternatives: 
- A: Tracking the issuer address always
- B: Tracking the issuer address always but using `address(0)` for the principal
- C: Tracking the issuer address only when delegating a vote or overruling

| Gas costs as a ...   |   A    |   B    |   C    |
|----------------------|--------|--------|--------|
| Principal            | 105696 | 90697  | 85776  |
| Representative       | 111742 | 111982 | 111968 |
| Principal overruling | 66060  | 51061  | 51462  |

Comparing results against current behavior:

| Gas diffs as a ...   |   A    |   B   |   C   |
|----------------------|--------|-------|-------|
| Principal            | 20330  |  5331 |  410  |
| Representative       | 26376  | 26616 | 26602 |
| Principal overruling | -19306 | -34305|-33904 |

As you can see, option C is the one with the smallest delta and almost as cheap as optionB in terms of overruling. Additionally, the overload a representative has compared to principals casting their own votes is pretty much the same in all cases.

apps/voting/contracts/Voting.sol
```diff
@@ -418,4 +588,32 @@ contract Voting is IForwarder, AragonApp {
...
+            vote_.nay = vote_.nay.add(voterStake);
+        }
+
+        vote_.voters[_voter] = _supports ? VoterState.Yea : VoterState.Nay;
+        emit CastVote(_voteId, _voter, _supports, voterStake);
+    }
+
+    function _overwriteIssuerIfNecessary(Vote storage vote_, address _voter) private {
+        address _currentIssuer = vote_.issuers[_voter];
+        if (_currentIssuer != address(0) && _currentIssuer != _voter) {
```
bingen: Second part of the condition is in case a voter votes in behalf of himself, right?

apps/voting/contracts/Voting.sol
```diff
@@ -117,6 +128,19 @@ contract Voting is IForwarder, AragonApp {
         emit ChangeMinQuorum(_minAcceptQuorumPct);
     }
 
+    /**
+    * @notice Change overrule window to a duration of `@transformTime(_newOverruleWindow)`
+    * @param _newOverruleWindow New overrule window in seconds
+    */
+    function changeOverruleWindow(uint64 _newOverruleWindow)
```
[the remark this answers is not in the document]
facuspagnuolo: Oh sorry, I missed that comment. Actually we didn't discussed that, I actually see it as feature, you can "disable" proxy voting by setting a large overrule window for example

apps/voting/contracts/Voting.sol
```diff
@@ -152,10 +176,46 @@ contract Voting is IForwarder, AragonApp {
...
+
+    /**
+    * @notice Vote `_supports ? 'yes' : 'no'` in vote #`_voteId` on behalf of many voters
+    * @dev Initialization check is implicitly provided by `voteExists()` as new votes can only be
+    *      created via `newVote(),` which requires initialization
+    * @param _voters Addresses of the voters voting on behalf of
+    * @param _voteId Id for vote
+    * @param _supports Whether the representative supports the vote
+    */
+    function voteOnBehalfOfMany(address[] _voters, uint256 _voteId, bool _supports) external voteExists(_voteId) {
```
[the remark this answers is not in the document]
facuspagnuolo: I see what you mean, I see two issues here:
- The representative may not be allowed to vote himself, it could be any address.
- If the representative is allowed, enforcing them through this entry point won't prevent them from changing their vote with the current `vote()` entry point afterwards. And adding a guard there would increase gas costs for all the voters, what do you think?

apps/voting/contracts/Voting.sol
```diff
@@ -418,4 +588,32 @@ contract Voting is IForwarder, AragonApp {
         uint256 computedPct = _value.mul(PCT_BASE) / _total;
         return computedPct > _pct;
     }
+
+    function _storeVote(Vote storage vote_, uint256 _voteId, bool _supports, address _voter) private {
+        uint256 voterStake = token.balanceOfAt(_voter, vote_.snapshotBlock);
+        VoterState state = vote_.voters[_voter];
```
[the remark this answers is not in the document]
facuspagnuolo: Good catch, I missed this one 👍

apps/voting/contracts/Voting.sol
facuspagnuolo: exactly :)

apps/voting/contracts/Voting.sol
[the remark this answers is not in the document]
bingen: Yes, sure, but I would see at "init" time. The problem would be to change it just before creating a vote or with a vote ongoing.

apps/voting/contracts/Voting.sol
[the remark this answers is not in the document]
bingen: Yes, you are absolutely right.

apps/voting/test/contracts/voting_delegation.js
```diff
@@ -0,0 +1,564 @@
...
+
+                  it('returns false', async () => {
+                    assert.isFalse(await voting.canVoteOnBehalfOf(voteId, voter, representative), 'should not be able to vote')
+                  })
+                })
+              })
+            })
+
+            context('when the voter has already voted', () => {
+              beforeEach('move within overrule window', async () => {
```
izqui:
```suggestion
              beforeEach('voter votes', async () => {
```

apps/voting/test/contracts/voting_delegation.js
```diff
@@ -0,0 +1,564 @@
...
+                assert.equal(await voting.getVoteIssuer(voteId, voter), anotherRepresentative, 'vote issuer does not match')
+
+                assertAmountOfEvents(receipt, 'CastVote', 1)
+                assertEvent(receipt, 'CastVote', { voter, voteId, supports: true, stake: bigExp(51, 18) })
+
+                assertAmountOfEvents(receipt, 'ProxyVote', 1)
+                assertEvent(receipt, 'ProxyVote', { voter, representative: anotherRepresentative, succeed: true, voteId, supports: true })
+              })
+
+              it('can be overrule by the voter', async () => {
```
izqui:
```suggestion
              it('can be overruled by the voter', async () => {
```

apps/voting/test/contracts/voting_delegation.js
```diff
@@ -0,0 +1,564 @@
...
+                assert.equal(await getVoterState(representative), VOTER_STATE.ABSENT, 'representative should not have voted')
+                assert.equal(await voting.getVoteIssuer(voteId, voter), voter, 'vote issuer does not match')
+
+                assertAmountOfEvents(receipt, 'CastVote', 1)
+                assertEvent(receipt, 'CastVote', { voter, voteId, supports: true, stake: bigExp(51, 18) })
+              })
+            })
+
+            context('when the voter has already voted', () => {
+              beforeEach('move within overrule window', async () => {
```
izqui:
```suggestion
              beforeEach('voter votes', async () => {
```

apps/voting/contracts/Voting.sol
izqui: Why don't we set it to `address(0)` then to get the gas refund if this is the case?

facuspagnuolo: Makes sense 😅

apps/voting/contracts/Voting.sol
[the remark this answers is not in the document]
facuspagnuolo: I don't see why not allowing that behavior, I don't see this as a core setting that should be one-shot as the voting time for example. Do you see any inconsistencies in case we allow that?

facuspagnuolo:
Thanks again @bingen and @izqui ! Already updated the base branch to `next`.
Please @sohkai let me know if you would like to take a look as well.

apps/voting/contracts/Voting.sol
```diff
@@ -268,10 +366,21 @@ contract Voting is IForwarder, AragonApp {
...
+
+    /**
+    * @dev Return the issuer of a given vote by its ID
+    * @param _voteId Vote identifier
+    * @param _voter Address of the voter
+    * @return Address of the issuer of the voter's vote
+    */
+    function getVoteIssuer(uint256 _voteId, address _voter) public view voteExists(_voteId) returns (address) {
+        return _voteIssuer(votes[_voteId], _voter);
     }
```
sohkai: Super nitpick: can we re-order the new internal and private functions so they're mostly in alphabetical order?

Unless something should be logically grouped, I find alphabetical definitions a pretty easy way to recognize where to put new functions :).

apps/voting/contracts/Voting.sol
```diff
@@ -306,45 +415,38 @@ contract Voting is IForwarder, AragonApp {
...
 
-        if (_executesIfDecided && _canExecute(_voteId)) {
+        if (_executesIfDecided && _canExecute(vote_)) {
             // We've already checked if the vote can be executed with `_canExecute()`
             _unsafeExecuteVote(_voteId);
         }
     }
 
+    /**
+    * @dev Internal function to check if a representative can vote on behalf of a voter. It assumes the queried vote exists.
```
sohkai:
```suggestion
    * @dev Internal function for a representative to cast a vote on behalf of a voter. It assumes the queried vote exists.
```

apps/voting/contracts/Voting.sol
```diff
@@ -306,45 +415,38 @@ contract Voting is IForwarder, AragonApp {
     * @dev Internal function to cast a vote. It assumes the queried vote exists.
     */
     function _vote(uint256 _voteId, bool _supports, address _voter, bool _executesIfDecided) internal {
```
sohkai: Could we change most of the internal and private functions to use `Vote storage _vote` instead?

apps/voting/contracts/Voting.sol
```diff
@@ -394,17 +496,85 @@ contract Voting is IForwarder, AragonApp {
...
+    * @dev Internal function to get the state of a voter. It assumes the queried vote exists.
+    */
+    function _voterState(Vote storage vote_, address _voter) internal view returns (VoterState) {
+        return vote_.voters[_voter];
+    }
+
+    /**
+    * @dev Internal function to get the issuer of a vote. It assumes the queried vote exists.
+    */
+    function _voteIssuer(Vote storage vote_, address _voter) internal view returns (address) {
```
sohkai: Nitpick on the variable names, since it's a bit confusing to read some of these new internal functions. We use a prefixed `_` (e.g. `_vote`) for the argument name and a suffix `_` in variable names (e.g. `vote_`) only if it would otherwise shadow a storage variable.

apps/voting/contracts/Voting.sol
```diff
@@ -394,17 +496,85 @@ contract Voting is IForwarder, AragonApp {
...
     /**
-    * @dev Internal function to check if a vote is still open
+    * @dev Internal function to check if a representative can vote on behalf of a voter on a certain vote. It assumes the queried vote exists.
+    * @return True if the given representative can vote on behalf of a voter on a certain vote, false otherwise
+    */
+    function _canVoteOnBehalfOf(Vote storage vote_, address _voter, address _representative) internal view returns (bool) {
+        return _canVote(vote_, _voter) &&
+               _isRepresentativeOf(_voter, _representative) &&
+               _hasNotVotedYet(vote_, _voter) &&
+               !_withinOverruleWindow(vote_);
```
sohkai: Nitpick: a small probabilistic optimization here is to reorder the logic.

I imagine most of the time, representative voters will attempt to vote within the time boundaries, and with correct arguments (but as mentioned above in `voteOnBehalfOfMany()`, may have some assumptions move underneath them). Hence:

- `_hasNotVotedYet()` is the most likely to fail, because a voter may have already voted themselves, without the representative knowing
- `_isRepresentativeOf()`, `_withinOverruleWindow()`, and `_canVote()` are mostly error checks to make sure that the representative does, indeed, have the ability to vote.

However, as mentioned in `voteOnBehalfOfMany()`, we may want to split off `_withinOverruleWindow()` and `_canVote()` into their own error check.

apps/voting/contracts/Voting.sol
```diff
@@ -152,10 +176,46 @@ contract Voting is IForwarder, AragonApp {
...
+    * @dev Initialization check is implicitly provided by `voteExists()` as new votes can only be
+    *      created via `newVote(),` which requires initialization
+    * @param _voter Address of the voter voting on behalf of
+    * @param _voteId Id for vote
+    * @param _supports Whether the representative supports the vote
+    */
+    function voteOnBehalfOf(address _voter, uint256 _voteId, bool _supports) external voteExists(_voteId) {
+        Vote storage vote_ = votes[_voteId];
+        require(_canVoteOnBehalfOf(vote_, _voter, msg.sender), ERROR_REPRESENTATIVE_CANT_VOTE);
+        _voteOnBehalfOf(_voteId, _supports, _voter, msg.sender);
```
sohkai: Perhaps we may want to invoke `voteOnBehalfOfMany()` here with an array of one, so it can follow the same logic (emitting events on fail instead of reverting).

apps/voting/contracts/Voting.sol
sohkai: Wanted to bring this up again; was there a decision around adding it to the `initialize()` function?

I was thinking about if a default of 0 made sense, and my conclusion is that it does (so adding this to the `initialize()` function is somewhat optional and really depends on if we want to make using it in templates cheaper).

apps/voting/contracts/Voting.sol
```diff
@@ -58,14 +63,20 @@ contract Voting is IForwarder, AragonApp {
     mapping (uint256 => Vote) internal votes;
     uint256 public votesLength;
 
+    uint64 public overruleWindow;
```
sohkai: A note as well, since we're not including the `overruleWindow` in each vote, is that someone may change this value underneath existing votes such that some representatives may not have had a chance to vote (assuming they go with the strategy of waiting until the very end).

apps/voting/contracts/Voting.sol
```diff
@@ -58,14 +63,20 @@ contract Voting is IForwarder, AragonApp {
...
+    mapping (address => address) internal representatives;
+
     event StartVote(uint256 indexed voteId, address indexed creator, string metadata);
     event CastVote(uint256 indexed voteId, address indexed voter, bool supports, uint256 stake);
+    event ProxyVote(uint256 indexed voteId, address indexed voter, address indexed representative, bool supports, bool succeed);
```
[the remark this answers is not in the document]
facuspagnuolo: Note that both events will be emitted actually, I added this one to avoid changing the signature of `CastVote`. IMO, logging only the values that will allow us to match them should be enough

apps/voting/contracts/Voting.sol
```diff
@@ -47,6 +51,7 @@ contract Voting is IForwarder, AragonApp {
         uint256 votingPower;
         bytes executionScript;
         mapping (address => VoterState) voters;
+        mapping (address => address) issuers;
```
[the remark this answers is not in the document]
facuspagnuolo: Let's use `voteCasters`

shared/test-helpers/ganache-cli.sh
```diff
@@ -18,26 +18,28 @@ setup_coverage_variables() {
   BALANCE=${BALANCE-100000}
   GAS_LIMIT=${GAS_LIMIT-0xfffffffffff}
   NETWORK_ID=${NETWORK_ID-16}
+  ACCOUNTS=${ACCOUNTS-200}
```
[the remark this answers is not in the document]
facuspagnuolo: https://github.com/aragon/aragon-apps/pull/923

apps/agent/package.json
```diff
@@ -34,6 +34,7 @@
     "@aragon/apps-shared-migrations": "1.0.0",
     "@aragon/cli": "~5.6.0",
     "@aragon/test-helpers": "^2.0.0",
+    "eth-ens-namehash": "^2.0.8",
```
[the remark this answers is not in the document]
facuspagnuolo: https://github.com/aragon/aragon-apps/pull/922

apps/voting/contracts/Voting.sol
facuspagnuolo: I did with most of them, but I decided not to change those that are actually relying on the id. For example, if we change this function, we will need to pass both the id and the storage pointer, which could lead to undesired behaviors if someone uses it wrongly. An alternative could be to turn these cases into `private` functions to ensure we're the only ones calling those. WDYT?

apps/voting/contracts/Voting.sol
facuspagnuolo: Any reason to do that only for the `internal` and `private` functions? I suggest tackling convention-related changes in all apps at once, otherwise we will never have the consistency we're actually trying to get. WDYT?

apps/voting/contracts/Voting.sol
```diff
@@ -166,6 +226,15 @@ contract Voting is IForwarder, AragonApp {
         _executeVote(_voteId);
     }
 
+    /**
+    * @notice `_allowed ? 'Set' : 'Remove'` `_representative` as the representative
+    * @param _representative Address of the representative to be allowed on behalf of the sender. Use the zero address for none.
+    */
+    function setRepresentative(address _representative) external isInitialized {
```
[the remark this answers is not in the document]
facuspagnuolo: Agree, let's make it public

apps/voting/contracts/Voting.sol
facuspagnuolo: Sure, do you mean I'm not following that convention here?

facuspagnuolo:
@sohkai thx for the review! I addressed most of your comments and left some answers for a few of them. Please let me know your thoughts 🤗

apps/voting/contracts/Voting.sol
```diff
@@ -418,4 +600,32 @@ contract Voting is IForwarder, AragonApp {
         uint256 computedPct = _value.mul(PCT_BASE) / _total;
         return computedPct > _pct;
     }
+
+    function _storeVote(Vote storage vote_, uint256 _voteId, bool _supports, address _voter) private {
```
sohkai: _Super_ nitpick, but perhaps we can name this `_castVote`? Store sounds like it's just manipulating storage, but "cast" sounds like an action that could also (and it does) emit an event.

apps/voting/contracts/Voting.sol
sohkai: Hmm yes, I didn't notice some of the new public functions weren't alphabetized.

I agree, we can do it all later (and most of the apps _should_ be fairly standard now...). I guess the biggest offenders are `setRepresentative()`, `withinOverruleWindow()`, and `isRepresentativeOf()`, which could be moved around a bit.

apps/voting/contracts/Voting.sol
sohkai: Ah yes, that makes sense. I see most of these have to emit events, which require the ID 👍.

apps/voting/contracts/Voting.sol
```diff
@@ -58,14 +65,21 @@ contract Voting is IForwarder, AragonApp {
     mapping (uint256 => Vote) internal votes;
     uint256 public votesLength;
 
+    uint64 public overruleWindow;
+    mapping (address => address) internal representatives;
+
     event StartVote(uint256 indexed voteId, address indexed creator, string metadata);
     event CastVote(uint256 indexed voteId, address indexed voter, bool supports, uint256 stake);
+    event ProxyVoteFailed(uint256 indexed voteId, address indexed voter, address indexed representative);
```
sohkai:
```suggestion
    event ProxyVoteFailure(uint256 indexed voteId, address indexed voter, address indexed representative);
```

apps/voting/contracts/Voting.sol
```diff
@@ -58,14 +65,21 @@ contract Voting is IForwarder, AragonApp {
...
+    event ProxyVoteSucceed(uint256 indexed voteId, address indexed voter, address indexed representative, bool supports);
```
sohkai:
```suggestion
    event ProxyVoteSuccess(uint256 indexed voteId, address indexed voter, address indexed representative, bool supports);
```

apps/voting/contracts/Voting.sol
```diff
@@ -47,6 +52,8 @@ contract Voting is IForwarder, AragonApp {
         uint256 votingPower;
         bytes executionScript;
         mapping (address => VoterState) voters;
+        mapping (address => address) casters;
```
sohkai: I think it would still be nice to add a bit of comments explaining how `casters` is used (only for representatives casting votes).

apps/voting/contracts/Voting.sol
sohkai: Yes, the `vote_` variable should be `_vote` since it's an argument.

facuspagnuolo: Note that we are always using `vote_` for vote storage pointers cause we would shadow the `_vote` function otherwise

sohkai: :(

Maybe we can find a different name (e.g. `_voteAsPrinciple()`)?

apps/voting/contracts/Voting.sol
facuspagnuolo: way better! thx :)

luisivan:
Congrats @facuspagnuolo for getting this merged! :fireworks:

PR #883, opened 2019-06-05
Voting: Complete contract inline documentation

facuspagnuolo:
Simply adding and polishing voting contract inline documentation.

### DIFF
--- a/apps/voting/contracts/Voting.sol
+++ b/apps/voting/contracts/Voting.sol
@@ -70,12 +70,12 @@ contract Voting is IForwarder, AragonApp {
     }
 
     /**
-     * @notice Initialize Voting app with `_token.symbol(): string` for governance, minimum support of `@formatPct(_supportRequiredPct)`%, minimum acceptance quorum of `@formatPct(_minAcceptQuorumPct)`%, and a voting duration of `@transformTime(_voteTime)`
-     * @param _token MiniMeToken Address that will be used as governance token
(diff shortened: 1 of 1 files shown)


apps/voting/contracts/Voting.sol
```diff
@@ -311,14 +335,17 @@ contract Voting is IForwarder, AragonApp {
         }
     }
 
+    /**
+     * @dev Internal function to execute a vote. It assumes the queried vote exists.
```
bingen: Really good to add this here. It's unrelated to this PR (and I have said this before) but I don't like how this is done, as it's prone to error. This is supposed to be a "safe" function but it's not as it could be writing to an uninitialized vote in the mapping. It could be ourselves if we ever change this contract or anybody else inheriting it as it's an `internal` function.

facuspagnuolo: I see your point, but ofc it's a trade off between that and optimizing gas costs, specially in those situations where we want to ensure multiple conditions. That's why I added this in the meantime, at least until we come up with a decision, I think it helps pointing that out.

facuspagnuolo:
Thanks @bingen! I don't if it is a standard for us or not the way I aligned the `*`s. It was the way I did for Payroll and I wanted to replicate it here. Let's roll it back and change all in case we like the idea.

bingen:
> It was the way I did for Payroll

Hm, I'm afraid that was my fault! Hahaha.

PR #884, opened 2019-06-06
voting: catch invalid script error

2color:
## What
- Fix an an caught error which would crash the background script 
- The error was detected with the following `newVote` [tx](https://etherscan.io/tx/0x5dc24de292dcc6e4728b973629a4d3f1a9eb1a2954a851d054f4507e273c2c01)
- Render votes with invalid scripts with an error message

## Bug Example 
- https://mainnet.aragon.org/#/futhorcecosystem/0xd9ed670476974e371607c9a6f8bffd44462510af

### DIFF
--- a/apps/voting/app/src/script.js
+++ b/apps/voting/app/src/script.js
@@ -239,6 +239,7 @@ async function loadVoteDescription(vote) {
   } catch (error) {
     console.error('Error describing vote script', error)
     vote.description = 'Invalid script. The result cannot be executed.'
+    // Clear metadata so ensure it's rendered with a description rather than question
     vote.metadata = ''
   }
 
--- a/apps/voting/app/src/script.js
+++ b/apps/voting/app/src/script.js
@@ -237,7 +237,7 @@ async function loadVoteDescription(vote) {
           .join('\n')
       : ''
   } catch (error) {
-    console.log('Error describing vote script', error)
+    console.error('Error describing vote script', error)
     vote.description = 'Invalid script. The result cannot be executed.'
     vote.metadata = ''
   }
--- a/apps/voting/app/src/script.js
+++ b/apps/voting/app/src/script.js
@@ -240,7 +240,7 @@ async function loadVoteDescription(vote) {
     console.error('Error describing vote script', error)
     vote.description = 'Invalid script. The result cannot be executed.'
     // Clear metadata so ensure it's rendered with a description rather than question
-    vote.metadata = ''
+    vote.metadata = null
   }
 
   return vote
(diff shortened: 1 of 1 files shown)


apps/voting/app/src/script.js
```diff
@@ -223,17 +223,24 @@ async function loadVoteDescription(vote) {
...
+            const app = step.name ? `${step.name}${identifier}` : `${step.to}`
+
+            return `${app}: ${step.description || 'No description'}`
+          })
+          .join('\n')
+      : ''
+  } catch (error) {
+    console.log('Error describing vote script', error)
+    vote.description = 'Invalid script. The result cannot be executed.'
+    vote.metadata = ''
```
AquiGorka: Why is this necessary?

apps/voting/app/src/script.js
```diff
@@ -223,17 +223,24 @@ async function loadVoteDescription(vote) {
...
+          .map(step => {
+            const identifier = step.identifier ? ` (${step.identifier})` : ''
+            const app = step.name ? `${step.name}${identifier}` : `${step.to}`
+
+            return `${app}: ${step.description || 'No description'}`
+          })
+          .join('\n')
+      : ''
+  } catch (error) {
+    console.log('Error describing vote script', error)
```
AquiGorka: Should this be our function `log`? (the one that only logs in dev)

2color: Actually this should be `console.error`, it appears to be standard practice to use it. 

Logging for debugging is a different story.  We probably want to introduce a lean logger at some point

apps/voting/app/src/script.js
2color: That's how actions are differentiated from question votes. It's not really clear here, but if you look at the reduced state, all votes with a script have empty metadata. Because this fails, we have to manually set it to an empty string. Otherwise it's rendered as a question vote.

2color: https://github.com/aragon/aragon-apps/blob/catch-invalid-script-error/apps/voting/app/src/components/VotePanel.js#L114-L117

AquiGorka: Sorry, about my lack of context, in the code before there was no setting of metadata but now there is, I was wondering what it was for. Thanks for clarifying.

AquiGorka: From the link you share, maybe setting the value to `null` would be better no?

izqui: What would happen now if a vote has both description (from a script) and metadata? The description should always take precedence

2color: In the vote panel they're each rendered if truthy. 

In the card metadata takes precedence:
https://github.com/aragon/aragon-apps/blob/51e480001c6975baf0721164a07f8815e973b97c/apps/voting/app/src/components/VotingCard/VotingCard.js#L79

Should we switch the two?

izqui: We should, yes

PR #886, opened 2019-06-07
Voting: Make early execution configurable

facuspagnuolo:
Related to https://github.com/aragon/aragon/issues/590

This PR aims to turn the fact that votes can be early executed once they have already been decided, into a configurable option.

### DIFF
--- a/apps/voting/contracts/Voting.sol
+++ b/apps/voting/contracts/Voting.sol
@@ -18,6 +18,7 @@ contract Voting is IForwarder, AragonApp {
     bytes32 public constant CREATE_VOTES_ROLE = keccak256("CREATE_VOTES_ROLE");
     bytes32 public constant MODIFY_SUPPORT_ROLE = keccak256("MODIFY_SUPPORT_ROLE");
     bytes32 public constant MODIFY_QUORUM_ROLE = keccak256("MODIFY_QUORUM_ROLE");
+    bytes32 public constant MODIFY_EARLY_EXECUTION_ROLE = keccak256("MODIFY_EARLY_EXECUTION_ROLE");
     bytes32 public constant MODIFY_OVERRULE_WINDOW_ROLE = keccak256("MODIFY_OVERRULE_WINDOW_ROLE");
 
     uint64 public constant PCT_BASE = 10 ** 18; // 0% = 0; 1% = 10^16; 100% = 10^18
@@ -34,6 +35,7 @@ contract Voting is IForwarder, AragonApp {
     string private constant ERROR_CAN_NOT_EXECUTE = "VOTING_CAN_NOT_EXECUTE";
     string private constant ERROR_CAN_NOT_FORWARD = "VOTING_CAN_NOT_FORWARD";
     string private constant ERROR_NO_VOTING_POWER = "VOTING_NO_VOTING_POWER";
+    string private constant ERROR_SAME_EARLY_EXECUTION = "VOTING_SAME_EARLY_EXECUTION";
     string private constant ERROR_WITHIN_OVERRULE_WINDOW = "VOTING_WITHIN_OVERRULE_WINDOW";
     string private constant ERROR_INVALID_OVERRULE_WINDOW = "VOTING_INVALID_OVERRULE_WINDOW";
     string private constant ERROR_DELEGATES_EXCEEDS_MAX_LEN = "VOTING_DELEGATES_EXCEEDS_MAX_LEN";
@@ -56,6 +58,7 @@ contract Voting is IForwarder, AragonApp {
         mapping (address => VoterState) voters;
         mapping (address => address) casters;
         uint64 overruleWindow;
+        bool earlyExecutionDisallowed;
     }
 
     MiniMeToken public token;
@@ -67,6 +70,7 @@ contract Voting is IForwarder, AragonApp {
     mapping (uint256 => Vote) internal votes;
     uint256 public votesLength;
 
+    bool private earlyExecutionDisallowed;
     uint64 public overruleWindow;
     mapping (address => address) internal representatives;
 
@@ -77,6 +81,7 @@ contract Voting is IForwarder, AragonApp {
     event ExecuteVote(uint256 indexed voteId);
     event ChangeSupportRequired(uint64 supportRequiredPct);
     event ChangeMinQuorum(uint64 minAcceptQuorumPct);
+    event ChangeEarlyExecution(bool allowed);
     event ChangeOverruleWindow(uint64 newOverruleWindow);
     event ChangeRepresentative(address indexed voter, address indexed newRepresentative);
 
@@ -134,6 +139,20 @@ contract Voting is IForwarder, AragonApp {
     }
 
     /**
+    * @notice `_earlyExecutionAllowed ? 'Allow' : 'Disallow'` early execution of votes
+    * @param _earlyExecutionAllowed Early execution setting for new votes
+    */
+    function changeEarlyExecution(bool _earlyExecutionAllowed)
+        external
+        authP(MODIFY_EARLY_EXECUTION_ROLE, arr(uint256(_earlyExecutionAllowed ? 1 : 0), uint256(earlyExecutionDisallowed ? 0 : 1)))
+    {
+        bool earlyExecutionDisallowed_ = !_earlyExecutionAllowed;
+        require(earlyExecutionDisallowed != earlyExecutionDisallowed_, ERROR_SAME_EARLY_EXECUTION);
+        earlyExecutionDisallowed = earlyExecutionDisallowed_;
+        emit ChangeEarlyExecution(_earlyExecutionAllowed);
+    }
+
+    /**
     * @notice Change overrule window to a duration of `@transformTime(_newOverruleWindow)`
     * @param _newOverruleWindow New overrule window in seconds
     */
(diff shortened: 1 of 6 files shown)


apps/voting/contracts/Voting.sol
```diff
@@ -86,9 +90,17 @@ contract Voting is IForwarder, AragonApp {
     * @param _supportRequiredPct Percentage of yeas in casted votes for a vote to succeed (expressed as a percentage of 10^18; eg. 10^16 = 1%, 10^18 = 100%)
     * @param _minAcceptQuorumPct Percentage of yeas in total possible votes for a vote to succeed (expressed as a percentage of 10^18; eg. 10^16 = 1%, 10^18 = 100%)
     * @param _voteTime Seconds that a vote will be open for token holders to vote (unless enough yeas or nays have been cast to make an early decision)
+    * @param _earlyExecutionAllowed Boolean telling whether votes are allowed to be early executed or not
     * @param _overruleWindow Seconds representing the period where a voter will be able to override a representative's decision on a vote
     */
-    function initialize(MiniMeToken _token, uint64 _supportRequiredPct, uint64 _minAcceptQuorumPct, uint64 _voteTime, uint64 _overruleWindow)
+    function initialize(
```
izqui: Not specific to this PR, but wondering whether we should keep old initializers for backwards compatibility. The old initializer could just use the current initializer and set the new parameters to the default zero value (which would have the same effect as if the app was initialized before the upgrade)

If we don't do that, as most templates get the app base from the latest version in the aragonPM repo, publishing a new version with a different initializer will break the old templates.

facuspagnuolo: I completely agree with your point, I actually choose that pattern in the past but I remember someone recommending not to do it. IMO, I think it's best to leave initializers as they are and add a separate setter to configure whatever is needed to ensure we are backwards compatible

apps/voting/contracts/Voting.sol
```diff
@@ -63,6 +65,7 @@ contract Voting is IForwarder, AragonApp {
     mapping (uint256 => Vote) internal votes;
     uint256 public votesLength;
 
+    bool public earlyExecutionAllowed;
```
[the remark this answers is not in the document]
facuspagnuolo: Sure, makes a lot of sense. I completely missed that

apps/voting/contracts/Voting.sol
```diff
@@ -464,31 +482,33 @@ contract Voting is IForwarder, AragonApp {
...
 
-        // Voting is already decided
-        if (_isValuePct(vote_.yea, vote_.votingPower, vote_.supportRequiredPct)) {
+        // If the vote is already decided and early execution is allowed, it can be executed
+        if (!earlyExecutionDisallowed && _isValuePct(vote_.yea, vote_.votingPower, vote_.supportRequiredPct)) {
             return true;
         }
 
-        // Vote ended?
+        // If the vote is still open but early execution is not allowed, it cannot be executed
```
bingen: I think this comment is not accurate. It should be something like:
```
If the vote is still open but early execution is not allowed or vote is not decided yet (checked just above), it cannot be executed
```
Or just leave:
```
If the vote is still open, it cannot be executed
```
and assume reader has gone through previous lines, which is probably more consistent with the rest of comments in this functions, that don't include information about previous checks.

apps/voting/test/contracts/voting.js
```diff
@@ -302,6 +317,20 @@ contract('Voting App', ([root, holder1, holder2, holder20, holder29, holder51, n
                     assert.equal(await executionTarget.counter(), 2, 'should have executed result')
                 })
 
+                it('cannot early execute vote when early execution is disallow even when decided', async () => {
+                    await voting.changeEarlyExecution(false)
+                    await voting.vote(voteId, true, true, { from: holder51 }) // causes execution
```
izqui:
```suggestion
                    await voting.vote(voteId, true, true, { from: holder51 })
```

apps/voting/contracts/Voting.sol
```diff
@@ -63,6 +65,7 @@ contract Voting is IForwarder, AragonApp {
     mapping (uint256 => Vote) internal votes;
     uint256 public votesLength;
 
+    bool public earlyExecutionDisallowed;
```
izqui: We could make it a private state variable and implement a `earlyExecutionAllowed` getter, as the rest of the contract's external interface (including the `ChangeEarlyExecution` event) uses `allowed` instead of `disallowed`

luisivan:
Included this for 0.9, the next release that will introduce smart contract changes. Thanks @facuspagnuolo!

apps/voting/contracts/Voting.sol
```diff
@@ -128,6 +132,20 @@ contract Voting is IForwarder, AragonApp {
...
 
+    /**
+    * @notice `_earlyExecutionAllowed? 'Allow' : 'Disallow'` votes early execution
+    * @param _earlyExecutionAllowed New votes early execution setting
+    */
+    function changeEarlyExecution(bool _earlyExecutionAllowed)
+        external
+        authP(MODIFY_EARLY_EXECUTION_ROLE, arr(uint256(_earlyExecutionAllowed ? 1 : 0), uint256(earlyExecutionDisallowed ? 0 : 1)))
+    {
+        bool _earlyExecutionDisallowed = !_earlyExecutionAllowed;
```
sohkai: Could we rename this to `earlyExecutionDisallowed_` instead, to avoid confusion with arguments and local variables?

apps/voting/test/contracts/voting.js
```diff
@@ -302,6 +317,20 @@ contract('Voting App', ([root, holder1, holder2, holder20, holder29, holder51, n
...
+                    await voting.vote(voteId, true, true, { from: holder51 })
+
+                    const [isOpen, isExecuted] = await voting.getVote(voteId)
+                    assert.isTrue(isOpen, 'vote should still be open')
+                    assert.isFalse(isExecuted, 'vote should have not been executed')
+                    assert.isFalse(await voting.canExecute(voteId), 'vote cannot be executed')
+
+                    await voting.mockIncreaseTime(VOTING_DURATION + 1)
+                    await voting.executeVote(voteId)
+                    assert.equal(await executionTarget.counter(), 2, 'should have executed result')
```
sohkai: Should we also add a check that `voting.canExecute()` is true now?

apps/voting/test/contracts/voting.js
```diff
@@ -114,6 +116,19 @@ contract('Voting App', ([root, holder1, holder2, holder20, holder29, holder51, n
...
+
+            const receipt = await voting.changeEarlyExecution(false)
+            assertAmountOfEvents(receipt, 'ChangeEarlyExecution')
+            assertEvent(receipt, 'ChangeEarlyExecution', { allowed: false })
+
+            assert.isFalse(await voting.isEarlyExecutionAllowed(), 'early execution should be disallowed')
+        })
+
+        it('fails trying to set the same early execution setting', async () => {
+            await assertRevert(voting.changeEarlyExecution(true), 'VOTING_SAME_EARLY_EXECUTION')
```
sohkai: We should also add a test to check that the Voting app is correctly configured after initialization (e.g. has the right support / quorum, but also its default values for early execution and overrule window).

apps/voting/test/contracts/voting.js
```diff
@@ -413,26 +466,56 @@ contract('Voting App', ([root, holder1, holder2, holder20, holder29, holder51, n
...
+                await voting.vote(voteId, true, true, { from: holder2 })
+
+                const [isOpen, isExecuted] = await voting.getVote(voteId)
+
+                assert.isTrue(isOpen, 'vote should still be open')
+                assert.isFalse(isExecuted, 'vote should have not been executed')
+                assert.isFalse(await voting.canExecute(voteId), 'vote cannot be executed')
+            })
+
+            it('creating vote as holder2 does not executes vote', async () => {
```
sohkai:
```suggestion
            it('creating vote as holder2 does not execute vote', async () => {
```

apps/voting/contracts/Voting.sol
```diff
@@ -464,31 +490,33 @@ contract Voting is IForwarder, AragonApp {
...
     function _canExecute(Vote storage vote_) internal view returns (bool) {
+        // If vote is already executed, it cannot be executed again
         if (vote_.executed) {
             return false;
         }
 
-        // Voting is already decided
-        if (_isValuePct(vote_.yea, vote_.votingPower, vote_.supportRequiredPct)) {
+        // If the vote is already decided and early execution is allowed, it can be executed
+        if (isEarlyExecutionAllowed() && _isValuePct(vote_.yea, vote_.votingPower, vote_.supportRequiredPct)) {
```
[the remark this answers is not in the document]
facuspagnuolo: Since I'll address your suggestion of storing this value within the vote struct, this line will change

apps/voting/test/contracts/voting.js
facuspagnuolo: isn't that being checked [here](https://github.com/aragon/aragon-apps/pull/886/files#diff-5af58e0b9d17bd3d4f22ad98562608bdR120)?

facuspagnuolo: will move it to a separate test to make it clearer

apps/voting/scripts/vote-stats.js
```diff
@@ -48,13 +48,13 @@ module.exports = async (cb) => {
   const voting = Voting.at(appAddress)
 
   const [
-    isOpen,
```
facuspagnuolo: I did this because I was hitting the `stack too deep` issue... Based on what I checked, this returned value wasn't being used in the frontend.

facuspagnuolo:
@sohkai thanks for the review! Already addressed your comments, please LMKYT!

apps/voting/scripts/vote-stats.js
sohkai: Sounds good! I'm quite sure we do all of the end date calculations locally in the frontend so no need to worry about it there.

This script is used to generate a CSV file that can be used in data visualization tools, and here as well, we shouldn't need the `isOpen` field.

However, I think it'd be worth introducing another public function to allow another contract on-chain to check for a vote's open status, if we're not hitting the contract size limit yet.

facuspagnuolo: Sure. I didn't want to add it to avoid increasing the bytecode, but if that’s not a problem let's go with it

PR #887, opened 2019-06-12
Delete duplicated "details_url" from manifest.json

ottodevs:
This commit pretends to remove "details_url" that seems duplicated from line 6 to me.
Was probably not harming too much, but IMHO is better to fix the duplication, does it make sense?


sohkai:
Hmm, good catch @ottodevs! This might've been a result of a merge conflict, I see a few other apps also have this issue :/. Will fix and merge this!

PR #895, opened 2019-07-02
Travis: remove deprecated sudo

sohkai:
See https://github.com/aragon/aragon/pull/856.


sohkai:
Not quite sure what's going on with the coverage report, but all the coverage tests (that aren't expected to fail) pass :shrug:

PR #898, opened 2019-07-03
Payroll: Move payroll example kit into `contracts/examples/` dir

facuspagnuolo:
Simply to avoid confusing with a real kit, it is only used for testing purposes

### DIFF
--- a/future-apps/payroll/.solcover.js
+++ b/future-apps/payroll/.solcover.js
@@ -2,6 +2,7 @@ module.exports = {
     norpc: true,
     copyPackages: ['@aragon/os', '@aragon/apps-finance', '@aragon/apps-vault', '@aragon/test-helpers'],
     skipFiles: [
+        'examples',
         'test',
         '@aragon/os',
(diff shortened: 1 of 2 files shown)


future-apps/payroll/.solcover.js
```diff
@@ -2,7 +2,6 @@ module.exports = {
     norpc: true,
     copyPackages: ['@aragon/os', '@aragon/apps-finance', '@aragon/apps-vault', '@aragon/test-helpers'],
     skipFiles: [
-        'PayrollKit.sol',
```
sohkai: Should we add `examples/` to this skip list?

facuspagnuolo: sure, forgot to update this one 😅

facuspagnuolo:
> We should also double check if https://github.com/aragon/aragon-apps/blob/master/future-apps/payroll/package.json#L27 still works (I think it does, assuming Truffle compiles it).

confirmed! 👌

PR #900, opened 2019-07-03
Payroll: Add owed salary getter function

facuspagnuolo:
*Fixing finding 6.7 of the audit report*

### 6.7 Create an external function to view employees balance

| Minor | Open | This issue is currently under review. |

#### Description

There is no simple way to find out how much money an employee can withdraw as a payroll. Even though the employee can withdraw the maximum amount by simply passing `0` as `_requestedAmount` parameter, it would still be useful to have that information accessible through the external view function.  

#### Remediation

Create an external view function that returns the amount of money that can be currently withdrawn from Payroll by an employee.


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -840,6 +844,17 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
         revert(ERROR_INVALID_PAYMENT_TYPE);
     }
 
+    function _getTotalOwedCappedSalary(uint256 _employeeId) private view returns (uint256) {
```
sohkai: I think this is a good candidate to make `internal` 🤷‍♂

facuspagnuolo: let's make it internal 😄

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -473,6 +478,15 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
         endDate = employee.endDate;
     }
 
+    /**
+     * @dev Get owed salary since last payroll for an employee. It will contemplate the accrued salary as well.
```
```suggestion
     * @dev Get owed salary since last payroll for an employee. It will take into account the accrued salary as well.
```

PR #901, opened 2019-07-03
Payroll: Remove allowed tokens

facuspagnuolo:
*Fixing finding 6.5 of the audit report*

⚠️ Note that I had to make other changes to apply this one since we're hitting the max size limit of a contract. For instance, I reduce the `getEmployeeByAddress` to simply `getEmployeeIdByAddress`.

### 6.5 Allowed tokens list should be modifiable

| Medium | Open | This issue is currently under review. |

#### Description

In the current implementation, there can only be a maximum of 20 allowed tokens that are used for payroll. There is no way to modify the `allowedTokens` list except for adding new tokens before the number of allowed tokens hits 20.

#### Examples

1. If there will be more than 20 tokens (e.g. stablecoins) bounded to the real-world currencies.
2. If the company switches multiple times from some set of coins to another.
3. The company does not want to make a payroll in some tokens anymore.
4. The address of a specific token is changed due to smart contract updates. 

#### Remediation

Allow modifying `allowedTokens` list while maintaining security. Modifying the list is tricky and might introduce new attack vectors by employer, as the employee payment allocation is currently bounded to the order of allowed tokens list.


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -51,6 +51,7 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
     string private constant ERROR_SENDER_DOES_NOT_MATCH = "PAYROLL_SENDER_DOES_NOT_MATCH";
     string private constant ERROR_FINANCE_NOT_CONTRACT = "PAYROLL_FINANCE_NOT_CONTRACT";
     string private constant ERROR_TOKEN_ALREADY_ALLOWED = "PAYROLL_TOKEN_ALREADY_ALLOWED";
+    string private constant ERROR_MISSING_ALLOWED_TOKEN = "PAYROLL_MISSING_ALLOWED_TOKEN";
```
sohkai: Perhaps `PAYROLL_TOKEN_NOT_ALLOWED` may be clearer and follows the other case.

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -115,16 +116,10 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
         uint256 exchangeRate,
         string paymentReference
     );
-    event AddAllowedToken(address indexed token);
+    event SetAllowedToken(address indexed token, bool indexed allowed);
```
sohkai: Don't think the bool really needs to be indexed

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -861,6 +841,29 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+    function _getAllowedTokenIndex(address _token) internal view returns (uint256) {
+        for (uint256 i = 0; i < allowedTokensArray.length; i++) {
+            if (allowedTokensArray[i] == _token) {
+                return i;
+            }
+        }
+        revert(ERROR_MISSING_ALLOWED_TOKEN);
+    }
+
+    function _ensureEmployeeTokenAllocationsAddsUp(Employee storage employee_) internal view {
```
sohkai: Maybe `_ensureEmployeeTokenAllocationsIsValid` is a bit better for the name :)

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -314,17 +305,13 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
             delete employee.allocation[allowedTokensArray[j]];
         }
 
-        // Check distribution sums to 100
-        uint256 sum = 0;
+        // Set distributions checking tokens are allowed
```
```suggestion
        // Set distributions only if given tokens are allowed
```

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -25,8 +25,8 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
     * bytes32 constant public SET_EMPLOYEE_SALARY_ROLE = keccak256("SET_EMPLOYEE_SALARY_ROLE");
     * bytes32 constant public ADD_BONUS_ROLE = keccak256("ADD_BONUS_ROLE");
     * bytes32 constant public ADD_REIMBURSEMENT_ROLE = keccak256("ADD_REIMBURSEMENT_ROLE");
-    * bytes32 constant public ALLOWED_TOKENS_MANAGER_ROLE = keccak256("ALLOWED_TOKENS_MANAGER_ROLE");
-    * bytes32 constant public CHANGE_PRICE_FEED_ROLE = keccak256("CHANGE_PRICE_FEED_ROLE");
+    * bytes32 constant public MODIFY_ALLOWED_TOKENS_ROLE = keccak256("MODIFY_ALLOWED_TOKENS_ROLE");
```
sohkai: Let's use the "manage" terminology instead for this.

PR #902, opened 2019-07-03
Payroll: Use PPF rate precision

facuspagnuolo:
*Fixing finding 6.6 of the audit report*

### 6.6 Decimal calculations `ONE`, can differ between Price feed and Payroll

| Minor | Open | This issue is currently under review. |

#### Description
The decimal calculation is using a hardcoded `ONE = 10 ** 18`; in both PPF (**uint256**) and Payroll (**uint128**) contracts. 

- In `PPF.sol`:

**code/ppf-contracts/packages/ppf-contracts/contracts/PPF.sol:L11**
```solidity
uint256 constant public ONE = 10 ** 18; // 10^18 is considered 1 in the price feed to allow for decimal calculations
```

- In `Payroll.sol`:

**code/payroll/future-apps/payroll/contracts/Payroll.sol:L32**
```solidity
```

This can result in an issue in some calculations as the types are different (uint128 v.s. uint 256) but mainly for future/thirdparty IFEED interfaces that specify different decimal points for `ONE`. 

#### Examples

- In `PPF.sol`:

**code/ppf-contracts/packages/ppf-contracts/contracts/PPF.sol:L113-L116**
```solidity
function get(address base, address quote) public view returns (uint128, uint64) {
    if (base == quote) {
        return (uint128(ONE), getTimestamp64());
    }
```

- In `Payroll.sol`:

**code/payroll/future-apps/payroll/contracts/Payroll.sol:L627-L630**
```solidity
// Convert amount (in denomination tokens) to payout token and apply allocation
uint256 tokenAmount = _totalAmount.mul(exchangeRate).mul(tokenAllocation);
// Divide by 100 for the allocation percentage and by ONE for the exchange rate precision
tokenAmount = tokenAmount / (100 * ONE);
```

#### Remediation

In `Payroll.sol` where `ONE` is used, read the value from PPF contract, or simply `require` them to be the same.

### DIFF
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -601,7 +601,7 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
                 // Convert amount (in denomination tokens) to payout token and apply allocation
                 uint256 tokenAmount = _totalAmount.mul(exchangeRate).mul(tokenAllocation);
                 // Divide by 100 for the allocation percentage and by the exchange rate precision
-                tokenAmount = tokenAmount.div(feed.ratePrecision().mul(100));
+                tokenAmount = tokenAmount.div(100).div(feed.ratePrecision());
 
                 // Finance reverts if the payment wasn't possible
                 finance.newImmediatePayment(token, employeeAddress, tokenAmount, paymentReference);
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -39,7 +39,6 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
     bytes32 constant public MODIFY_PRICE_FEED_ROLE = 0x74350efbcba8b85341c5bbf70cc34e2a585fc1463524773a12fa0a71d4eb9302;
     bytes32 constant public MODIFY_RATE_EXPIRY_ROLE = 0x79fe989a8899060dfbdabb174ebb96616fa9f1d9dadd739f8d814cbab452404e;
 
-    uint128 internal constant ONE = 10 ** 18; // 10^18 is considered 1 in the price feed to allow for decimal calculations
     uint256 internal constant MAX_ALLOWED_TOKENS = 20; // prevent OOG issues with `payday()`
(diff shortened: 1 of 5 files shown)


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -601,8 +600,8 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
 
                 // Convert amount (in denomination tokens) to payout token and apply allocation
                 uint256 tokenAmount = _totalAmount.mul(exchangeRate).mul(tokenAllocation);
-                // Divide by 100 for the allocation percentage and by ONE for the exchange rate precision
-                tokenAmount = tokenAmount / (100 * ONE);
+                // Divide by 100 for the allocation percentage and by the exchange rate precision
+                tokenAmount = tokenAmount.div(feed.ratePrecision().mul(100));
```
```suggestion
                tokenAmount = tokenAmount.div(100).div(feed.ratePrecision());
```

Might be a bit clearer. We could also remove the safe math, since division does nothing.

I was also wondering if it's worth the PITA to do this in assembly so we can use a static call... I can't see the price feed re-entering into this contract, but it's possible it'd be able to change rates on the spot or do something else funny in itself on this call.

Although, of course, we already know the price feed is pretty much a trusted component at this point 😉.

facuspagnuolo: Yea, I'd skip the staticcall for now, we'd have to handle reverts as well

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -601,8 +600,8 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
 
                 // Convert amount (in denomination tokens) to payout token and apply allocation
                 uint256 tokenAmount = _totalAmount.mul(exchangeRate).mul(tokenAllocation);
-                // Divide by 100 for the allocation percentage and by ONE for the exchange rate precision
-                tokenAmount = tokenAmount / (100 * ONE);
+                // Divide by 100 for the allocation percentage and by the exchange rate precision
+                tokenAmount = tokenAmount.div(100).div(feed.ratePrecision());
```
sohkai: Actually, another thought on this, we could cache this value before going into the for loop since it'll always be the same.

facuspagnuolo: sure, will do that in the last PR to avoid triggering the CI again 😓

PR #903, opened 2019-07-04
Payroll: Add payroll remainders as accrued salary

facuspagnuolo:
*Fixing finding 6.3 of the audit report*

### 6.3 Employee may lose up to a second of the salary

| Medium | Open | This issue is currently under review. |

#### Description

An employee can lose up to a second of their salary on calling `payday` function. It happens if the employee requests a specific amount of money that cannot be formed by multiplying salary-per-second base-unit `denominationTokenSalary` and any amount of seconds. On that case, rounding is done in favour of an employer. 

**code/payroll/future-apps/payroll/contracts/Payroll.sol:L751-L759**
```solidity
uint256 timeDiff = _paidAmount.div(employee.denominationTokenSalary);

// We check if the division was perfect, and if not, take its ceiling to avoid giving away
// tiny amounts of salary.
// Employees may lose up to a second's worth of payroll if they use the "request partial
// amount" feature or reach salary amounts that get capped by uint max.
if (timeDiff.mul(employee.denominationTokenSalary) != _paidAmount) {
    timeDiff = timeDiff.add(1);
}
```

This issue has also been [discussed](https://github.com/aragon/aragon-apps/tree/payroll-readme/future-apps/payroll#losing-one-second-of-pay-when-using-partial-payments) by the project team.

#### Remediation

While losing up to a second might not sound like a big issue, it can be easily fixed by adding this rounding error to the accrued salary of the employee instead of just keeping it to the employer. Adding this to the accrued salary, might have side-effects e.g. that a terminated employee cannot immediately be removed from storage after cashing out, because `payday` accounted the lost fraction of a second to accrued salary.

### DIFF
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -655,7 +655,41 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
             // No need to use SafeMath here as we already know _paymentAmount > accruedSalary
             currentSalaryPaid = _paymentAmount - accruedSalary;
         }
-        employee.lastPayroll = _getLastPayrollDate(_employeeId, currentSalaryPaid);
+        _updateEmployeeLastPayrollDate(_employeeId, currentSalaryPaid);
+    }
+
+    /**
+     * @dev Update the last payroll date for an employee based on the requested payment amount. If the requested amount
+     *      cannot be represented by a multiple of the employee's salary per second, it will be added as accrued salary.
+     * @param _employeeId Employee's identifier
+     * @param _paidAmount Requested amount to be paid to the employee
+     */
+    function _updateEmployeeLastPayrollDate(uint256 _employeeId, uint256 _paidAmount) internal {
+        Employee storage employee = employees[_employeeId];
+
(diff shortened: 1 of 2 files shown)


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -655,7 +655,41 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+        if (timeDiff.mul(employee.denominationTokenSalary) != _paidAmount) {
+            timeDiff = timeDiff.add(1);
+            // No need to use SafeMath here, we already now that _paidAmount is lower than its ceil
+            uint256 remainder = timeDiff.mul(employee.denominationTokenSalary) - _paidAmount;
+
+            uint256 currentAccruedSalary = employee.accruedSalary;
+            uint256 newAccruedSalary = currentAccruedSalary + remainder;
+            if (newAccruedSalary < currentAccruedSalary) {
+                newAccruedSalary = MAX_UINT256;
+            }
```
sohkai: I think we can just use `employee.accruedSalary = employee.accruedSalary.add(remainder)` here; it should be impossible for this to overflow (we always reset accrued salary first).

facuspagnuolo:
Thanks @sohkai for #906, it looks much clearer now 😄

PR #905, opened 2019-07-05
Add prepublish script before publishing to aragonPM

sohkai:
Requires https://github.com/aragon/aragon-cli/pull/571 (and the aragonCLI upgrade).

cc @0xGabi


0xGabi:
LGTM 

We will probably release a major version cause we introduced a breaking change making ipfs optional. 

We could also release a minor only with the PRs for updating `apm publish`.

0xGabi:
`aragon apm publish` now needs confirmation so I included the `--skip-confirmation` flag to skip confirmation. 

@sohkai What do you think about propagate content. Is this something we want to do as default on each publish? There is other flag: `--propagate-content` that we might want to configure as well.

sohkai:
@0xGabi Having that flag would be useful too, although if it prompts users by default, I think that's fine as well.

0xGabi:
Oh ok, I thought it might break the CI. But I just realize that all checks have pass 😅

Will revert last commit then.

sohkai:
Ah, we don't publish from the CI 😄

0xGabi:
@sohkai Ready 👍

facuspagnuolo (approved):

PR #906, opened 2019-07-05
Payroll: optimize current payday remainders

sohkai:
Inlines `_updateEmployeeLastPayrollDate()` and optimizes a bit of the math / SSTOREs.

### DIFF
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -651,42 +651,27 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
         if (accruedSalary > 0) {
             // Employee is cashing out a mixed amount between previous and current owed salaries;
             // first use up their accrued salary
-            employee.accruedSalary = uint256(0);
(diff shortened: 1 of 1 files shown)


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -651,42 +651,27 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+        uint256 timeDiff = currentSalaryPaid.div(salary);
+
+        // If they're being paid an amount that doesn't match perfectly with the adjusted time
+        // (up to a seconds' worth of salary), add the second and put the extra remaining salary
+        // into their accrued salary
+        uint256 extraSalary = currentSalaryPaid % salary;
+        if (extraSalary > 0) {
+            timeDiff.add(1);
+            employee.accruedSalary = extraSalary;
+        } else if (accruedSalary > 0) {
```
facuspagnuolo: I think we already checked that the accrued salary is greater than zero in L651, we can do an `else` here simply

sohkai: Ah yes, this is to avoid touching the storage at all if `accuredSalary` was 0 from the start.

PR #907, opened 2019-07-07
Payroll: Add payroll remainders as accrued salary

facuspagnuolo:
Merging #903 against master

### DIFF
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -655,7 +655,41 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
             // No need to use SafeMath here as we already know _paymentAmount > accruedSalary
             currentSalaryPaid = _paymentAmount - accruedSalary;
         }
-        employee.lastPayroll = _getLastPayrollDate(_employeeId, currentSalaryPaid);
+        _updateEmployeeLastPayrollDate(_employeeId, currentSalaryPaid);
+    }
+
+    /**
+     * @dev Update the last payroll date for an employee based on the requested payment amount. If the requested amount
+     *      cannot be represented by a multiple of the employee's salary per second, it will be added as accrued salary.
+     * @param _employeeId Employee's identifier
+     * @param _paidAmount Requested amount to be paid to the employee
+     */
+    function _updateEmployeeLastPayrollDate(uint256 _employeeId, uint256 _paidAmount) internal {
+        Employee storage employee = employees[_employeeId];
+
+        uint256 timeDiff = _paidAmount.div(employee.denominationTokenSalary);
+
+        // We check if the division was perfect, and if not, take its ceiling to avoid giving away tiny amounts of
+        // salary and add the remainder to the accrued salary.
+        if (timeDiff.mul(employee.denominationTokenSalary) != _paidAmount) {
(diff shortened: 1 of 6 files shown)


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -652,11 +651,30 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
...
+        uint256 salary = employee.denominationTokenSalary;
+        uint256 timeDiff = currentSalaryPaid.div(salary);
+
+        // If they're being paid an amount that doesn't match perfectly with the adjusted time
+        // (up to a seconds' worth of salary), add the second and put the extra remaining salary
+        // into their accrued salary
+        uint256 extraSalary = currentSalaryPaid % salary;
+        if (extraSalary > 0) {
+            timeDiff = timeDiff.add(1);
+            employee.accruedSalary = salary - currentSalaryPaid;
```
sohkai: @facuspagnuolo Not sure I understand this change in https://github.com/aragon/aragon-apps/pull/903/commits/ae3ee3d298a403fc894f5835ffc91a2c737ce562#diff-9d7adaffbcce19867f027a983a30bcc7R666 (Good catch on the `timeDiff` though!).

We've already calculated the remaining salary with `extraSalary`, and `salary - currentSalaryPaid` looks like it would underflow (`salary` is salary per second, `currentSalaryPaid` is some multiple of that).

future-apps/payroll/test/contracts/Payroll_payday.test.js
```diff
@@ -597,7 +597,7 @@ contract('Payroll payday', ([owner, employee, anyone]) => {
             })
           })
 
-          context('when the employee has a huge salary', () => {
+          context.only('when the employee has a huge salary', () => {
```
sohkai: Oops, let's also revert the `.only` 😄

future-apps/payroll/contracts/Payroll.sol
```diff
@@ -39,7 +39,6 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
     bytes32 constant public MODIFY_PRICE_FEED_ROLE = 0x74350efbcba8b85341c5bbf70cc34e2a585fc1463524773a12fa0a71d4eb9302;
     bytes32 constant public MODIFY_RATE_EXPIRY_ROLE = 0x79fe989a8899060dfbdabb174ebb96616fa9f1d9dadd739f8d814cbab452404e;
 
-    uint128 internal constant ONE = 10 ** 18; // 10^18 is considered 1 in the price feed to allow for decimal calculations
```
sohkai: It's quite weird the changes in #902 are showing up in this diff :/

sohkai: It doesn't show in the commit history though :shrug:

e52aae4bcae478633e57e4106bcbd136f0ad3441

facuspagnuolo: weird :/

future-apps/payroll/contracts/Payroll.sol
facuspagnuolo: True.. I was thinking the case where you request less than a second of your salary, but ofc the scenario you're pointing out is also true. I think the tests are passing because we're using 1 USD per second which will always fit perfectly

PR #911, opened 2019-07-08
Voting: custom path demo

bpierre:
Requires https://github.com/aragon/aragon.js/pull/326

Only a Q&D demo for now, will evolve into the actual implementation after app paths get implemented in the client and aragonAPI.


bpierre:
Closing in favor of https://github.com/aragon/aragon-apps/pull/1029

PR #912, opened 2019-07-08
Payroll: handle both cases of extra salary being sent to accrued balance

sohkai:
Handles both the cases, where someone may:

- Request less than their salary per second
- Request a multiple (and then some) of their salary per second

### DIFF
--- a/future-apps/payroll/contracts/Payroll.sol
+++ b/future-apps/payroll/contracts/Payroll.sol
@@ -660,11 +660,14 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
         // If they're being paid an amount that doesn't match perfectly with the adjusted time
         // (up to a seconds' worth of salary), add the second and put the extra remaining salary
         // into their accrued salary
-        uint256 extraSalary = currentSalaryPaid % salary;
+        // The extra check is to handle the case where someone requested less than one second of their salary
+        uint256 extraSalary = currentSalaryPaid < salary ? salary - currentSalaryPaid : currentSalaryPaid % salary;
         if (extraSalary > 0) {
             timeDiff = timeDiff.add(1);
-            employee.accruedSalary = salary - currentSalaryPaid;
+            employee.accruedSalary = extraSalary;
         } else if (accruedSalary > 0) {
+            // We finally need to clear their accrued salary, but as an optimization, we only do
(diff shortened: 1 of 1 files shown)


future-apps/payroll/contracts/Payroll.sol
```diff
@@ -660,11 +660,14 @@ contract Payroll is EtherTokenConstant, IForwarder, IsContract, AragonApp {
         // If they're being paid an amount that doesn't match perfectly with the adjusted time
         // (up to a seconds' worth of salary), add the second and put the extra remaining salary
         // into their accrued salary
-        uint256 extraSalary = currentSalaryPaid % salary;
+        // The extra check is to handle the case where someone requested less than one second of their salary
+        uint256 extraSalary = currentSalaryPaid < salary ? salary - currentSalaryPaid : currentSalaryPaid % salary;
```
Corsaire: `salary - extraSalary` should go to acrued salary in case when `currentSalaryPaid > salary `

I think the following code should work fine, or am I missing something?
```
uint256 extraSalary = currentSalaryPaid % salary;
if (extraSalary > 0) {
            timeDiff = timeDiff.add(1);
            employee.accruedSalary = salary - extraSalary;
        }
```

Examples:
1. `Salary == 10` and you want to get `currentSalaryPaid == 21`.
Time difference would be 3. Contract will transfer `currentSalaryPaid == 21` and 10 - 21%10 = 9 will go to the accrued salary. 
2. `Salary == 10` and you want to get `currentSalaryPaid == 1`.
Time difference would be 1. Contract will transfer `currentSalaryPaid == 1` and 10 - 1%10 = 9 will go to the accrued salary.

sohkai: Yes... this is a good catch. I think the current state doesn't represent the 

> `currentSalaryPaid == 21` and 10 - 21%10 = 9

case correctly; it'd put `1` into the accrued salary instead of 9.

cc @facuspagnuolo may also want to double check this, but your judgement on this simplification seems correct @Corsaire :).

facuspagnuolo: True, in fact we did that initially [here](https://github.com/aragon/aragon-apps/pull/903/commits/ae3ee3d298a403fc894f5835ffc91a2c737ce562) 😅

PR #913, opened 2019-07-08
Payroll: cover cases for moving extra pay from payday into accrued salary

sohkai:
Modified a few of the tests to try and cover the case we had with `payday()`'s extra remainder salary being moved into the employee's accrued salary.

### DIFF
--- a/future-apps/payroll/test/contracts/Payroll_payday.test.js
+++ b/future-apps/payroll/test/contracts/Payroll_payday.test.js
@@ -307,7 +307,7 @@ contract('Payroll payday', ([owner, employee, anyone]) => {
                   itHandlesPayrollProperlyNeverthelessExtrasOwedAmounts(requestedAmount, currentOwedSalary)
                 })
 
-                context.only('when the requested amount is lower than the total owed salary', () => {
+                context('when the requested amount is lower than the total owed salary', () => {
                   context('when the requested amount represents less than a second of the earnings', () => {
                     const requestedAmount = salary.div(2).floor()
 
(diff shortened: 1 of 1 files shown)


future-apps/payroll/test/contracts/Payroll_payday.test.js
```diff
@@ -385,8 +386,14 @@ contract('Payroll payday', ([owner, employee, anyone]) => {
                   itHandlesPayrollProperlyNeverthelessExtrasOwedAmounts(requestedAmount, totalOwedSalary)
                 })
 
-                context('when the requested amount is greater than the previous owed salary but lower than the total owed', () => {
-                  const requestedAmount = totalOwedSalary.div(2)
+                context('when the requested amount is greater than the previous owed salary but greater than one second of additional salary', () => {
```
facuspagnuolo: Shouldn't this be
```suggestion
                context('when the requested amount is greater than the previous owed salary but lower than one second of additional salary', () => {
```

future-apps/payroll/test/contracts/Payroll_payday.test.js
```diff
@@ -301,20 +309,13 @@ contract('Payroll payday', ([owner, employee, anyone]) => {
 
                 context('when the requested amount is lower than the total owed salary', () => {
                   context('when the requested amount represents less than a second of the earnings', () => {
-                    const requestedAmount = salary.div(2)
+                    const requestedAmount = salary.div(2).floor()
```
facuspagnuolo: This is what I noticed yesterday, after turning the base salary to 1 USD per second, this became flaky

future-apps/payroll/test/contracts/Payroll_payday.test.js
sohkai: Yes, good catch 🙏 !

PR #915, opened 2019-07-10
Survey: Add sample survey kit

facuspagnuolo:
Moving the survey kit from https://github.com/aragon/dao-kits as a sample of the Survey app (similar to what we have for Payroll)


facuspagnuolo:
@sohkai should we carry the changes of https://github.com/aragon/dao-kits/pull/63 as well?

sohkai:
@facuspagnuolo We will also need the deploy scripts (ideally in an accessible manner from the npm scripts) so aragen can easily deploy the kit again.

facuspagnuolo:
@sohkai done!

apps/survey/package.json
```diff
@@ -17,7 +17,15 @@
...
+    "prepublishOnly": "truffle compile --all && npm run abi:extract -- --no-compile",
+    "publish:kit:rpc": "aragon apm publish major $(npm run deploy:kit:rpc | tail -n 1) --environment default",
+    "deploy:kit:rpc": "aragon contracts compile --all && aragon contracts exec --network rpc scripts/ deploy.js",
+    "publish:kit:staging": "aragon apm publish major $(npm run deploy:kit:staging | tail -n 1) --environment staging",
+    "deploy:kit:staging": "aragon contracts compile --all && ENS=0xfe03625ea880a8cba336f9b5ad6e15b0a3b5a939 aragon contracts exec --network rinkeby scripts/ deploy.js",
+    "publish:kit:rinkeby": "aragon apm publish major $(npm run deploy:kit:rinkeby | tail -n 1) --environment rinkeby",
+    "deploy:kit:rinkeby": "aragon contracts compile --all && ENS=0x98Df287B6C145399Aaa709692c8D308357bC085D aragon contracts exec --network rinkeby scripts/ deploy.js",
+    "publish:kit:mainnet": "aragon apm publish major $(npm run deploy:kit:mainnet | tail -n 1) --environment mainnet",
+    "deploy:kit:mainnet": "aragon contracts compile --all && ENS=0x314159265dd8dbb310642f98f50c066173c1259b aragon contracts exec --network mainnet scripts/ deploy.js"
```
sohkai: It would be nice to not have most of these npm scripts, since the example kit is unlikely to be published very many times by us.

cc @0xGabi which scripts does aragen depend on?

apps/survey/contracts/examples/SurveyKit.sol
```diff
@@ -74,23 +81,22 @@ contract SurveyKit is APMNamehash, KitBase {
...
         Survey survey = Survey(dao.newAppInstance(SURVEY_APP_ID, latestVersionAppBase(SURVEY_APP_ID)));
-
-        // Set escapeHatch address as the default vault, in case a token rescue is required
-        dao.setApp(dao.APP_BASES_NAMESPACE(), dao.recoveryVaultAppId(), escapeHatch);
-
         survey.initialize(signalingToken, participation, duration);
 
+        // Install a vault to let it be the escape hatch
+        Vault vault = Vault(dao.newAppInstance(KERNEL_DEFAULT_VAULT_APP_ID, latestVersionAppBase(KERNEL_DEFAULT_VAULT_APP_ID)));
+        vault.initialize();
```
sohkai: We may also want to emit an `InstalledApp` event here

apps/survey/package.json
facuspagnuolo: Sure, I honestly don't know how they are used and if they are used in a certain way actually. I migrated all the scripts related stuff that was in the survey kit repo. Let me know if I should remove those

0xGabi: aragen is only using the `publish:kit:rpc` and `deploy:kit:rpc` but I believe it would be posible to use aragonCLI directly in the bash file ([here](https://github.com/aragon/aragen/blob/master/scripts/publish-templates#L13)) 

My only unknown is how to reference the different template deploy scripts, I believe something like `$SCRIPT_NAME`.

sohkai: > My only unknown is how to reference the different template deploy scripts, I believe something like $SCRIPT_NAME.

Hmm, do you mean which script aragen would use? Perhaps we can standardize on a `deploy_kit.js` or something that will be used by aragen?

0xGabi: Yes exactly. You are right, we should standardize. Will check if that's possible with current `deploy_kit.js`

facuspagnuolo:
@sohkai @0xGabi are we ready to merge this one?

0xGabi:
@facuspagnuolo I think we are ready. I just removed the template scripts cause as @sohkai mentioned they are not going to be used that much by users. 

On `aragen` I have decided to deploy only the new "curated" templates of `dao-template` repo. So for the `aragon-app` repo I don't need the scripts anymore 👍

0xGabi:
Hey I broke the CI but I'm not sure why or where 🤔

PR #918, opened 2019-07-11
Tokens: new style

bpierre:
Preview

## How to test

### First time

```sh
cd ~/projects
```

```sh
cd ~/projects/aragon-ui
npm link
```

### Getting updates

```sh
cd ~/projects/aragon-ui
git checkout newstyle
git pull origin newstyle
npm install

cd ~/projects/aragon
git checkout newstyle
git pull origin newstyle
npm install
npm link @aragon/ui

cd ~/projects/aragon-apps/apps/token-manager/app
git checkout newstyle-tokens
git pull origin newstyle-tokens
npm install
npm link @aragon/ui
```

### Running it

```sh
cd ~/projects/aragon-apps/apps/token-manager/app
npm start
```

```sh
cd ~/projects/aragon
```

## Depends on

https://github.com/aragon/aragon-ui/pull/391
https://github.com/aragon/aragon/pull/826


apps/token-manager/app/src/App.js
```diff
@@ -86,100 +98,151 @@ class App extends React.PureComponent {
...
   }
 }
 
 export default () => {
   const { api, appState, connectedAccount, requestMenu } = useAragonApi()
+  const theme = useTheme()
+  const themeMode = useThemeMode()
+  const { layoutName } = useLayout()
+
+  useEffect(() => {
```
AquiGorka: Let's implement a hook for this, how does `useThemeSwitcher(value => themeMode.set(value))` sound?

apps/token-manager/app/src/App.js
```diff
@@ -86,100 +98,151 @@ class App extends React.PureComponent {
...
+        </SidePanel>
+      </IdentityProvider>
     )
   }
 }
 
 export default () => {
   const { api, appState, connectedAccount, requestMenu } = useAragonApi()
+  const theme = useTheme()
+  const themeMode = useThemeMode()
```
AquiGorka: I've been maintaining a consistent interface to what hooks expose, when it is a handler prepend it with `handle` -> `handleDoSomething` instead of just exposing the inner `useState` method to update the state var, which we usually prepend with `set`.
`const { set: handleThemeModeChange } = useThemeMode()`
What do you think?

apps/token-manager/app/src/App.js
```diff
@@ -86,100 +98,151 @@ class App extends React.PureComponent {
...
-                assignTokensConfig.mode === 'assign'
-                  ? 'Add tokens'
-                  : 'Remove tokens'
-              }
-              opened={sidepanelOpened}
-              onClose={this.handleSidepanelClose}
-              onTransitionEnd={this.handleSidepanelTransitionEnd}
+              <h1
+                css={`
+                  ${textStyle(layoutName === 'small' ? 'title3' : 'title2')};
```
AquiGorka: Hmm, is `layoutName === 'small'` the same as `below('medium')`? I like having more utils but maybe we should consistently only use one no? To avoid having to update two places in case something changes and having to maintain these in sync?

AquiGorka: On a second idea, the styles for the `Header` `primary` prop (`title2` & `title2`), shouldn't these be set at component level? You mentioned there should only be one `Header` as it is implemented so I'd assume it takes care of the styling too

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -1,166 +1,163 @@
...
+  tokenDecimalsBase,
+  tokenName,
+  tokenSupply,
+  tokenSymbol,
+  tokenTransfersEnabled,
+  userAccount,
+}) {
+  const network = useNetwork()
+  return (
+    <Split
```
AquiGorka: Is this the magical component that structures the 1/2 column layout when the viewport size changes? Nice! ✨ ✨

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -1,166 +1,163 @@
...
+      primary={
+        <DataView
+          mode="table"
+          fields={groupMode ? ['Owner'] : ['Holder', 'Balance']}
```
AquiGorka: What is this `groupMode` prop for? I see it defaults to `false` and don't see any handler to update it?

apps/token-manager/app/src/App.js
```diff
@@ -86,100 +98,151 @@ class App extends React.PureComponent {
...
 }
 
 export default () => {
   const { api, appState, connectedAccount, requestMenu } = useAragonApi()
+  const theme = useTheme()
+  const themeMode = useThemeMode()
+  const { layoutName } = useLayout()
+
+  useEffect(() => {
+    const fn = ({ data }) => {
```
AquiGorka: Super nitpick here: `useCallback` on the handler function

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -1,166 +1,163 @@
...
+  onRemoveTokens,
+  singleToken,
+  canAssign,
+}) {
+  const theme = useTheme()
+  const [label, showLocalIdentityModal] = useIdentity(address)
 
-const ResponsiveTable = styled(Table)`
-  margin-top: 16px;
+  const editLabel = useCallback(() => showLocalIdentityModal(address), [
```
AquiGorka: Nitpick for consistency: all these handlers, could you update the names to `handleActionsToDo`?

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -1,166 +1,163 @@
...
 
-  ${breakpoint(
-    'medium',
-    `
-      opacity: 1;
-      margin-top: 0;
-    `
-  )};
-`
+  const actions = [
```
AquiGorka: Super nitpick here: 
```suggestion
  const actions = [
    ...[canAssign ? [assignTokens, IconAdd, 'Add tokens'] : []]
```

Using `unshift` below is mutating the contents, nothing big, but if we allow it here, what stops us to do other operations on it elsewhere (or maybe some other part of the code) and I try to avoid it by setting everything in one go.

You can disregard if this is not a concern for you, I just wanted to let you know my line of thinking.

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -1,166 +1,163 @@
...
+                {isCurrentUser && <You />}
+              </React.Fragment>,
+            ]
 
-    return (
-      <Viewport>
-        {({ below }) => {
-          const tabbedNavigation = below('medium')
-          const compactTable = below('medium')
+            if (balance) {
```
AquiGorka: I see you used the same pattern here, in case my other comment resonated and you updated I'd suggest you do the same here, if not, then nothing to see here.

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -1,166 +1,163 @@
...
-    } = this.props
-    const { selectedTab } = this.state
+            const values = [
+              <React.Fragment>
+                <LocalIdentityBadge
+                  entity={address}
+                  networkType={network.type}
+                  connectedAccount={isCurrentUser}
+                />
+                {isCurrentUser && <You />}
```
AquiGorka: `You` is not defined.

apps/token-manager/app/src/components/InfoBoxes.js
```diff
@@ -0,0 +1,209 @@
...
+  useTheme,
+} from '@aragon/ui'
+import { useNetwork } from '@aragon/api-react'
+import { formatBalance, stakesPercentages } from '../utils'
+import You from './You'
+import LocalIdentityBadge from './LocalIdentityBadge/LocalIdentityBadge'
+
+const DISTRIBUTION_ITEMS_MAX = 7
+
+const displayedStakes = (accounts, total, colors) => {
```
AquiGorka: Should this function be memoized?

apps/token-manager/app/src/components/InfoBoxes.js
```diff
@@ -0,0 +1,209 @@
...
+      holders,
+      network,
+      tokenAddress,
+      tokenDecimalsBase,
+      tokenName,
+      tokenSupply,
+      tokenSymbol,
+      userAccount,
+      theme,
+      ...rest
```
AquiGorka: You were probably using this in `dev`, not needed anymore.

apps/token-manager/app/src/components/InfoBoxes.js
```diff
@@ -0,0 +1,209 @@
...
+        <Box heading="Token Info">
+          <ul
+            css={`
+              color: ${theme.surfaceContent};
+            `}
+          >
+            {[
+              [
+                'Total supply',
+                <strong>
```
AquiGorka: Should this be a `textStyle`?

apps/token-manager/app/src/components/InfoBoxes.js
```diff
@@ -0,0 +1,209 @@
...
+                css={`
+                  display: flex;
+                  justify-content: space-between;
+                  list-style: none;
+                  & + & {
+                    margin-top: ${2 * GU}px;
+                  }
+
+                  > span:nth-child(1) {
+                    font-weight: 400;
```
AquiGorka: Should this be a `textStyle`?

apps/token-manager/app/src/components/InfoBoxes.js
```diff
@@ -0,0 +1,209 @@
...
+                  > span:nth-child(3) {
+                    flex-shrink: 1;
+                  }
+                  > strong {
+                    text-transform: uppercase;
+                  }
+                `}
+              >
+                <span>{label}</span>
+                <span>:</span>
```
AquiGorka: This is not showing, is it on purpose? The `opacity: 0` is the how, but maybe there's a why?

apps/token-manager/app/src/App.js
bpierre: > Hmm, is layoutName === 'small' the same as below('medium')?

Not exactly, it also:

- Takes into consideration the spacing around a given layout.
- More importantly, can take another width than the viewport width (which makes it work on non-iframe apps too).

See the implementation: https://github.com/aragon/aragon-ui/blob/newstyle/src/components/Layout/Layout.js#L13-L24

> I like having more utils but maybe we should consistently only use one no? To avoid having to update two places in case something changes and having to maintain these in sync?

Yes the idea is to move to `Layout`, except when the actual viewport dimensions are needed.

>  On a second idea, the styles for the Header primary prop (title2 & title2), shouldn't these be set at component level? You mentioned there should only be one Header as it is implemented so I'd assume it takes care of the styling too

Yes it’s something that remains to be solved… [styles are applied](https://github.com/aragon/aragon-ui/blob/newstyle/src/components/Header/Header.js#L33-L44), except if a React node is passed, in which case we can’t apply the styles since we don’t control the component. There are different ways to address this (like applying default CSS styles, which might be surprising since we don’t do that anywhere else), but I wanted to let it some time before going for one solution or another.

apps/token-manager/app/src/App.js
bpierre: It sounds like I need to remove this part entirely haha  :see_no_evil:  It was for the theme switching experimentation and I shouldn’t have committed it sorry!

apps/token-manager/app/src/App.js
bpierre: Same than above, I’ll remove this part.

apps/token-manager/app/src/screens/Holders.js
bpierre: It is indeed! :laughing: https://github.com/aragon/aragon-ui/blob/newstyle/src/components/Split/Split.js

apps/token-manager/app/src/App.js
AquiGorka: It's alright, it gives us an idea of how this will have to be implemented and top level hook here makes sense.

apps/token-manager/app/src/screens/Holders.js
bpierre: What’s not to love about some well contained mutations? :laughing: I know this can be a controversial subject, but I think that deciding to use local immutability, like imperative vs. functional code, should also take into consideration readability and performances.

Having pure functions and components is a big advantage, but trying to achieve immutability everywhere often comes at a readability and performance cost with JavaScript, so I don’t really have a problem with variable mutations as long as it doesn’t have any side effects outside of the current scope. On this regard, I like how Immer is fully embracing this idea: https://medium.com/hackernoon/introducing-immer-immutability-the-easy-way-9d73d8f71cb3

Now about this specific case: I find them both readable, and the performance cost is not relevant here, so let’s go for your suggestion :+1:

apps/token-manager/app/src/screens/Holders.js
bpierre: It comes from the app state reducer => https://github.com/aragon/aragon-apps/blob/newstyle-tokens/apps/token-manager/app/src/app-state-reducer.js#L44

apps/token-manager/app/src/App.js
bpierre: I’ve been using the `handleX` less and less since we moved to hooks.

With class components, we couldn’t pass methods declared on the class prototype to event handlers, since we needed `this` to be set properly, so we often had these arrow functions being used as wrappers, and often directly setting the state or other things. It is also quite frequent to see these methods doing several things in reaction to a particular event coming from a children, making the “handle + initiator + event” a good name.

With hooks, I find it to be too different to apply the same convention: the callbacks we create are often focused on are often focused in one task, and reused in different cases (e.g. called from another hook), and calling a function named “handleClick” from something else than a click event doesn’t seem right… so I tend to go for a name that reflect “what it does”, rather than “what it should be called from” now.

apps/token-manager/app/src/screens/Holders.js
bpierre: Oops :+1:

apps/token-manager/app/src/App.js
AquiGorka: > With hooks, I find it to be too different to apply the same convention: the callbacks we create are often focused on are often focused in one task

```
const { handleNameChange } = useName()
const handleEffectOnNameChange = () => {
  /* side effect */
  handleNameChange()
}
<Component onChange={handleNameChange} />
<OtherComponentTriggersSideEffect onChange={handleEffectOnNameChange} />
```

Sometimes one function yes but we can implement in a way that does not take away from the `handleX` format no?

> and reused in different cases (e.g. called from another hook), and calling a function named “handleClick” from something else than a click event doesn’t seem right… so I tend to go for a name that reflect “what it does”, rather than “what it should be called from” now

I don't think I'd call a method exported from a hook `handleClick` but I would pass it down to a component on the `onClick` handler

Using a "name that reflects what it does" is exactly what I want to use the `X` for, and just preprend it with `handle` makes it explicit, the method is the handler.

I'd rather we stick to a convention that's widely used throughout the code base instead of switching randomly to another one - again as with all team conventions, I call forth the team so that we can decide as a group cc/ @sohkai @2color

bpierre: I’m not sure I follow :thinking: the `handle` prefix was used to indicate that a function was handling an event, but in your examples it is the other way? e.g. `handleNameChange` is not “handling the name change event”, which is how we usually use these prefixes.

apps/token-manager/app/src/components/InfoBoxes.js
bpierre: Yes absolutely! (I’ll memoize its result from the component function)

apps/token-manager/app/src/components/InfoBoxes.js
bpierre: It’s coming from the previous version and passed to the wrapping component, but it doesn’t seem like it was ever being used indeed :+1:

apps/token-manager/app/src/components/InfoBoxes.js
bpierre: Accessibility :-)

apps/token-manager/app/src/App.js
AquiGorka: Let me take a step backwards then:

```
const useName = () => {
const [name, setName]=useState('')
// either (which would make sense in a hook named more like useNameWithInput or what not
return { handleNameChange: ({ target: value }) => setName(value) }
// or depending on the interface that you may want to provide
return { handleNameChange: setName }
}
```

I do understand there is a difference between the two names and this is definitely an over-simplified example, but the idea is that instead of sending out a `setX` from a hook, we can continue using the `handle` prefix as before to refer to the method that will take action when/if the state var changes.

sohkai: Not quite sure what the React community consensus might be, but for me the distinction for `handleX` is that it's going to handle an event or action from one of its children.

I think `setX` being returned from props is the original intention, and the API we return from a hook doesn't necessarily have to be paired with its internal state.

E.g.

```js
const useName = () => {
const [name, setName]=useState('')
// either (which would make sense in a hook named more like useNameWithInput or what not
return { setNameFromEvent: ({ target: value }) => setName(value) }
// or depending on the interface that you may want to provide
return { setName }
}
```

apps/token-manager/app/src/App.js
```diff
@@ -86,100 +98,151 @@ class App extends React.PureComponent {
...
 
     return (
-      <Main assetsUrl="./aragon-ui">
-        <div css="min-width: 320px">
-          <IdentityProvider
-            onResolve={this.handleResolveLocalIdentity}
-            onShowLocalIdentityModal={this.handleShowLocalIdentityModal}
-          >
-            <SyncIndicator visible={isSyncing} />
-            <AppLayout
```
sohkai: Should we also remove `AppLayout.js` and this import?

apps/token-manager/app/src/components/InfoBoxes.js
```diff
@@ -0,0 +1,209 @@
...
+                  networkType={network.type}
+                />,
+              ],
+            ].map(([label, content], index) => (
+              <li
+                key={index}
+                css={`
+                  display: flex;
+                  justify-content: space-between;
+                  list-style: none;
```
sohkai: Should we just set this on the `ul`?

apps/token-manager/app/src/components/InfoBoxes.js
```diff
@@ -0,0 +1,209 @@
...
+        </Box>
+        <Box heading="Ownership Distribution">
+          <p
+            css={`
+              ${textStyle('body2')}
+            `}
+          >
+            Token Holder Stakes
+          </p>
+          <div
```
sohkai: Maybe it's worth making this into its own component now, perhaps with a `legend: PropTypes.bool` or `renderLegend: PropTypes.func` prop that controls whether the names are listed at the end?

apps/token-manager/app/src/App.js
```diff
@@ -86,100 +98,151 @@ class App extends React.PureComponent {
...
+            tokenSupply={tokenSupply}
+            tokenSymbol={tokenSymbol}
+            tokenTransfersEnabled={tokenTransfersEnabled}
+            userAccount={connectedAccount}
+            onAssignTokens={this.handleLaunchAssignTokens}
+            onRemoveTokens={this.handleLaunchRemoveTokens}
+          />
+        ) : (
+          !isSyncing && (
+            <EmptyState onActivate={this.handleLaunchAssignTokensNoHolder} />
```
sohkai: We should probably double check this and give it a light update @dizzypaty (super low priority; no normal user should ever see this empty state on the Tokens app)

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -1,166 +1,163 @@
...
+      }
+    />
+  )
+}
 
-  handleSelectTab = index => {
-    this.setState({ selectedTab: index })
-  }
+Holders.propTypes = {
+  holders: PropTypes.array,
```
sohkai: Hehe, maybe this is a good time to just list all of the props (and just declare them all to be required)?

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -1,166 +1,163 @@
...
+  return (
+    <Split
+      primary={
+        <DataView
+          mode="table"
+          fields={groupMode ? ['Owner'] : ['Holder', 'Balance']}
+          entries={holders.map(({ address, balance }) =>
+            groupMode ? [address] : [address, balance]
+          )}
+          renderEntry={([address, balance]) => {
```
sohkai: One thing that I'm not sure whether it's possible to style, is that the [balance in the designs](https://www.figma.com/file/M9AbLkUkeRCFKQt6qdFUsCks/Tokens?node-id=1%3A2) is right-aligned while currently the app has it center-aligned

apps/token-manager/app/src/App.js
AquiGorka: > a hook doesn't necessarily have to be paired with its internal state
Sometimes for sure, when it is a generic hook we expose, but when it is a hook directly related to a component (or component tree) I do think it makes sense to be explicit about the handlers's names.

apps/token-manager/app/package.json
```diff
@@ -6,7 +6,7 @@
   "dependencies": {
     "@aragon/api": "^2.0.0-beta.4",
     "@aragon/api-react": "^2.0.0-beta.4",
-    "@aragon/ui": "^0.40.1",
+    "@aragon/ui": "^1.0.0-alpha.7",
```
AquiGorka: Reminder: we might need to come back here.

apps/token-manager/app/src/components/InfoBoxes.js
bpierre: Yes I removed it, it’s already in the text style :-)

apps/token-manager/app/src/App.js
bpierre: Yes!

apps/token-manager/app/src/components/InfoBoxes.js
bpierre: Would it change anything? Only one class is generated by babel-plugin-styled-components, and CSS wise, they would still apply to every `<li>` via inheritance from the `<ul>`  (`list-style` is specific to `<li>`).

I could move the `color` declaration from the `ul` to the `li` instead, to have all the CSS will be in one place?

apps/token-manager/app/src/components/InfoBoxes.js
bpierre: Yes!

bpierre: :point_right: https://github.com/aragon/aragon-ui/pull/435

apps/token-manager/app/src/screens/Holders.js
bpierre: I added them, but I didn’t make them all required because many stay `undefined` until the data starts to be retrieved from the script.

apps/token-manager/app/src/screens/Holders.js
bpierre: It was a spacing bug in DataView, thanks for having noticed it :hugs: https://github.com/aragon/aragon-ui/pull/391/commits/1e82aec0ae52b29277e8ff2cbe0f4739b20cf79f

apps/token-manager/app/src/components/InfoBoxes.js
bpierre: Maybe yes, we could have stylistic variants for some of the text styles, like `body2-bold`, which could be useful to have directly. Let’s discuss it with the design team.

For the `strong` element itself, I think we should keep it as it also carries a semantic value.

apps/token-manager/app/src/screens/Holders.js
bpierre: If it can add something to the [other discussion](https://github.com/aragon/aragon-apps/pull/918#discussion_r305695753):

I don’t consider these to be handle any event, even if they happen to be used as such later on. When a function is meant to handle an event specifically, I usually name it using the `handle` prefix with the emitted event − not the action it triggers.

For example here, a `handle`-prefixed name I could use would be `handleClick()`, since they “handle a click” (which would be a problem since they are three :laughing:). But they are not really specific to the click event, so it could be limiting to do that (e.g. if we want to call it from somewhere else at some point).

`handleEditLabel` wouldn’t work following the same logic, because the function doesn’t “handle an edit”, but kind of the opposite: it triggers an edit.

apps/token-manager/app/src/components/InfoBoxes.js
sohkai: Ahh, I had no idea `list-style` was specific to `<li>`; I had always assumed it to be a `<ul>` concern but it makes sense to be able to individually target different `<li>`s with different `list-style`s.

apps/token-manager/app/src/hooks.js
```diff
@@ -76,26 +73,20 @@ export function useArrowKeysFocus(query, containerRef = useRef()) {
...
-    [highlightedIndex, query]
-  )
-  useEffect(
-    () => {
-      document.addEventListener('keydown', handleKeyDown)
-      return () => document.removeEventListener('keydown', handleKeyDown)
-    },
-    [handleKeyDown]
-  )
+  useEffect(() => {
```
AquiGorka: By updating to `_AutoComplete` from `aragonUI` these hooks will become deprecated

apps/token-manager/app/src/screens/Holders.js
AquiGorka: `handleClickForEdit`?

apps/token-manager/app/src/hooks.js
sohkai: Done in 1f492cf.

sohkai:
@bpierre I've pushed a few commits that add a bit of polish and update various deprecated things in aragonUI. Easiest to review commit-by-commit.

bpierre:
@sohkai All good! :checkered_flag: :pray:

sohkai:
@bpierre I've applied a fix for group mode in 88ec435 (test via [this org](https://preview.aragon.org/#/0xe520428C232F6Da6f694b121181f907931fD2211/0x3c16dc46b84a6647f8375235ca88dad2c27edb8b)).

However, the `DataView` layout is a bit funny when there's just one column:

Screen Shot 2019-08-11 at 5 36 43 AM

Is there a way we can align the column to the left?

bpierre:
@sohkai Thanks!

I fixed the issue with `DataView` and single fields https://github.com/aragon/aragon-ui/pull/449

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -1,166 +1,171 @@
...
+            return values
+          }}
+          renderEntryActions={([address, balance]) => (
+            <EntryActions
+              address={address}
+              balance={balance}
+              onAssignTokens={onAssignTokens}
+              onRemoveTokens={onRemoveTokens}
+              singleToken={!balance || balance.eq(tokenDecimalsBase)}
+              canAssign={!!balance && balance.lt(maxAccountTokens)}
```
bpierre: I don’t remember discussing it, but I generally prefer to use `Boolean()`, `String()` and `Number()` for type conversions, rather than relying on the usual equivalents relying on type coercion rules (`!!`, `+""` and `-0`), for readability.

```suggestion
              canAssign={Boolean(balance) && balance.lt(maxAccountTokens)}
```

or

```suggestion
              canAssign={Boolean(balance && balance.lt(maxAccountTokens))}
```

apps/token-manager/app/src/screens/Holders.js
bpierre: I don’t think we should limit it to handling clicks :laughing: I think it is more important to name this function after what it does (“edit a label”), rather than its usage (being passed to `onClick`). There is nothing related to handling a click event in this function, it might be called from anywhere.

apps/token-manager/app/src/screens/Holders.js
sohkai: I've since simplified this logic and made it based on `groupMode` in 8ba0f83.

apps/token-manager/app/src/screens/Holders.js
AquiGorka: > it might be called from anywhere

This is what I'm pointing at, it might, but it is not, it is being used in a click handler. When the methods are specific to an implementation it doesn't need a generic name, if/when it responds to some other type of event maybe the abstract name should be updated but for now, only one use case: click.

AquiGorka:
@sohkai and @bpierre the scrolling issue happens here too, is it something we plan on fixing for this iteration or can it be pushed to a second iteration
[Ref](https://www.loom.com/share/13ad103fe60c442d8f559de5255eeb74)

sohkai:
@AquiGorka Since it happens everywhere, and ideally this should come through an aragonUI upgrade, I think we can do it separately?

sohkai:
@bpierre Pushed a couple commits, to squeeze out the final bits I noticed.

Adds an updated empty state that can be improved later (again, most users should **never** see this):

Screen Shot 2019-08-16 at 10 40 14 AM

--------

Depending on how https://github.com/aragon/aragon-ui/pull/486 goes I have another commit updating this app's usage of the ContextMenu.

PR #921, opened 2019-07-13
chore: upgrade solidity-coverage to v0.6+

sohkai:
cc @facuspagnuolo Let's see how long coverage takes now :).


future-apps/payroll/package.json
```diff
@@ -59,9 +59,9 @@
     "@aragon/cli": "~5.6.0",
     "@aragon/test-helpers": "^2.0.0",
     "eth-gas-reporter": "^0.2.0",
-    "ethereumjs-testrpc-sc": "^6.1.6",
+    "ethereumjs-testrpc-sc": "^6.4.5-sc.3",
```
facuspagnuolo: since the new `solidity-coverage` version is using ganache, do we still need this one?

sohkai: I think so, the reason was that `lerna` hoisting makes running `testrpc-sc` a bit of a pain unless we explicitly declare this dependency here.

facuspagnuolo: I see, but I we shouldn't need to run it anymore right?

sohkai: I think we still have to; the `ganache-cli.sh` script runs `testrpc-sc` and AFAIK coverage still requires the modified version of testrpc.

sohkai:
So... looking into this a bit more, unfortunately there's no good way to get Payroll to compile on a coverage environment.

`solidity-coverage` preprocesses all copied npm dependencies, and in our case, this leads to some mismatches between skipped test files and their dependencies (specifically, the `ERC20` from `@aragon/os` and `MaliciousToken`).

`solidity-coverage` has said recently that they'll be moving away from requiring any preprocessing step, and we can revisit this once that's ready. It's also highly unlikely at the current moment that Payroll's coverage tests would run without requiring us to split them up, like we did for normal tests :).

PR #927, opened 2019-07-17
Travis: move Payroll's payday() tests into its own integration test stage

sohkai:
Moves this long-running (1h+) test into its own stage, so that it doesn't block the rest of the coverage tests from running.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ jobs:
       name: "Agent"
     - script: npm run test:finance
       name: "Finance"
-    - script: npm run test:payroll:only:payday
+    - script: npm run test:payroll:except:payday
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -68,5 +66,9 @@ jobs:
       name: "Vault"
     - script: npm run coverage:voting
       name: "Voting"
+
+    - stage: integration
+      script: npm run test:payroll:except:payday
+      name: "Payroll payday"
```
facuspagnuolo: This file is a bit confusing, the `name` key corresponds to the `script` right above haha
I think you wanted to move `npm run test:payroll:payday` to a separate stage

sohkai: Yes... good catch!

PR #929, opened 2019-07-18
Finance: refactor to new styles for 0.8

AquiGorka:
image

## How to test

### First time

```sh
cd ~/projects
```

```sh
cd ~/projects/aragon-ui
npm link
```

### Getting updates

```sh
cd ~/projects/aragon-ui
git checkout newstyle
git pull origin newstyle
npm install

cd ~/projects/aragon
git checkout newstyle
git pull origin newstyle
npm install
npm link @aragon/ui

cd ~/projects/aragon-apps/apps/finance/app
git checkout newstyle/finance
git pull origin newstyle/finance
npm install
npm link @aragon/ui
```

### Running it

```sh
cd ~/projects/aragon-apps/apps/finance/app
npm start
```

```sh
cd ~/projects/aragon
```


apps/finance/app/src/components/Transfers.js
```diff
@@ -161,143 +158,184 @@ const Transfers = React.memo(({ dao, tokens, transactions }) => {
...
+              onTokenChange={handleTokenChange}
+              transferTypeFilter={selectedTransferType}
+              onTransferTypeChange={handleTransferTypeChange}
+              compactMode={compactMode}
+              symbols={symbols}
+              transferTypes={TRANSFER_TYPES_STRING}
+            />
           )}
-        </div>
+          {emptyResultsViaFilters && (
```
sohkai: cc @bpierre the location of this (in the header, to avoid the table headers) is kind of surprising. Maybe we should add an "empty" mode to `DataView`?

apps/finance/app/src/components/Transfers.js
```diff
@@ -161,143 +158,184 @@ const Transfers = React.memo(({ dao, tokens, transactions }) => {
...
+            css={`
+              font-weight: 600;
+              color: ${isIncoming ? newTheme.positive : newTheme.negative};
+            `}
+          >
+            {formattedAmount} {symbol}
+          </span>,
+        ]
+      }}
+      renderSelectionCount={count =>
```
sohkai: Can we remove the selection related logic now?

apps/finance/app/src/components/Transfers.js
```diff
@@ -161,143 +158,184 @@ const Transfers = React.memo(({ dao, tokens, transactions }) => {
...
+        </ContextMenu>
       )}
-    </section>
+    />
   )
 })
 
 Transfers.propTypes = {
-  dao: PropTypes.string.isRequired,
+  dao: PropTypes.string,
```
sohkai: Technically this is still required, but I see how you would get errors because we're missing an empty state to avoid rending `Transfers` if not enough information is available.

apps/finance/app/src/components/Transfers.js
```diff
@@ -161,143 +158,184 @@ const Transfers = React.memo(({ dao, tokens, transactions }) => {
...
+        const { symbol, decimals } = tokenDetails[toChecksumAddress(token)]
+        const formattedAmount = formatTokenAmount(
+          amount,
+          isIncoming,
+          decimals,
+          true,
+          { rounding: 5 }
+        )
+
+        return [
```
sohkai: It's nice that we don't have to use a fragment, but the array syntax always scares me a bit (always hard to figure out where to put the `,`, and it's harder to read).

Does it work if we wrap it in a fragment?

apps/finance/app/src/components/NewTransfer/Deposit.js
```diff
@@ -64,7 +64,7 @@ const initialState = {
       loading: false,
     },
     error: NO_ERROR,
-    index: 0,
+    index: -1,
     value: '',
   },
 }
```
sohkai: @dizzypaty Is this the intended look of this warning message?

Screen Shot 2019-07-29 at 1 10 40 PM

apps/finance/app/src/components/NewTransfer/TokenSelector.js
```diff
@@ -90,16 +86,18 @@ class TokenSelector extends React.Component {
...
         <Field label={label}>
           <DropDown
+            label="Select a token"
             items={items}
-            active={activeIndex}
+            selected={activeIndex}
             onChange={this.handleChange}
             required
             wide
+            width="390px"
```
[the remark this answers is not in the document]
AquiGorka: We can't :(
The new `DropDown` is using the `Popover` component and unless we specify a width it will wrap to it's contents size

apps/finance/app/src/components/DateRange/DateRangeInput.js
```diff
@@ -158,29 +191,53 @@ class DateRangeInput extends React.PureComponent {
...
+            : `1px solid ${border}`};
+          border-radius: ${RADIUS}px;
+        `}
         ref={el => (this.rootRef = el)}
         onClick={this.handleClick}
       >
-        <StyledTextInput
+        <Labels text={this.getValueText()} />
+        <TextInput
+          css="width: 28ch;"
```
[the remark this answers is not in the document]
AquiGorka: `Labels` was a q&d fix until we re-implement the whole component, the input is still used below the labels and it needs this width to show the Calendar icon in the correct place.

apps/finance/app/src/components/Transfers.js
```diff
@@ -161,143 +158,184 @@ const Transfers = React.memo(({ dao, tokens, transactions }) => {
...
-                Show Older Transfers
+            <div
+              css={`
+                color: ${theme.content};
+                ${textStyle('body1')}
+              `}
+            >
+              Transfers
+            </div>
+            <div css="text-align: right;">
```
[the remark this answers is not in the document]
AquiGorka: Yep, if the grid-item has more available space it will align its contents to the left.

apps/finance/app/src/components/Transfers.js
AquiGorka: I don't think there is a way to hide the headers currently (please confirm @bpierre ) so I added this quick fix but this should definitely be refactored.

bpierre: Yes! It could be used for the filters, when there is no data at all, and when the data is loading (we might want to separate this state though).

@owisixseven @dizzypaty What would you think of providing the default loading state and empty states for DataView?

We already have the “no results for these filters”:

image

And we could add two variants: “no data” and “loading data”. It would only be the default, we could still let app authors change the illustration, the content or even everything if needed.

bpierre: Oops sorry @AquiGorka GitHub didn’t show your message until I posted mine.

> I don't think there is a way to hide the headers currently (please confirm @bpierre ) so I added this quick fix but this should definitely be refactored.

Ah no we can’t hide them… an immediate solution would be to render the filters in an empty state component when there are no results, but I think we should provide this empty state screen from DataView directly, as discussed above.

apps/finance/app/src/components/NewTransfer/TokenSelector.js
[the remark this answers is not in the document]
sohkai: This sounds like a problem we should solve with the `Dropdown`; I think we’ll want the `wide` behaviour quite often (e.g. in most panels with drop down inputs).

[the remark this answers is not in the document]
sohkai: This sounds like a problem we should solve with the `Dropdown`; I think we’ll want the `wide` behaviour quite often (e.g. in most panels with drop down inputs).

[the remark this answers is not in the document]
AquiGorka: Yeah, we might need to figure out the node's width via `js` and then set the corresponding value for `Popover`s container. Will add an issue for this.

apps/finance/app/src/components/BalanceToken.js
```diff
@@ -13,68 +12,58 @@ const splitAmount = amount => {
...
+            height="20"
+            src={`https://chasing-coins.com/coin/logo/${symbol}`}
+          />
+        )}
+        {symbol || '?'}
+      </div>
+      <div>
+        <div
+          css={`
+            .fractional {
```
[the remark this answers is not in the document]
AquiGorka: Seems so: https://github.com/aragon/aragon-apps/blob/e8fda94a90d81757e2cbd5fcaad5116cb80822aa/apps/finance/app/src/components/BalanceToken.js#L10

apps/finance/app/src/components/DateRange/DateRangeInput.js
```diff
@@ -158,29 +191,53 @@ class DateRangeInput extends React.PureComponent {
...
+          border: ${startDateProps && endDateProps
+            ? `1px solid ${accent}`
+            : `1px solid ${border}`};
+          border-radius: ${RADIUS}px;
+        `}
         ref={el => (this.rootRef = el)}
         onClick={this.handleClick}
       >
-        <StyledTextInput
+        <Labels text={this.getValueText()} />
```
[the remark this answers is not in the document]
AquiGorka: Yes, this is something I expect to be solved/refactored/addressed when we implement this component in `aragonUI`

apps/finance/app/src/components/Transfers.js
AquiGorka: It does not work that way (correct me if I'm wrong here @bpierre) I share the concern but this is a way to set the content per each column.

bpierre: Yes this is an array of separated nodes that correspond to each field, and can be rendered in the context of a table row or a list block.

We could also accept an object here, using the field labels as IDs:

```jsx
<DataView
  renderEntry={entry => {
    return {
      'Date': <Date />,
      'Source/recipient': <Source />,
      'Reference': <Reference />,
      'Amount': <Amount />,
    }
  }}
/>
```

We could also have an optional ID on the fields, to make it a bit nicer:

```jsx
<DataView
  fields={[
    { id: 'date', label: 'Date' },
    { id: 'source', label: 'Source/recipient' },
    { id: 'reference', label: 'Reference' },
    { id: 'amount', label: 'Amount' },
  ]}
  renderEntry={entry => {
    return {
      date: <Date />,
      source: <Source />,
      reference: <Reference />,
      amount: <Amount />,
    }
  }}
/>
```

But having optional IDs might bring too much confusion for users and I would rather make them mandatory in that case… which would also make simple cases more verbose by removing the possibility to use strings directly for fields.

bpierre: Actually we could accept both styles: a function returning an array of elements, or an array of components.

apps/finance/app/src/components/BalanceToken.js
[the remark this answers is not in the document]
sohkai: Let's move the `.fractional` styles just to that component then 😅

apps/finance/app/src/App.js
```diff
@@ -85,90 +88,73 @@ class App extends React.Component {
...
-                transactions.length === 0 && (
-                  <EmptyScreen>
-                    <EmptyStateCard
-                      icon={<img src={addFundsIcon} alt="" />}
-                      title="There are no funds yet"
-                      text="Create a new transfer to get started."
-                      actionText="New transfer"
-                      onActivate={this.handleNewTransferOpen}
-                    />
-                  </EmptyScreen>
```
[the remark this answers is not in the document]
dizzypaty: We have defined one, but I don't think it was implemented in the DataView component yet. Will sync with @bpierre on this.

apps/finance/app/src/components/NewTransfer/Deposit.js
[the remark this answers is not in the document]
dizzypaty: We have designs for the buttons disabled states and Pierre is currently working on them to make them available on aragonUI

dizzypaty: Obviously not 😂

We've defined the warning component appearance here: https://www.figma.com/file/WN8Yuc544XoH5NosMBPclTBx/Info-and-warning-components?node-id=0%3A1

dizzypaty: Also, we don't display icons anymore and in case of displaying a title, it should use the text style `LABEL 2`. (Which is not the case on the designs for this screen).

FINANCE_TRANSFER

bpierre: Just as a note, the `Info` update merged a few hours ago takes care of that, including the text style if `title` is defined:

```jsx
<Info mode="warning">
  This organization doesn’t have any tokens available to withdraw
</Info>
```

```jsx
<Info mode="warning" title="Action impossible">
  This organization doesn’t have any tokens available to withdraw
</Info>
```

apps/finance/app/src/components/Transfers.js
bpierre: I was just thinking of another option so I’m adding it here as well.

We could be to move these render functions / components into `fields`:

```jsx
<DataView
  fields={[
    { label: 'Date', render: Date },
    { label: 'Source/recipient', render: Source },
    { label: 'Reference', render: Reference },
    { label: 'Amount', render: Amount },
  ]}
/>
```

And we could still have a concise mode:

```jsx
<DataView
  fields={[
    [ 'Date', ({ entry }) => entry.date ],
    [ 'Account', ({ entry }) => entry.account ],
  ]}
/>
```

Now it would be a bit weird to have `renderEntryActions` and `renderEntryChild` separated from the rest, so maybe we could also move `fields`, `renderEntryActions` and `renderEntryChild` under a single `entry` (or `entryConfig`) prop, containing everything related to the configuration of individual entries:

```jsx
<DataView
  entry={
    fields: [
      { 'Date', render: Date },
      { label: 'Source/recipient', render: Source },
      { label: 'Reference', render: Reference },
      { label: 'Amount', render: Amount },
    ],
    actions: () => null,
    child: () => null,
  }
/>
```

The concise mode would look like this:

```jsx
<DataView
  entry={
    fields: [
      [ 'Date', ({ entry }) => entry.date ],
      [ 'Account', ({ entry }) => entry.account ],
    ]
  }
/>
```

And we could even allow to skip `fields` if there is no action or child content:

```jsx
<DataView
  entry={[
    [ 'Date', ({ entry }) => entry.date ],
    [ 'Account', ({ entry }) => entry.account ],
  ]}
/>
```

apps/finance/app/src/App.js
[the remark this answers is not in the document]
AquiGorka: This is not only about the DataView, but the whole app, like the first time the user comes here and no txs have occured and there is nothing in the vault.

apps/finance/app/src/components/Transfers.js
AquiGorka: I really like this last option, I think it makes more sense to put these together.

AquiGorka: Maybe identity as default?

apps/finance/app/index.html
```diff
@@ -5,6 +5,15 @@
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <meta name="theme-color" content="#000000">
     <title>Finance</title>
+    <style>
```
sohkai: Not sure this is the way we'd like to solve this. We should probably find a more generic solution to this with @bpierre.

AquiGorka: I'm open to suggestions.

AquiGorka: cc @chadoh maybe he has 2 satoshis to share about this

bpierre: Yes, we should try to avoid requiring boilerplate CSS to make aragonUI components work, if possible.

A solution might be to use [`BaseStyles`](https://github.com/aragon/aragon-ui/blob/newstyle/src/components/BaseStyles/BaseStyles.js) for this, and use `Layout` for the `height: 100%` + `overflow: auto` (the same way `AppView` was doing it).

The idea with Layout is that it can be nested to set another width (to simulate the behavior of iframe apps on internal apps or global settings), so we should ensure it works properly in these cases. We should also make sure this works well with the client itself.

Another solution I can see would be to apply `overflow: hidden` to `html` and `body` when the SidePanel starts its opening transition, but it would create a jump :-/

AquiGorka: > Another solution I can see would be to apply overflow: hidden to html and body when the SidePanel starts its opening transition, but it would create a jump :-/

Yeah, I want to avoid that as it is the `JS` solution and I think this can be solved by setting the right css rules.

AquiGorka: The `html` and `body` overflow hidden is more of a safeguard as `#root` should not take more than `100vh`, what I'm thinking is should/could we provide a `main` container for apps for this or should app developers be responsible for it?

bpierre: I think we should provide it yes :-) If we want to avoid having another component, `Layout` is seems the best place to do that? We could add in the documentation that if the layout is disabled (on `Main`), then it is the author’s responsibility to do it.

sohkai:
@AquiGorka I've pushed a number of commits to do some housecleaning / polishing.

cc @dizzypaty the main UI changes, amongst a number of other bug fixes to the deposit panel's token validation:

**1. Fixing the error state; it now looks like this (I think good enough for now, until we can revisit all error states properly):**

Screen Shot 2019-08-11 at 2 04 16 AM

**2. Adding the `TokenBadge` inline to all token balance statements in the deposit panel:**

Screen Shot 2019-08-11 at 1 58 02 AM

Screen Shot 2019-08-11 at 1 59 19 AM

It used to look like (and with a "fake" `TokenBadge`):

Screen Shot 2019-08-11 at 1 59 00 AM

**3. A number of grid-unit consistency fixes (most things are now multiples of `8px`), with the exception of `Field`s in panels

apps/finance/app/src/components/EmptyFilteredTransfers.js
```diff
@@ -0,0 +1,71 @@
...
+        css={`
+          max-width: 270px;
+          text-align: center;
+          margin-bottom: ${4 * GU}px;
+          color: ${theme.surfaceContentSecondary};
+          font-size: 16px;
+        `}
+      >
+        We can’t find any item matching your search query.{' '}
+        <Button
```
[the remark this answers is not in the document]
sohkai: Done in https://github.com/aragon/aragon-ui/pull/451.

apps/finance/app/src/components/Transfers.js
sohkai: Moved to https://github.com/aragon/aragon-ui/issues/468.

apps/finance/app/src/components/EmptyTransactions.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import { Box, GU, useTheme } from '@aragon/ui'
+import noTransfersSvg from './assets/no-transfers.svg'
```
sohkai: @owisixseven This one should also be optimized; it's at 2.5M :).

apps/finance/app/src/components/Transfers.js
```diff
@@ -175,6 +176,10 @@ const Transfers = React.memo(({ dao, tokens, transactions }) => {
       selectedDateRange.start ||
       selectedDateRange.end)
 
+  if (!transactions.length) {
```
sohkai: What do you think about moving this to `App` instead? That way we can also change the `transactions` prop in `Transfers` to be required.

apps/finance/app/src/components/Balances.js
```diff
@@ -54,6 +54,23 @@ class Balances extends React.Component {
...
+        <Box heading="Token Balances">
+          <div
+            css={`
+              margin: ${5 * GU}px auto;
+              text-align: center;
+              ${textStyle('body1')};
+              color: ${theme.content};
+            `}
+          >
+            No token balances yet.
```
sohkai: cc @dizzypaty This would also be shown if someone had transferred out all of their tokens (e.g. they transferred 1 ETH in, and then transferred 1 ETH out).

Perhaps just cutting off the "yet" bit (so it becomes `No token balances`) is good enough for now to not be confusing in that case, and we can decide how to better show this case later.

apps/finance/app/src/components/NewTransfer/Deposit.js
```diff
@@ -318,18 +318,14 @@ class Deposit extends React.Component {
...
-                Remember, Mainnet organizations use real (not test) funds.{' '}
-                <StyledSafeLink href={MAINNET_RISKS_BLOG_POST} target="_blank">
-                  Learn more
-                </StyledSafeLink>{' '}
-                about the risks and what's been done to mitigate them here.
-              </p>
-              <VSpace size={1} />
-            </React.Fragment>
+            <p>
+              Remember, Mainnet organizations use <strong>real tokens</strong>.
```
sohkai: cc @dizzypaty I would still suggest "funds" as the terminology, since ETH is not _really_ a token.

apps/finance/app/src/components/NewTransfer/Deposit.js
```diff
@@ -318,18 +318,14 @@ class Deposit extends React.Component {
...
+              Configure your deposit above, and sign the transaction with your
```
sohkai: We don't really need this bit here, do we? It's already in the section below.

apps/finance/app/src/components/NewTransfer/TokenSelectorInstance.js
```diff
@@ -17,9 +18,7 @@ class TokenSelectorInstance extends React.PureComponent {
         {symbol && <TokenSymbol>{symbol}</TokenSymbol>}
         {name && <TokenName>({name})</TokenName>}
         {!addressesEqual(address, ETHER_TOKEN_FAKE_ADDRESS) && (
-          <StyledAddressTag title={address}>
-            {shortenAddress(address, shorten ? 5 : 10)}
-          </StyledAddressTag>
+          <LocalIdentityBadge readOnly compact entity={address} />
```
sohkai: Should this be `badgeOnly`?

AquiGorka: Ja! Good catch

apps/finance/app/src/components/TransfersFilters.js
```diff
@@ -16,152 +13,46 @@ const TransfersFilters = ({
...
+    <div
+      css={`
+        margin-bottom: ${1 * GU}px;
+        display: inline-grid;
+        grid-gap: ${1.5 * GU}px;
+        grid-template-columns: auto auto auto;
+      `}
+    >
+      <DropDown
+        label="Type"
```
sohkai: cc @dizzypaty Does it make sense for this to be renamed "Transfer type" or "Transaction type"?

apps/finance/app/src/components/TransfersFilters.js
```diff
@@ -16,152 +12,39 @@ const TransfersFilters = ({
...
+    <div
+      css={`
+        margin-bottom: ${1 * GU}px;
+        display: inline-grid;
+        grid-gap: ${1.5 * GU}px;
+        grid-template-columns: auto auto auto;
+      `}
+    >
+      <DropDown
+        placeholder="Type"
```
sohkai: cc @dizzypaty Does it make sense for this to be renamed "Transfer type" or "Transaction type"?

sohkai:
@dizzypaty I've lowered the vertical padding of the empty state just a tad: the designs have them at around 160px (20 grid units) but I've taken them down to 120px (15 grid units) because on a 1440x900 display (usually a 13' laptop), the text gets cut off:

Screen Shot 2019-08-18 at 1 02 22 AM

PR #932, opened 2019-07-24
Shared: Move numbers helpers to shared sub repo

facuspagnuolo:


apps/voting/test/helpers/voting.js
```diff
@@ -1,10 +1,17 @@
-const getVoteState = async (voting, id) => {
-  const isOpen = await voting.isVoteOpen(id)
-  const [isExecuted, startDate, snapshotBlock, support, quorum, overruleWindow, earlyExecution, yeas, nays, votingPower, script] = await voting.getVote(id)
+module.exports = web3 => {
+  const { bigExp } = require('@aragon/test-helpers/numbers')
```
sohkai:
```suggestion
  const { bigExp } = require('@aragon/test-helpers/numbers')(web3)
```

PR #933, opened 2019-07-25
Fix TokenManager radspec typo

macor161:
Fixes https://github.com/aragon/aragon-cli/issues/524 
Turns out it was simply a typo in the radspec of the `TokenManager` `initialize` function.


0xGabi:
Nice catch Mathew 👏

Issue #24, opened 2017-11-09
Assert that calculatePrice function is running in correct period


izqui:
`calculatePrice(...)` is only called by functions that have already transitioned sale periods if needed. We should add an assertion in the function to ensure that the sale is in the correct period. Given that the check is redundant and under normal execution shouldn't happen, it should be checked with an `assert(...)`

https://github.com/aragon/aragon-apps/blob/d4615757447fec2903935b1885129e2e513742ff/apps/fundraising/contracts/Fundraising.sol#L288

PR #935, opened 2019-07-26
Voting new styles for 0.8

### LINKED ISSUE #714: Voting: show execution target for binding votes
On votes that include an execution script, we should try to parse it to figure out what the execution target address is and display it to the user.

Finding the function name and decoding parameters will be problematic in the generic case as it's unlikely we'll get the ABI, but we could figure it out for installed apps if we realize the target address is to one of them.

AquiGorka:
Fixes #714.

## How to test

Either ask the author if the latest changes have been deployed to `preview.aragon.org`

or

### First time

```sh
cd ~/projects
```

```sh
cd ~/projects/aragon-ui
npm link
```

### Getting updates

```sh
cd ~/projects/aragon-ui
git checkout newstyle
git pull origin newstyle
npm install

cd ~/projects/aragon
git checkout newstyle
git pull origin newstyle
npm install
npm link @aragon/ui

cd ~/projects/aragon-apps/apps/token-manager/app
git checkout newstyle/voting
git pull origin newstyle/voting
npm install
npm link @aragon/ui
```

### Running it

```sh
cd ~/projects/aragon-apps/apps/token-manager/app
npm start
```

```sh
cd ~/projects/aragon
```


apps/voting/app/src/components/Vote.js
```diff
@@ -0,0 +1,274 @@
...
+                  width: auto;
+                  border-radius: 100px;
+                  padding: 0 ${1 * GU}px;
+                  background: ${theme.infoSurface.alpha(0.08)};
+                  color: ${theme.info};
+                  ${textStyle('label2')};
+                `}
+              >
+                <IconCheck size="tiny" /> Voted
+              </div>
```
sohkai: @dizzypaty Should this be a `<Tag size="compact">`?

Currently using that displays:

Screen Shot 2019-08-09 at 3 10 53 PM

Which is quite different than the [design](https://www.figma.com/file/O09HJRCFDwDTni4VE3tiRvwI/Voting?node-id=375%3A866).

apps/voting/app/src/components/DateRange/DateRangeInput.js
```diff
@@ -0,0 +1,348 @@
+import React from 'react'
```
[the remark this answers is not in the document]
AquiGorka: I am mostly eager to have the time to implement or to import the implemented `DateRangeInput` from `aragonUI`. Having said that, if these differences are noticeable yeah, we should merge

apps/voting/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
```diff
@@ -51,7 +51,7 @@ const Label = styled.span`
   white-space: nowrap;
 `
 
-const StyledBadge = styled(Badge)`
+const StyledBadge = styled(Tag).attrs({ mode: 'app' })`
```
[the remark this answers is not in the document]
AquiGorka: I was keeping a backwards compatible change

apps/voting/app/src/screens/Votes.js
```diff
@@ -1,17 +1,296 @@
...
+
+  return (
+    <React.Fragment>
+      {selectedVote !== null && (
+        <React.Fragment>
+          <Bar>
+            <BackButton onClick={handleBackClick} />
+          </Bar>
+          <Vote vote={selectedVote} onVote={onVote} onExecute={onExecute} />
+        </React.Fragment>
```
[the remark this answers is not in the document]
AquiGorka: I avoided refactoring the code to focus on the new styles so that this does not push back the release but for sure, we could refactor the code, maybe after the release?

apps/voting/app/src/screens/Votes.js
```diff
@@ -1,17 +1,296 @@
...
+          trendFilter === 0 ||
+          ((open && (trendFilter === 1 && yea.cmp(nay) === 1)) ||
+            (trendFilter === 2 && yea.cmp(nay) !== 1))) &&
+        // outcome
+        (outcomeFilter === -1 ||
+          outcomeFilter === 0 ||
+          ((!open &&
+            (outcomeFilter === 1 &&
+              (voteStatus === VOTE_STATUS_ACCEPTED ||
+                voteStatus === VOTE_STATUS_EXECUTED))) ||
```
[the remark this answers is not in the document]
AquiGorka: I think votes that have been accepted include votes pending enactment no?

apps/voting/app/src/screens/Votes.js
```diff
@@ -1,17 +1,296 @@
...
+        // app
+        // date range
+        (!(start || end) ||
+          isWithinInterval(new Date(startTimestamp), {
+            start: startOfDay(start),
+            end: endOfDay(end),
+          }) ||
+          isWithinInterval(endDate, {
+            start: startOfDay(start),
+            end: endOfDay(end),
```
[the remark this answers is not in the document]
AquiGorka: I do believe there is a check elsewhere, I need to verify.

apps/voting/app/src/screens/Votes.js
```diff
@@ -1,17 +1,296 @@
...
+    voteOutcomeFilter,
+    handleVoteOutcomeFilterChange,
+    voteTrendFilter,
+    handleVoteTrendFilterChange,
+    voteAppFilter,
+    handleVoteAppFilterChange,
+    voteDateRangeFilter,
+    handleVoteDateRangeFilterChange,
+    handleClearFilters,
+  } = useFilterVotes(votes, votes.map(({ voteId }) => voteId))
```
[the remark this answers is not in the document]
AquiGorka: Yeah, I was testing something and this must've slipped in through the commits

apps/voting/app/src/screens/Votes.js
```diff
@@ -1,17 +1,296 @@
...
+          <Bar>
+            <BackButton onClick={handleBackClick} />
+          </Bar>
+          <Vote vote={selectedVote} onVote={onVote} onExecute={onExecute} />
+        </React.Fragment>
+      )}
+      {!selectedVote && (
+        <div
+          css={`
+            & > div > div {
```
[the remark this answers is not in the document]
AquiGorka: Yep, it's a quick fix while the component got the fix itself

apps/voting/app/src/components/VotingCard/VotingCardGroup.js
```diff
@@ -1,52 +1,55 @@
...
+        </div>
+        <span
+          css={`
+            margin-left: 10px;
+            display: flex;
+            align-items: center;
+            justify-content: center;
+          `}
+        >
+          <Tag>{count}</Tag>
```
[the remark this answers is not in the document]
AquiGorka: `Tag` should apply the necessary styles cc @dizzypaty

apps/voting/app/src/components/Vote.js
```diff
@@ -0,0 +1,274 @@
...
+import VoteActions from './VoteActions'
+import { getQuorumProgress } from '../vote-utils'
+import { VOTE_NAY, VOTE_YEA } from '../vote-types'
+import { percentageList, round, safeDiv } from '../math-utils'
+import { formatDate } from '../utils'
+
+const DEFAULT_DESCRIPTION =
+  'No additional description has been provided for this proposal.'
+const SEPARATOR = "for '"
+const EXECUTING = ' , executing'
```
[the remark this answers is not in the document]
AquiGorka: Ok, then we need to bring in @dizzypaty into this discussion, we don't have a clear way to define what is the title and what is the rest of the text - or, help me figure this one out.

apps/voting/app/src/components/Vote.js
```diff
@@ -0,0 +1,274 @@
...
+              margin-top: ${2.5 * GU}px;
+            `}
+          >
+            <div
+              css={`
+                ${textStyle('title2')};
+              `}
+            >
+              <span css="font-weight: bold;">Vote #{voteId}:</span>{' '}
+              <VoteText text={title} />
```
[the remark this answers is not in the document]
AquiGorka: Ok, that link does not really resolve to a specific component, can you tag me in the actual file and I'll review there?

apps/voting/app/src/components/VoteActions.js
```diff
@@ -0,0 +1,257 @@
...
+    return (
+      <React.Fragment>
+        {canExecute && !executed && isVoteAction(vote) && (
+          <React.Fragment>
+            <Button mode="strong" onClick={onExecute} wide>
+              Enact this vote
+            </Button>
+            <Info>
+              The voting perdiod is closed and the vote status is passed.{' '}
+              <span css="font-weight: bold;">Anyone</span> can now enact this
```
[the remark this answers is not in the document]
AquiGorka: `em` makes things look like _italics_ no?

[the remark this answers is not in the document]
AquiGorka: I don't mind either, just want to make sure we're all aligned cc @dizzypaty

apps/voting/app/src/components/VoteActions.js
```diff
@@ -0,0 +1,257 @@
...
+            <div
+              css={`
+                ${textStyle('body1')};
+              `}
+            >
+              You must enable your account to vote this proposal
+            </div>
+            <div
+              css={`
+                ${textStyle('body2')};
```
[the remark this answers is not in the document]
AquiGorka: Oh, `textStyle` does not cover color, it has to be set independently.

apps/voting/app/src/screens/Votes.js
```diff
@@ -1,17 +1,296 @@
...
+          ((open && statusFilter === 1) || (!open && statusFilter === 2))) &&
+        // trend
+        (trendFilter === -1 ||
+          trendFilter === 0 ||
+          ((open && (trendFilter === 1 && yea.cmp(nay) === 1)) ||
+            (trendFilter === 2 && yea.cmp(nay) !== 1))) &&
+        // outcome
+        (outcomeFilter === -1 ||
+          outcomeFilter === 0 ||
+          ((!open &&
```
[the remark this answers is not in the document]
AquiGorka: True, this filter is only going to be eligible if the vote status filter is set to open.

apps/voting/app/src/components/Vote.js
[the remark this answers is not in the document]
sohkai: Hopefully this link will work: https://www.figma.com/file/O09HJRCFDwDTni4VE3tiRvwI/Voting?node-id=1604%3A1442

Otherwise I've also left a comment there :).

apps/voting/app/src/screens/Votes.js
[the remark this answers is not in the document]
sohkai: I see that the DatePicker won't let us only select one of the dates, but this seems quite brittle (if we ever change it afterwards to allow just one).

What do you think about doing something like:

```js
!start || isAfter(startTimestamp, startOfDay(start)) &&
!end || isEnd(startTimestamp, endOfDay(end))
```

apps/voting/app/src/components/VotingCard/VotingCardGroup.js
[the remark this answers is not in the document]
sohkai: Here I imagine we should also switch to the `Count` component?

apps/voting/app/src/components/Vote.js
```diff
@@ -102,14 +102,14 @@ function Vote({ vote, onVote, onExecute }) {
               margin-top: ${2.5 * GU}px;
             `}
           >
-            <div
+            <h1
```
sohkai: Could we also use h3s or etc. for things like the `DESCRIPTION`, `CREATED BY`, etc. headings?

apps/voting/app/src/components/DateRange/DateRangeInput.js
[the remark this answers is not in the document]
sohkai: Added into Finance in https://github.com/aragon/aragon-apps/pull/929/commits/e3aaad06b9fa9eae2cb545db2b3dc966f481db17.

The changes do fix some of the border shadow for the new design system so it makes sense to update it there.

sohkai:
cc @owisixseven We should really find a way to optimize `no-votes.svg`; the file is currently 3.9MB 😱

apps/voting/app/src/components/VoteActions.js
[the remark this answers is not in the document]
sohkai: Hmm yes, perhaps `strong` is better since `em` usually defaults to italics. [They're roughly the same semantically.](http://html5doctor.com/i-b-em-strong-element/)

apps/voting/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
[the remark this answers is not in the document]
sohkai: Since we're going to have to move towards an `AppBadge` for these cases anyway, I've mocked it in d1f93fb.

apps/voting/app/src/components/NewVotePanel.js
```diff
@@ -49,61 +58,50 @@ class NewVotePanelContent extends React.PureComponent {
...
-              </Info.Action>
-            </div>
-          </Form>
-        )}
-        {screenIndex === 1 && (
-          <div>
-            <Info.Action title="These votes are binding">
-              <div
+        <Form onSubmit={this.handleSubmit}>
+          <label>
```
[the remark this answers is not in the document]
sohkai: cc @dizzypaty Let's do this independently outside of Voting, since we'll have to apply this change holistically across all the apps and the Client.

apps/voting/app/src/components/Vote.js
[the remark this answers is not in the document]
sohkai: 55eead6 applies the new designs:

- [Card](https://www.figma.com/file/O09HJRCFDwDTni4VE3tiRvwI/Voting?node-id=42%3A218)
- [Detail](https://www.figma.com/file/O09HJRCFDwDTni4VE3tiRvwI/Voting?node-id=42%3A218)

apps/voting/app/src/components/Vote.js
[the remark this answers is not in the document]
sohkai: 55eead6 applies the new designs:

- [Card](https://www.figma.com/file/O09HJRCFDwDTni4VE3tiRvwI/Voting?node-id=42%3A218)
- [Detail](https://www.figma.com/file/O09HJRCFDwDTni4VE3tiRvwI/Voting?node-id=42%3A218)

apps/voting/app/src/components/VoteActions.js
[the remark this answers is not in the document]
sohkai: Fixed in 89e3d6f

apps/voting/app/src/components/Vote.js
sohkai: Done in 56cda30.

apps/voting/app/src/components/VoteActions.js
```diff
@@ -0,0 +1,257 @@
...
+              Enact this vote
+            </Button>
+            <Info>
+              The voting perdiod is closed and the vote status is passed.{' '}
+              <span css="font-weight: bold;">Anyone</span> can now enact this
+              vote to execute its action.
+            </Info>
+          </React.Fragment>
+        )}
+        {connectedAccountVote !== VOTE_ABSENT && (
```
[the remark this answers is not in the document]
sohkai: Done in f625c92.

apps/voting/app/src/components/VoteStatus.js
```diff
@@ -1,67 +1,78 @@
...
   VOTE_STATUS_ACCEPTED,
   VOTE_STATUS_EXECUTED,
+  VOTE_STATUS_PENDING_ENACTMENT,
 } from '../vote-types'
 import { isVoteAction, getVoteStatus } from '../vote-utils'
 
+const POSITIVE = Symbol('positive')
+const NEGATIVE = Symbol('negative')
+
 const ATTRIBUTES = {
```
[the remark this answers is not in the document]
sohkai: Done in 39fd435.

apps/voting/app/src/components/VoteStatus.js
```diff
@@ -1,67 +1,78 @@
...
-      ? 'Accepted'
-      : ATTRIBUTES[status].label
+  const theme = useTheme()
+  const { pctBase } = useSettings()
+  const status = getVoteStatus(vote, pctBase)
+  const { Icon, color, label } = ATTRIBUTES[
+    status === VOTE_STATUS_ACCEPTED &&
+    (isVoteAction(vote) && !vote.data.executed)
+      ? VOTE_STATUS_PENDING_ENACTMENT
+      : status
```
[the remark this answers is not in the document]
sohkai: Done in 39fd435.

apps/voting/app/src/screens/Votes.js
```diff
@@ -1,17 +1,296 @@
...
+            (outcomeFilter === 1 &&
+              (voteStatus === VOTE_STATUS_ACCEPTED ||
+                voteStatus === VOTE_STATUS_EXECUTED))) ||
+            (outcomeFilter === 2 && voteStatus === VOTE_STATUS_REJECTED) ||
+            (outcomeFilter === 3 && voteStatus === VOTE_STATUS_EXECUTED) ||
+            (outcomeFilter === 4 &&
+              voteStatus === VOTE_STATUS_ACCEPTED &&
+              isVoteAction(vote) &&
+              !executed))) &&
```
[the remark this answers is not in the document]
sohkai: Done in 39fd435.

apps/voting/app/src/screens/Votes.js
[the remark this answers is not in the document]
sohkai: In 39fd435, I've separated this out such that the "lifecycle" of a vote's status is now

```
open
   |
   |----> rejected
   |
   |----> accepted
             |
             |--- (if action) --> pending enactment ---> enacted
```

apps/voting/app/src/vote-types.js
```diff
@@ -8,3 +8,6 @@ export const VOTE_STATUS_ONGOING = Symbol('VOTE_STATUS_ONGOING')
 export const VOTE_STATUS_REJECTED = Symbol('VOTE_STATUS_REJECTED')
 export const VOTE_STATUS_ACCEPTED = Symbol('VOTE_STATUS_ACCEPTED')
 export const VOTE_STATUS_EXECUTED = Symbol('VOTE_STATUS_EXECUTED')
+export const VOTE_STATUS_PENDING_ENACTMENT = Symbol(
```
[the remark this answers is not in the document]
sohkai: Done in 39fd435.

apps/voting/app/src/screens/Votes.js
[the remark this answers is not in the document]
sohkai: Done in 8928076 and fe92175.

apps/voting/app/src/components/Vote.js
```diff
@@ -0,0 +1,274 @@
+import React, { useCallback } from 'react'
```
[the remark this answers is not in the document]
sohkai: Done in 8928076 and fe92175.

apps/voting/app/src/screens/Votes.js
```diff
@@ -1,17 +1,296 @@
...
+        // app
+        // date range
+        (!(start || end) ||
+          isWithinInterval(new Date(startTimestamp), {
```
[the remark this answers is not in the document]
sohkai: Done in 21a130b.

apps/voting/app/src/components/VoteActions.js
```diff
@@ -0,0 +1,271 @@
...
+              You must enable your account to vote on this proposal
+            </div>
+            <div
+              css={`
+                ${textStyle('body2')};
+                color: ${theme.surfaceContentSecondary};
+                margin-top: ${2 * GU}px;
+              `}
+            >
+              Connect to your Ethereum provider by clicking on the Enable
```
sohkai: Should we add the lightning icon here?

cc @dizzypaty Perhaps we could also add emphasis on the "Enable account"?

apps/voting/app/src/components/VoteActions.js
```diff
@@ -0,0 +1,271 @@
...
+        )}
+      </div>
+    )
+  }
+
+  return (
+    <div>
+      <Buttons disabled />
+      <Info mode="warning">
+        The currently connected account did not hold any {tokenSymbol} at the
```
sohkai: cc @dizzypaty Given that we add emphasis everywhere else for the token symbol, would it also make sense in this text?

apps/voting/app/src/components/VoteActions.js
```diff
@@ -0,0 +1,271 @@
...
+        time this vote began ({formatDate(startDate)}), and therefore cannot
+        participate in this vote. Make sure your accounts are holding{' '}
+        {tokenSymbol} at the time a vote begins if you'd like to vote using this
+        Voting app.{' '}
+        <ExternalLink
+          target="_blank"
+          href="https://wiki.aragon.org/documentation/aragon_network_token/"
```
sohkai: @dizzypaty This was a gross oversight on my part. We should not use this link, since this is specific to only `governance.aragonproject.eth`'s Voting apps.

Some ideas instead:

- Some text saying "Please ask the organization's administrators how to obtain tokens"
- A link to the token contract on etherscan

apps/voting/app/src/components/VoteActions.js
```diff
@@ -0,0 +1,271 @@
...
+            </Button>
+            <Info>
+              The voting period is closed and the vote has passed.{' '}
+              <span css="font-weight: bold;">Anyone</span> can now enact this
+              vote to execute its action.
+            </Info>
+          </React.Fragment>
+        )}
+        {hasVoted && (
+          <div
```
sohkai: We probably want to center this with the content cc @dizzypaty?

This is how it currently looks:

Screen Shot 2019-08-09 at 2 35 34 PM

apps/voting/app/src/utils.js
```diff
@@ -8,3 +10,7 @@ export function pluralize(count, singular, plural, re = PLURALIZE_RE) {
 }
 
 export function noop() {}
+
+export function formatDate(date) {
+  return `${format(date, 'dd/MM/yy')} at ${format(date, 'HH:mm')} UTC`
```
sohkai: @dizzypaty This produces strings like:

```
07/12/18 at 16:22 UTC
```

But perhaps a more natural sounding alternative:

```
16:22 UTC on 07/12/18
```

Or alternatively one that spells out the month and day, since `07/12` is ambiguous across regions (July 12, or Dec. 7)?

apps/voting/app/src/screens/VoteDetail.js
```diff
@@ -0,0 +1,276 @@
...
+                    width: auto;
+                    border-radius: 100px;
+                    padding: 0 ${1 * GU}px;
+                    background: ${theme.infoSurface.alpha(0.08)};
+                    color: ${theme.info};
+                    ${textStyle('label2')};
+                  `}
+                >
+                  <IconCheck size="tiny" /> Voted
+                </div>
```
sohkai: @dizzypaty Should this be a `<Tag size="small">`?

Currently using that displays:

Screen Shot 2019-08-12 at 1 11 36 AM

Which is quite different than the [design](https://www.figma.com/file/O09HJRCFDwDTni4VE3tiRvwI/Voting?node-id=375%3A866).

apps/voting/app/src/components/VoteActions.js
```diff
@@ -0,0 +1,271 @@
...
+                .
+              </div>
+            </div>
+          </div>
+        )}
+      </React.Fragment>
+    )
+  }
+
+  if (canUserVote && hasVoted && !changeVote) {
```
sohkai: Did we decide to not show the "vote success" bit when you can change your vote ([design](https://www.figma.com/file/O09HJRCFDwDTni4VE3tiRvwI/Voting?node-id=1266%3A0))?

This in particular seems to have deviated from the design a bit. It currently looks like:

Screen Shot 2019-08-12 at 1 16 40 AM

apps/voting/app/src/components/Vote.js
```diff
@@ -0,0 +1,274 @@
...
+import { VOTE_NAY, VOTE_YEA } from '../vote-types'
+import { percentageList, round, safeDiv } from '../math-utils'
+import { formatDate } from '../utils'
+
+const DEFAULT_DESCRIPTION =
+  'No additional description has been provided for this proposal.'
+const SEPARATOR = "for '"
+const EXECUTING = ' , executing'
+
+function Vote({ vote, onVote, onExecute }) {
```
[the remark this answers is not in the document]
sohkai: @AquiGorka This is definitely something we should do for this iteration 🔒.

Perhaps this is something we'll build into `CardLayout` eventually, but for now we can apply a custom solution here cc @bpierre.

apps/voting/app/src/components/VoteActions.js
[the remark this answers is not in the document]
dizzypaty: Please use strong to empathize.

apps/voting/app/src/components/VoteActions.js
dizzypaty: Yes, the lighting icon should be there and I agree with making 'Enable account' strong

apps/voting/app/src/components/VoteActions.js
dizzypaty: Agree!

apps/voting/app/src/components/VoteActions.js
dizzypaty: Your suggestions are great, let's make that change.

apps/voting/app/src/components/VoteActions.js
dizzypaty: Yes, the content should be centered. I also changed the copy to make the action clearer, making this module a bit more compact and reduced the icon size. 

Should look like this: 

VOTING - PASSED

apps/voting/app/src/utils.js
dizzypaty: Yes, I think it's clearer if we use this format (if possible): 

`16:22 UTC on 12th of July, 2019`

apps/voting/app/src/screens/VoteDetail.js
dizzypaty: It should use the large tag size.

apps/voting/app/src/components/VoteActions.js
dizzypaty: No, the feedback module (vote success) should be display above the button and info box. I added all the possible vote status to the Figma file. This screen should look like this: 

VOTING - PASSED

apps/voting/app/src/screens/VoteDetail.js
dizzypaty: Also, I know this is not correct in many of the designs, but the vote title 'Vote #24' should use the text style `Title 2`.

apps/voting/app/src/components/VotingCard/VotingCardGroup.js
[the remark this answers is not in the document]
dizzypaty: Yes, the original designs were using a custom text size of 10px, but I just standardized them to the available text styles in aragonUI: `Label 2`. We might need to add one more style in the future but need to discuss with Adri.

apps/voting/app/src/screens/Votes.js
```diff
@@ -1,17 +1,296 @@
...
-    const dateDiff = b.data.endDate - a.data.endDate
-    // Order by descending voteId if there's no end date difference
-    return dateDiff !== 0 ? dateDiff : b.voteId - a.voteId
+const sortVotes = (a, b) => {
+  const dateDiff = b.data.endDate - a.data.endDate
+  // Order by descending voteId if there's no end date difference
+  return dateDiff !== 0 ? dateDiff : b.voteId - a.voteId
+}
+
+const useFilterVotes = votes => {
```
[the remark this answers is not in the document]
AquiGorka: Can you help me understand the benefit of merging `useFilteredVotes` with `useVotes`? I am inclined on having them independent as I see filtering an effect happening outside of votes, easier to decouple, not tied together (hope the image I'm picturing was correctly communicated).

apps/voting/app/src/screens/Votes.js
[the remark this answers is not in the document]
AquiGorka: > if we ever change it afterwards to allow just one

Well, that sort of change is bound to break things no? I'll update these checks

apps/voting/app/src/screens/Votes.js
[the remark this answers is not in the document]
AquiGorka: Quick update here: I had to move `useFilterVotes` one component up (`App`) to make the filters persist when the user navigates to a vote and back.

apps/voting/app/src/useScrollTop.js
```diff
@@ -0,0 +1,9 @@
+import { useEffect } from 'react'
+
+function useScrollTop(dependency) {
+  useEffect(() => {
+    window.scrollTo(0, 0)
```
bpierre: Just as a note: this won’t work if we decide to move to a local scroll (for `SidePanel`), or if we want to reuse it in other cases (e.g. internal apps in aragon/aragon).

Another option is to use `scrollIntoView()` on an element, and let the browser do the rest: https://github.com/aragon/aragon/blob/master/src/apps/Permissions/Permissions.js#L224-L228

owisixseven:
> cc @owisixseven We should really find a way to optimize `no-votes.svg`; the file is currently 3.9MB 😱

Yes, i think i know how optimize all this illustrations

apps/voting/app/src/screens/Votes.js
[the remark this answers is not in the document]
sohkai: > The benefit of merging useFilteredVotes with useVotes

I had the image of just doing the filtering inside `useVotes()`, so that it was just one hook to produce the rendered array of votes. In this particular case I see it as:

```
input votes --> transform for rendering (hook: filter, split) --> render
```

And it's easier if that transform was done in a single hook to ease debugging and reasoning about the code later.

But no particular strong feelings either way, I'd just lean towards less hook definitions than more, especially if they end up manipulating the same state.

apps/voting/app/src/components/VoteActions.js
AquiGorka: @sohkai I thought the date was not easy to fetch (pinged you in Figma about this)

apps/voting/app/src/useScrollTop.js
AquiGorka: > this won’t work if we decide to move to a local scroll (for SidePanel)

Is that something that has been decided? It currently works as is. I'm fond of the `yagni` idea, if/when the change occurs then we should update this.

bpierre: Sounds good :+1:  I posted this so we remember that an update might be needed.

apps/voting/app/src/components/VoteActions.js
sohkai: Yes, we should remove the date from this.

apps/voting/app/src/useScrollTop.js
AquiGorka: Perf, thanks 😄

apps/voting/app/src/components/VoteActions.js
AquiGorka: This was not worked on, I'll remove this an we can push to a next iteration the correct copy + link

apps/voting/app/src/utils.js
AquiGorka: How should we treat this:
image

AquiGorka: Going for this format:
image

apps/voting/app/src/components/VoteActions.js
```diff
@@ -254,10 +254,9 @@ const TokenReference = ({
   startDate,
 }) => (
   <Info>
-    Voting with <span css="font-weight: bold;">{userBalance}</span> of your{' '}
-    <span css="font-weight: bold;">{tokenSymbol}</span> at block{' '}
-    <span css="font-weight: bold;">{snapshotBlock}</span> due to the vote
-    starting at <span css="font-weight: bold;">{formatDate(startDate)}</span>.
+    Voting with <strong>{userBalance}</strong> of your{' '}
+    <strong>{tokenSymbol}</strong> at block <strong>{snapshotBlock}</strong> due
+    to the vote starting at <strong>{formatDate(startDate)}</strong>.
```
sohkai: I'm really pulling at hairs here with what I consider semantic html, but to clarify on the `<strong>` vs. css styling; in cases where the bold is **not** really meant for emphasis, it's fine to use css. I don't see the user balance, token symbol, and snapshot block as being emphasized; you wouldn't stress them when reading them out loud (unlike, for example, a stress on "anyone").

Anyhow, there's no need to change anything but perhaps we'll all learn something from a discussion (again, this is what I've considered semantic and I may well be wrong) :).

apps/voting/app/src/components/DateRange/DateRangeInput.js
```diff
@@ -336,13 +336,8 @@ const Wrap = styled.div`
...
-  const { accent, border } = useTheme()
+  const theme = useTheme()
   return (
-    <DateRangeInput
-      {...props}
-      compactMode={below('medium')}
-      accent={accent}
-      border={border}
-    />
+    <DateRangeInput {...props} compactMode={below('medium')} theme={theme} />
```
sohkai: I believe this component has become different from Finance's again (from the diff in https://github.com/aragon/aragon-apps/pull/929/commits/b85af29fa9848d6c7396f3d5d3985f58dd408caa). Could we make them consistent?

apps/voting/app/src/components/SummaryRows.js
```diff
@@ -70,6 +74,17 @@ function SummaryRow({ color, label, pct, token }) {
   )
 }
 
+const You = () => (
```
sohkai: Perhaps we should extract this out into its own component since it's used in a few places?

apps/voting/app/src/components/VoteActions.js
```diff
@@ -132,20 +132,11 @@ const VoteActions = React.memo(({ vote, onVoteYes, onVoteNo, onExecute }) => {
...
-          css={`
-            color: ${theme.infoSurfaceContent};
-          `}
-        >
-          Find out how to get tokens.
-        </Link>
+        The currently connected account did not hold any{' '}
+        <strong>{tokenSymbol}</strong> at the time this vote began (
+        {formatDate(startDate)}), and therefore cannot participate in this vote.
+        Make sure your accounts are holding {tokenSymbol} at the time a vote
```
sohkai: We should make this instance of the `tokenSymbol` bold as well

apps/voting/app/src/App.js
```diff
@@ -68,13 +83,29 @@ function App() {
...
+              handleVoteStatusFilterChange={handleVoteStatusFilterChange}
+              voteOutcomeFilter={voteOutcomeFilter}
+              handleVoteOutcomeFilterChange={handleVoteOutcomeFilterChange}
+              voteTrendFilter={voteTrendFilter}
+              handleVoteTrendFilterChange={handleVoteTrendFilterChange}
+              voteAppFilter={voteAppFilter}
+              handleVoteAppFilterChange={handleVoteAppFilterChange}
+              voteDateRangeFilter={voteDateRangeFilter}
+              handleVoteDateRangeFilterChange={handleVoteDateRangeFilterChange}
+              handleClearFilters={handleClearFilters}
```
sohkai: Not a big fan of this; why not just use the hook directly in `Votes`?

apps/voting/app/src/App.js
```diff
@@ -68,13 +83,29 @@ function App() {
           />
           {selectedVote ? (
             <VoteDetail
+              selectVote={selectVote}
```
sohkai: AFAIK this isn't used by `VoteDetail`?

sohkai:
We'll get @owisixseven to re-export the icons with their correct path widths so we don't have to use the `stroke-width` override :).

apps/voting/app/src/components/VoteActions.js
AquiGorka: Yeah, I don't disagree but then is it really ok to add visual effects that have nothing to do with the interpretation of the text but use the same style of other effects that do give the text semantic meaning? Anyhow, let's figure this one out and update as needed.

apps/voting/app/src/App.js
AquiGorka: Because filter state gets lost when selecting a vote and coming back to the list, this way it persists and the user can come back to a filtered list of votes.

apps/voting/app/src/components/VoteActions.js
sohkai: Yeah this is true, especially when they're inline with other text. Perhaps if they were on different headings or etc. it would be more appropriate to use just a css style.

apps/voting/app/src/components/DateRange/DateRangeInput.js
AquiGorka: Changes have been submitted all around:
- [x] New PR for experimental DateRange with consolidated changes
- [x] Updates to this PR
- [x] Updates to the Finance PR

apps/voting/app/src/components/SummaryRows.js
AquiGorka: `aragonUI`? or local?

bpierre: Already there :-)

```jsx
<Tag size="small">you</Tag>
```

AquiGorka: I'll add one locally, later we can decide if we want one for `aragonUI`

AquiGorka: Update: added one locally before I saw your comment @bpierre - I'll update

apps/voting/app/src/utils-hooks.js
```diff
@@ -71,3 +71,9 @@ export function usePanelState({ onDidOpen = noop, onDidClose = noop } = {}) {
     [requestOpen, requestClose, visible, didOpen, onTransitionEnd]
   )
 }
+
+export function useScrollTop(dependency) {
```
AquiGorka: I find this type of files: `utils-x` become a bag of cats down the road, by naming files for hooks (like we do components) we can group together behaviour and dependencies that go together. Maybe use a `hook-utils` or `hooks` folder with the hooks in it?

apps/voting/app/src/components/VoteActions.js
```diff
@@ -39,16 +38,24 @@ const VoteActions = React.memo(({ vote, onVoteYes, onVoteNo, onExecute }) => {
   const hasVoted = [VOTE_YEA, VOTE_NAY].includes(connectedAccountVote)
 
   useEffect(() => {
+    let cancelled = false
```
AquiGorka: Hmm, for this you could've used the `usePromise` hook that the inner functions rely on.

apps/voting/app/src/utils-hooks.js
sohkai: A `hooks` folder sounds good; we may also eventually want to have a `utils` folder to organize the root a bit more.

apps/voting/app/src/components/VoteActions.js
sohkai: Hmm yes, but `usePromise()`'s API is based on returning the resulting value from the promise (so you can use it) whereas we just want to cancel it here and avoid a side effect.

AquiGorka: Pseudo code here (this is what I was thinking):

``` JavaScript
const promise = Promise.all([canUserVotePromise, canExecutePromise, userBalancePromise])
const value = usePromise(promise, false)
setReady(value)
```

But yeah, there are other things that might need to be worked on.

apps/voting/app/src/utils-hooks.js
sohkai: Done in 9e1222e.

apps/voting/app/src/hooks/useExtendedVoteData.js
```diff
@@ -15,31 +15,19 @@ export default function useExtendedVoteData(vote) {
   const tokenContract = useTokenContract()
 
   const canExecutePromise = useMemo(() => getCanExecute(vote, api), [api, vote])
-  const canExecute = usePromise(
-    canExecutePromise,
-    [vote && vote.voteId, api],
```
sohkai: @AquiGorka I'm 99% sure we don't need these extra dependencies in the `usePromise()`; if they change, the `useMemo()` should also refire and create a new promise (which will trigger the `usePromise()`).

AquiGorka: Yep, you are right, I did not update any of that code, just added a way to get a hold of the promise itself, this should be updated.

sohkai: Updated in 7842779 😄

AquiGorka: Although I know am curious if the `canExecutePromise` should be a dep here

sohkai: It's an implicit dependency in `usePromise()`

apps/voting/app/src/components/VoteActions.js
sohkai: Yeah, I'm worried there might be funny things with how the state gets updated / triggered (compared to now, which is relatively obvious what happens on cancellation / updates), and to be honest `usePromise()` is something we should find a better abstraction for since it's kind of clunky.

PR #939, opened 2019-07-26
Voting: deploy preview to now

sohkai:
Similar to #937, allows someone to deploy to `nightly-voting.aragon.org` with

```
now --target production
```

### DIFF
--- a/apps/voting/app/now.json
+++ b/apps/voting/app/now.json
@@ -1,0 +1,10 @@
+{
+  "version": 2,
+  "public": true,
+  "scope": "aragon",
+  "name": "voting",
+  "alias": "nightly-voting.aragon.org",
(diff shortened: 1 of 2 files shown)


sohkai:
@AquiGorka Unfortunately as of now, this is not automatic (want to do some other things that will eventually allow us to make automatic staging deploys possible). But, once you install now's desktop app or `@now/cli`, you can (and should!) use the command above to deploy (assuming you've hooked up your now account properly).

The `@aragon/ui` version does not need to be changed if new alphas are published (they should get fetched automatically), but it is not a bad idea to keep moving it up anyway :). I've also updated the comment, it does **not** currently require you to link `@aragon/ui`, that was my mistake earlier :).

Issue #656, opened 2019-02-08
Agent: add ability to guard against token transfers


sohkai:
Given the security implications of the Agent app's `execute()` functionality, it would be nice if users had a way to ensure that their ETH and token holdings are consistent before and after the call (such that no transfers happened, or if they did, at least their app's balance wasn't changed).

This doesn't scale generally, but we could allow users to add token addresses onto a list that we check balances of before and after each invocation of `exeucte()` if the user specified it (either as an argument or another function, e.g. `dangerousExecute()`). For most users, I can imagine this list to be quite small (DAI, ANT, etc.), at least for now. ETH is already taken care of by the existing contract as it's only transferred based on an argument.

Note that the forwarding functionality wouldn't be able to be protected by this, as it doesn't yet accept an input. This is protected under another role that is meant to be higher-privileged than `execute()`'s, and so I would always treat it as `dangerousExecute()`.

cc @lkngtn

izqui:
I thought about guarding against `approve` and `transfer` signatures but it really doesn't seem like a good idea to me.

lkngtn:
The token list check approach seems reasonable, I suspect that even a limited list of tokens would be sufficient for most cases. 

This would not be sufficient to prevent all token transfers or to prevent asset transfers for things like NFTs, but in cases where we know there are specific assets we care about not being transferred then this should provide relatively strong protections. I think this might be a good solution to voting with bonding curve collateral (cc @osarrouy) 

It also reminds me a bit of https://marble.org, which I thought was a pretty interesting mechanism--though I don't think the "flash lending" concept has really taken off.

osarrouy:
That’s sound like a great idea for the collateral pool, indeed!

I guess it would make sense to check the before /after equality of `allowance` too, right ?

sohkai:
Ah yes, checking `allowance` would be good too :+1:.

PR #941, opened 2019-07-31
Merge Pool into Agent

osarrouy:
Fixes https://github.com/aragon/aragon-apps/issues/656.

### DIFF
--- a/apps/agent/contracts/Agent.sol
+++ b/apps/agent/contracts/Agent.sol
@@ -62,7 +62,7 @@ contract Agent is IERC165, ERC1271Bytes, IForwarder, IsContract, Vault {
     * @param _data Calldata for the action
     * @return Exits call frame forwarding the return data of the executed call (either error or success data)
     */
-    function safeExecute(address _target, bytes _data) external auth(SAFE_EXECUTE_ROLE) {
+    function safeExecute(address _target, bytes _data) external authP(SAFE_EXECUTE_ROLE, arr(_target, uint256(getSig(_data)))) {
         uint256 protectedTokensLength = protectedTokens.length;
         address[] memory _protectedTokens = new address[](protectedTokensLength);
         uint256[] memory balances = new uint256[](protectedTokensLength);
--- a/apps/agent/contracts/Agent.sol
+++ b/apps/agent/contracts/Agent.sol
@@ -14,21 +14,99 @@ import "@aragon/os/contracts/common/IForwarder.sol";
 
 
 contract Agent is IERC165, ERC1271Bytes, IForwarder, IsContract, Vault {
+    /* Hardcoded constants to save gas
+    bytes32 public constant SAFE_EXECUTE_ROLE = keccak256("SAFE_EXECUTE_ROLE");
     bytes32 public constant EXECUTE_ROLE = keccak256("EXECUTE_ROLE");
     bytes32 public constant RUN_SCRIPT_ROLE = keccak256("RUN_SCRIPT_ROLE");
+    bytes32 public constant ADD_PROTECTED_TOKEN_ROLE = keccak256("ADD_PROTECTED_TOKEN_ROLE");
+    bytes32 public constant REMOVE_PROTECTED_TOKEN_ROLE = keccak256("REMOVE_PROTECTED_TOKEN_ROLE");
     bytes32 public constant ADD_PRESIGNED_HASH_ROLE = keccak256("ADD_PRESIGNED_HASH_ROLE");
     bytes32 public constant DESIGNATE_SIGNER_ROLE = keccak256("DESIGNATE_SIGNER_ROLE");
+    */
+
+    bytes32 public constant SAFE_EXECUTE_ROLE = 0x0a1ad7b87f5846153c6d5a1f761d71c7d0cfd122384f56066cd33239b7933694;
+    bytes32 public constant EXECUTE_ROLE = 0xcebf517aa4440d1d125e0355aae64401211d0848a23c02cc5d29a14822580ba4;
+    bytes32 public constant RUN_SCRIPT_ROLE = 0xb421f7ad7646747f3051c50c0b8e2377839296cd4973e27f63821d73e390338f;
+    bytes32 public constant ADD_PROTECTED_TOKEN_ROLE = 0x6eb2a499556bfa2872f5aa15812b956cc4a71b4d64eb3553f7073c7e41415aaa;
+    bytes32 public constant REMOVE_PROTECTED_TOKEN_ROLE = 0x71eee93d500f6f065e38b27d242a756466a00a52a1dbcd6b4260f01a8640402a;
+    bytes32 public constant ADD_PRESIGNED_HASH_ROLE = 0x0b29780bb523a130b3b01f231ef49ed2fa2781645591a0b0a44ca98f15a5994c;
+    bytes32 public constant DESIGNATE_SIGNER_ROLE = 0x23ce341656c3f14df6692eebd4757791e33662b7dcf9970c8308303da5472b7c;
+
+    uint256 public constant PROTECTED_TOKENS_CAP = 10;
 
     bytes4 private constant ERC165_INTERFACE_ID = 0x01ffc9a7;
 
+    string private constant ERROR_TOKENS_CAP_REACHED = "AGENT_TOKENS_CAP_REACHED";
+    string private constant ERROR_TOKEN_NOT_ETH_OR_CONTRACT = "AGENT_TOKEN_NOT_ETH_OR_CONTRACT";
+    string private constant ERROR_TOKEN_ALREADY_PROTECTED = "AGENT_TOKEN_ALREADY_PROTECTED";
+    string private constant ERROR_TOKEN_NOT_PROTECTED = "AGENT_TOKEN_NOT_PROTECTED";
+    string private constant ERROR_TARGET_PROTECTED = "AGENT_TARGET_PROTECTED";
+    string private constant ERROR_PROTECTED_TOKENS_MODIFIED = "AGENT_PROTECTED_TOKENS_MODIFIED";
+    string private constant ERROR_BALANCE_NOT_CONSTANT = "AGENT_BALANCE_NOT_CONSTANT";
     string private constant ERROR_DESIGNATED_TO_SELF = "AGENT_DESIGNATED_TO_SELF";
 
     mapping (bytes32 => bool) public isPresigned;
     address public designatedSigner;
+    address[] public protectedTokens;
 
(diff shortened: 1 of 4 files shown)


apps/agent/contracts/Agent.sol
```diff
@@ -59,6 +137,28 @@ contract Agent is IERC165, ERC1271Bytes, IForwarder, IsContract, Vault {
         }
     }
 
+    /**
+    * @notice Add `_token.symbol(): string` to the list of protected tokens
+    * @param _token Address of the token to be protected
+    */
+    function addProtectedToken(address _token) external auth(ADD_PROTECTED_TOKEN_ROLE) {
+        require(protectedTokens.length < PROTECTED_TOKENS_CAP, ERROR_TOKENS_CAP_REACHED);
+        require(isContract(_token) || _token == ETH, ERROR_TOKEN_NOT_ETH_OR_CONTRACT);
```
izqui: What if in addition to ensuring that `_token` is a contract, we perform a `balance(_token)` to make sure that the balance check doesn't revert for the token. If a token is added and `token.balanceOf()` reverts, it will block `safeExecute` until the token is removed from the protected list.

A malicious token could still be added that starts reverting after a state change or some period of time. To prevent this we could add a public function (without authentication) to remove a token that reverts, but this is a bit of an overkill IMO.

apps/agent/contracts/Agent.sol
```diff
@@ -14,22 +14,100 @@ import "@aragon/os/contracts/common/IForwarder.sol";
...
+    bytes32 public constant SAFE_EXECUTE_ROLE = keccak256("SAFE_EXECUTE_ROLE");
     bytes32 public constant EXECUTE_ROLE = keccak256("EXECUTE_ROLE");
     bytes32 public constant RUN_SCRIPT_ROLE = keccak256("RUN_SCRIPT_ROLE");
+    bytes32 public constant ADD_PROTECTED_TOKEN_ROLE = keccak256("ADD_PROTECTED_TOKEN_ROLE");
+    bytes32 public constant REMOVE_PROTECTED_TOKEN_ROLE = keccak256("REMOVE_PROTECTED_TOKEN_ROLE");
     bytes32 public constant ADD_PRESIGNED_HASH_ROLE = keccak256("ADD_PRESIGNED_HASH_ROLE");
     bytes32 public constant DESIGNATE_SIGNER_ROLE = keccak256("DESIGNATE_SIGNER_ROLE");
+    */
+
+    bytes32 public constant SAFE_EXECUTE_ROLE = 0x0a1ad7b87f5846153c6d5a1f761d71c7d0cfd122384f56066cd33239b7933694;
```
izqui: Let's not forget to add these new roles to the `arapp.json` file

osarrouy:
I just updated the roles in the `arapp.json` [cause I forgot before]. I also took the freedom to add the `WITHDRAW_ROLE` inherited from the `Vault` which otherwise won't appear in the client permissions list [or will be unreadable].

apps/agent/contracts/Agent.sol
```diff
@@ -14,22 +14,100 @@ import "@aragon/os/contracts/common/IForwarder.sol";
...
+    function safeExecute(address _target, bytes _data) external auth(SAFE_EXECUTE_ROLE) {
+        address[] memory _protectedTokens = new address[](protectedTokens.length);
+        uint256[] memory balances = new uint256[](protectedTokens.length);
+        bytes32 size;
+        bytes32 ptr;
+
+        for (uint256 i = 0; i < protectedTokens.length; i++) {
+            address token = protectedTokens[i];
+            // we don't care if target is ETH [0x00...0] as it can't be spent anyhow [though you can't invoke anything at 0x00...0]
+            require(_target != token || token == ETH, ERROR_TARGET_PROTECTED);
```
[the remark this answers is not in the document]
osarrouy: Well it was mostly a convenient thing for us regarding our use case: basically the `Controller` could then expose an `addCollateralToken` function which would forward the call to `MarketMaker`, `Tap` and `Pool` [no matter whether this is an ERC20 or ETH].

You're right that it does not make a lot of sense into the `Agent` as such. I will update it and handle this specific use case in the `Controller` contract.

apps/agent/contracts/Agent.sol
osarrouy: Done now [see my comment above].

apps/agent/contracts/Agent.sol
```diff
@@ -14,22 +14,100 @@ import "@aragon/os/contracts/common/IForwarder.sol";
...
+            mstore(0x40, add(ptr, size))
+            returndatacopy(ptr, 0, size)
+        }
+
+        if (result) {
+            // if the underlying call has succeeded, we check that the protected tokens
+            // and their balances have not been modified and return the call's return data
+            for (uint256 j = 0; j < _protectedTokens.length; j++) {
+                require(protectedTokens[j] == _protectedTokens[j], ERROR_PROTECTED_TOKENS_MODIFIED);
+                require(balances[j] == balance(_protectedTokens[j]), ERROR_BALANCE_NOT_CONSTANT);
```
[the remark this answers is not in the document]
osarrouy: I think it makes sense indeed ... Gonna update that.

apps/agent/contracts/Agent.sol
```diff
@@ -14,22 +14,100 @@ import "@aragon/os/contracts/common/IForwarder.sol";
...
     event PresignHash(address indexed sender, bytes32 indexed hash);
     event SetDesignatedSigner(address indexed sender, address indexed oldSigner, address indexed newSigner);
 
+    /**
+    * @notice Safe execute '`@radspec(_target, _data)`' on `_target`
+    * @param _target Address where the action is being executed
+    * @param _data Calldata for the action
+    * @return Exits call frame forwarding the return data of the executed call (either error or success data)
+    */
+    function safeExecute(address _target, bytes _data) external auth(SAFE_EXECUTE_ROLE) {
```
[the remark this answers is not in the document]
osarrouy: Sounds good.

apps/agent/contracts/Agent.sol
```diff
@@ -14,22 +14,100 @@ import "@aragon/os/contracts/common/IForwarder.sol";
...
+        assembly {
+            size := returndatasize
+            ptr := mload(0x40)
+            mstore(0x40, add(ptr, size))
+            returndatacopy(ptr, 0, size)
+        }
+
+        if (result) {
+            // if the underlying call has succeeded, we check that the protected tokens
+            // and their balances have not been modified and return the call's return data
```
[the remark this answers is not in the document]
osarrouy: Yeah. This could also allow us to make sure no protected tokens has been added [though i'm not sure this could be an attack vector].

apps/agent/contracts/Agent.sol
osarrouy: > What if in addition to ensuring that _token is a contract, we perform a balance(_token) to make sure that the balance check doesn't revert for the token. If a token is added and token.balanceOf() reverts, it will block safeExecute until the token is removed from the protected list

Hmmmm. Yeah. I will add that too.

> A malicious token could still be added that starts reverting after a state change or some period of time. To prevent this we could add a public function (without authentication) to remove a token that reverts, but this is a bit of an overkill IMO.

Yeah I think it will mostly make things harder to understand [regarding roles and all] for end users while the treat is really low ...

osarrouy: Thinking about it: what exactly is the threat model here?

1. If we are trying to deal with malicious tokens : what about the situation where calling `balanceOf` would actually empty your balance ? It's unlikely cause if the token is malicious / not trustless it could directly move your tokens [which would be pointless cause they would enp be worth nothing in the end].

2. We are just trying to prevent users to do a mistake while copy / pasting the address of the token ?

osarrouy:
@izqui All comments should be fixed now [excepts for the ERC20 check thing cause i'm waiting for your answer on that].

apps/agent/contracts/Agent.sol
izqui: For me the threat model is adding a contract to the list of protected tokens that can brick `safeExecute` until the token is removed. This could happen by mistake or maliciously and it could be specially bad if the permission to remove a token from the list is assigned to a long/complicated process or even burned.

But as you point out, adding a malicious token could bypass any check that we add here, so we would really just be protecting from mistakes.

> what about the situation where calling balanceOf would actually empty your balance ?

This shouldn't even be a problem since `balance(_token)` uses a `STATICCALL`.

apps/agent/test/agent_shared.js
```diff
@@ -94,6 +100,144 @@ module.exports = (
...
+        context('> but target is a protected ERC20', () => {
+          it('it should revert', async () => {
+            const approve = token1.contract.approve.getData(target.address, 10)
+
+            await assertRevert(() => agent.safeExecute(token1.address, approve, { from: authorized }))
+          })
+        })
+
+        context('> and target is not a protected ERC20 but action affects a protected ERC20 balance', () => {
+          it('it should revert', async () => {
```
izqui: We could add a test case for the Agent's balance increasing as a result of the execution.

izqui: We should also test the execution modifying the protected token list which should cause a revert.

apps/agent/contracts/Agent.sol
```diff
@@ -14,22 +14,101 @@ import "@aragon/os/contracts/common/IForwarder.sol";
...
 
+    /**
+    * @notice Execute '`@radspec(_target, _data)`' on `_target` ensuring that protected tokens can't be spent
+    * @param _target Address where the action is being executed
+    * @param _data Calldata for the action
+    * @return Exits call frame forwarding the return data of the executed call (either error or success data)
+    */
+    function safeExecute(address _target, bytes _data) external authP(SAFE_EXECUTE_ROLE, arr(_target, uint256(getSig(_data)))) {
+        uint256 protectedTokensLength = protectedTokens.length;
+        address[] memory _protectedTokens = new address[](protectedTokensLength);
```
izqui:
```suggestion
        address[] memory protectedTokens_ = new address[](protectedTokensLength);
```

We normally use this notation when the variable is not an argument to the function (and we need to add an `_`)

osarrouy:
Added all the test cases plus some additionnal checks for ERC20.

For now the contract just calls `balance` on the token to check if the added token is an ERC20. The result is that if the token is not an ERC20 the error message is `SAFE_ERC_20_BALANCE_REVERTED`.

Do you think it's worth making a low level call to catch the error and return a clearer error message like `AGENT_PROTECTED_TOKEN_NOT_ERC20` ?

izqui:
@osarrouy I don't think it is worth it to do the low-level call, that error is fine IMO.

PR #949, opened 2019-08-16
Finance: verified tokens list update

sliux:


sohkai:
Thanks @sliux, but I've already done the same in #929 and we will be merging that quite soon :).

PR #952, opened 2019-08-27
Agent: add initial background script

sohkai:
Not yet tested, and has a few todos to complete (pending more work on aragon.js).

Passes to the frontend data in the form of:

```js
{
  balances: [
    // This is the same as Finance's; all numerical values are strings meant
    // to be converted into normal numbers or BN.js objects on the frontend
    {
      address,
      amount,
      decimals,
      name,
      symbol,
      verified
    }
  ],
  transactions: [
    {
      id,  // <transactionHash>.<transactionIndex>.<logIndex>; not meant to be shown to humans
      date,  // Similar to Finance's; in milliseconds as a normal number
      description,  // string
      safe,  // boolean; only true if `SafeExecute()` was used (for later use)
      tokenTransfers: [
        {
          amount,  // string; BN number
          from,  // address of sender if transfer was to the Agent app, null if not
          to,  // address of the recipient if transfer was from the Agent app, null if not 
          token,  // token address
        }
      ],
      transactionHash, // string bytes64 hash for the transaction
      type, // string; see transaction-types.js (one of 'direct deposit', 'direct transfer', or 'contract interaction')
    }
  ],
}
```

The `null` possibility for the `from` and `to` fields of the `tokenTransfers` is intended to help front-end filtering based on incoming / outgoing transactions, but can be changed to suit the frontend as necessary.

Note that `tokenTransfers` is an array that can hold multiple token transactions (a single transaction that executes a contract can include many internal token transfers). For flexibility / simplicity, it currently does not do any grouping on the transfers, although we may want to do so in the frontend (sum all amounts for the same token) to meet the current designs.

There are a few edge cases in the current reducing that will require more complex detection to provide good UX. One example is a single transaction invoking multiple Agent executions that each transfer tokens. Another is directly receiving ETH onto the Agent's proxy from a contract interaction, as it's difficult to correlate receiving ETH with a contract execution based solely on the events in the general case. In the current state, these cases will be presented slightly confusingly to users (most likely symptom is duplicate / multiple transaction entries for a single data).

---------

TODO:

- [ ] Test, especially decoding token transfers and receiving ETH through contract interactions
- [ ] Upgrade aragon.js to automatically include AppProxy events ABI (`ProxyDeposit` is not currently automatically decoded)
- [x] Upgrade aragon.js with API to fetch current app's contract address
- [x] Upgrade aragon.js with API for describing radspec
- [ ] Add description fallback with check to function signature registry


AquiGorka:
Transactions has a tokenTransfers array which means each transaction can have multiple token transfers, the designs expect one amount (colored to refer to deposit/withdrawal):

image

What would be the best approach here?

sohkai:
The best option might be to use the DataView's "entry expansion" to show multiple transactions.

If we don't want to do that now though, we could just say "multiple tokens" and "multiple" for source / recipient for the row.

AquiGorka:
How about this:
- If there is only 1 tx use that data
- If there's more use the word `multiple` as you suggested

apps/agent/app/src/script.js
```diff
@@ -47,10 +73,46 @@ const retryEvery = async (
...
 }
 
-retryEvery(() => {
+// There are no settings to check, so we do a throwaway check to `initialized()`
+// just to make sure we can connect properly
+retryEvery(() =>
+  // TODO: add call to fetch current agent app's address once available in aragon.js
+  //       and add it to settings
+  app
+    .call('initialized')
```
sohkai:
```suggestion
    .call('hasInitialized')
```

apps/agent/app/src/script.js
```diff
@@ -47,10 +73,46 @@ const retryEvery = async (
...
+    .toPromise()
+    .then(() => initialize())
+    .catch(_ => {
+      throw new Error(
+        'Could not start background script execution due to the contract not being connected to a network:'
```
sohkai:
```suggestion
        'Could not start background script execution due to the contract not being connected to a network'
```

sohkai:
Closing for #1082

PR #954, opened 2019-08-28
Updates for aragonUI

sohkai:
Fixes a few things that were updated in aragonUI directly.


apps/voting/app/src/components/VoteCard/VoteCard.js
```diff
@@ -45,66 +45,69 @@ const VoteCard = ({ vote, onOpen }) => {
...
-      css={`
-        display: grid;
-        grid-template-columns: 1fr;
-        grid-template-rows: auto 1fr auto auto;
-        grid-gap: ${1 * GU}px;
-        padding: ${3 * GU}px;
-      `}
-    >
+    <Card onClick={handleOpen}>
       <div
```
sohkai: The changes in https://github.com/aragon/aragon-ui/pull/533 seem to have broken this component's layout. I've now shifted the grid positioning to a new inner layout, but still having problems with long descriptions not automatically line breaking cc @AquiGorka @bpierre :

Screen Shot 2019-08-28 at 2 52 22 PM

apps/voting/app/src/components/AutoLink.js
```diff
@@ -3,9 +3,7 @@ import Linkify from 'react-linkify'
 import { Link } from '@aragon/ui'
 
 const AutoLink = ({ children }) => (
-  <Linkify component={Link} properties={{ target: '_blank' }}>
-    {children}
-  </Linkify>
+  <Linkify component={Link}>{children}</Linkify>
```
AquiGorka: Doesn't this remove the `target` prop on the link?

sohkai: `Link` applies it by default if an `href` is present

apps/voting/app/src/components/VoteCard/VoteCard.js
AquiGorka: Adding `css='display: block'` to the card should fix this.

bpierre: This should fix it https://github.com/aragon/aragon-ui/pull/541

There shouldn’t be any layout difference when switching to an interactive card, so let’s see if it’s entirely fixed now, otherwise we’ll move back to using a `div` with a custom focus ring.

sohkai: Now fixed with latest aragonUI.

apps/finance/app/src/components/EmptyFilteredTransfers.js
```diff
@@ -42,9 +42,7 @@ function EmptyFilteredTransfers({ onClear }) {
         `}
       >
         We can’t find any item matching your filter selection.{' '}
-        <ButtonText horizontalPadding="none" onClick={onClear}>
-          Clear filters
-        </ButtonText>
+        <Link onClick={onClear}>Clear filters</Link>
```
bpierre: :massage_man:

PR #967, opened 2019-09-03
Voting: add apps filter

sohkai:
Adds the apps filters to Voting. Requires a new release of all aragonAPI libraries.

Screen Shot 2019-09-03 at 6 14 59 PM

@dizzypaty I've added two mandatory options to this list:

- `Voting (this)`: selects votes without a script (i.e. polls)
- `External`: selects votes to target contracts outside of the current organization

The dropdown list is sorted alphabetically by name. None are currently loaded, but identifiers (no custom labels yet) will be shown as, e.g. `Tokens (ANT)`.

I was thinking it'd be nice to have external, but it's unlikely someone would be using the Voting app to make external calls at this point (they'd likely just be making interactions against an installed Agent app), so we can remove it.


dizzypaty:
@sohkai re your comments: 

- I think Voting (this) is a bit confusing, specially in the case of having multiple voting apps installed. You think it's not right to leave it just as 'Voting'? If not, how about Voting (polls) or Voting (questions)?
- It's nice to include a filter for External calls / contracts.

sohkai:
> I think Voting (this) is a bit confusing, specially in the case of having multiple voting apps installed. You think it's not right to leave it just as 'Voting'? If not, how about Voting (polls) or Voting (questions)?

Yeah, I'm not sure what the most clear way of expressing this will be. On an org with multiple voting apps (e.g. `governance`), having

```
Voting
Voting (50% 50%)
```

Would be quite confusing. `This app` doesn't really make much sense either.

dizzypaty:
Discussed offline for filter items: in the case of multiple app instances, the filter items should display the custom label. If custom labels are not available, we'll use `tagIndicator` to communicate which app instance users are currently interacting with. ([Figma file](https://www.figma.com/file/XW5MImCLiXn3OW1RYH9L5e/Dropdowns-headers-and-placeholders?node-id=0%3A1)

VOTING

sohkai:
@dizzypaty I'll work on the other dropdown-related issues in their own issue.

This is how the dropdown currently looks:

Screen Shot 2019-09-04 at 10 59 06 PM

---------

For now let's not add logic for detecting multiple app instances and refresh this in its own iteration afterwards (with custom label resolution too). The "this app" tag and identifiers in parenthesis will always appear in the filter.

PR #982, opened 2019-09-10
Voting copy changes

luisivan:


apps/voting/app/src/components/VoteActions.js
```diff
@@ -235,8 +233,10 @@ const TokenReference = ({
       </span>
     ) : (
       ''
-    )}{' '}
-    due to the vote starting at <strong>{formatDate(startDate)}</strong>.
+    )}
+    at block{' '}
+    <strong>{snapshotBlock}</strong>{' '}
+    due at <strong>{formatDate(startDate)}</strong>.
```
sohkai: @luisivan This change sounds a bit weird:

```
Voting with <balance> <symbol> (you currently hold <balance> <symbol) at block <block> due at <date>
```

The "at block" is confusing because it's the snapshot block and not the current block, and the "due at" is a bit weird (what's due?)

luisivan: solved at https://github.com/aragon/aragon-apps/commit/dd4e47d911cb146fb72d935df32cb0a553068044

PR #983, opened 2019-09-10
Voting copy fix

luisivan:


apps/voting/app/src/components/VoteActions.js
```diff
@@ -223,20 +223,11 @@ const TokenReference = ({
...
-        <strong>
-          {userBalanceNow} {tokenSymbol}
-        </strong>
-        )
-      </span>
-    ) : (
-      ''
-    )}
-    at block{' '}
+    due to snapshot taken at block{' '}
```
sohkai: due to **the** snapshot

apps/voting/app/src/components/VoteActions.js
```diff
@@ -223,20 +223,11 @@ const TokenReference = ({
...
     <strong>{snapshotBlock}</strong>{' '}
-    due at <strong>{formatDate(startDate)}</strong>.
+    at <strong>{formatDate(startDate)}</strong>.
+    Yout current balance is{' '}
```
sohkai: **Your**

But perhaps we should only show this if it's different than your snapshot balance?

PR #987, opened 2019-09-10
Vote: add rules to wrap links in container

AquiGorka:


apps/voting/app/src/components/VoteText.js
```diff
@@ -16,25 +16,34 @@ const VoteText = React.memo(
...
-              ) : (
-                <span key={index}>{part}</span>
-              )
-            )}
-            <br />
-          </React.Fragment>
-        ))}
+        <span
+          css={`
+            a {
```
sohkai: Which component is this selecting? If it's the badge, we should be able to style it directly?

PR #1006, opened 2019-09-19
Voting: tolerate undefined executionTargets

bpierre:
Just an attempt to fix the app crashing, until we find the root issue.


sohkai:
Hmm, perhaps this is due to https://github.com/aragon/aragon.js/issues/400; it's likely due to Firefox users not having their caches destroyed on the 0.8 update and then running into old votes that were cached and not updated.

PR #1012, opened 2019-10-02
Voting: fix a crash that was happening in certain cases

bpierre:
This change ensures `icon()` only gets called when it exists.

/cc @john-light

### DIFF
--- a/apps/voting/app/src/hooks/useVotes.js
+++ b/apps/voting/app/src/hooks/useVotes.js
@@ -5,12 +5,6 @@ import { VOTE_ABSENT } from '../vote-types'
 import { EMPTY_ADDRESS } from '../web3-utils'
 import useNow from './useNow'
 
-// Temporary fix to make sure executionTargets always returns an array, until
-// we find out the reason why it can sometimes be missing in the cached data.
-function getVoteExecutionTargets(vote) {
-  return vote.data.executionTargets || []
-}
-
 // Decorate the votes array with more information relevant to the frontend
 function useDecoratedVotes() {
   const { votes, connectedAccountVotes } = useAppState()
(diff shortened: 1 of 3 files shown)


apps/voting/app/src/hooks/useVotes.js
```diff
@@ -58,7 +59,14 @@ function useDecoratedVotes() {
         executionTargetData = {
           address: appAddress,
           name,
-          iconSrc: icon(24),
+
+          // This check ensures icon() only gets called if it actually exists,
+          // as users were reporting a crash happening in some cases. The
+          // problem seems to come from the multiple targets condition that
+          // was not including icon(), but this ensures it won’t crash anymore,
+          // until we can 100% reproduce the issue.
```
AquiGorka: Maybe enforce a shape on `targetApp`?
```
const TARGET_APP_SHAPE = {
  icon: () => null,
  identifier: '...'
}
targetApp = {
...TARGET_APP_SHAPE,
...targetApp
}
```

AquiGorka: And I was able to reproduce consistently: using Safari, no Frame, reload go to Voting -> Finance -> Voting, it happened every time.

bpierre: Oh cool, could you check if [this alone](https://github.com/aragon/aragon-apps/pull/1012/files#diff-e7b7e89ee1e9dda895d3987bea4d8532R39) solves the issue, so we don’t have to check for `icon()` to be a function?

AquiGorka: It does, that's the way I was fixing the problem until I saw this PR and that's why I suggested it too, there are 3 scenarios here, so instead of manually setting things in each case we can use the enforce object interface once :)

sohkai:
@bpierre I've added a few commits to hopefully simplify the "fallback" logic to handle both the issue here and https://github.com/aragon/aragon-apps/pull/1006.

PR #1013, opened 2019-10-02
Replace Identifier misspelling in license headers

ottodevs:
Even though it was fun to spot projects reusing or forking aragon contracts: https://github.com/search?q=SPDX-License-Identitifer&type=Code

But I thought it might be interesting to fix the misspelling, in case any automated script might eventually parse that header to show the license.

Related: https://github.com/aragon/aragonOS/pull/558


sohkai:
Thanks @ottodevs! I've just rebased this change against `next` since that's where any new contract changes go :).

PR #1016, opened 2019-10-03
Holders: set max width on smaller screens

AquiGorka:
Before:

After:
image

### DIFF
--- a/apps/token-manager/app/src/screens/Holders.js
+++ b/apps/token-manager/app/src/screens/Holders.js
@@ -11,6 +11,7 @@ import {
   IconRemove,
   Split,
   GU,
+  useLayout,
   useTheme,
 } from '@aragon/ui'
 import { formatBalance } from '../utils'
@@ -33,6 +34,8 @@ function Holders({
   tokenSymbol,
   tokenTransfersEnabled,
 }) {
+  const { layoutName } = useLayout()
+  const compact = layoutName === 'small'
   const connectedAccount = useConnectedAccount()
   const mappedEntries = useMemo(
     () => holders.map(({ address, balance }) => [address, balance]),
@@ -54,6 +57,7 @@ function Holders({
                 css={`
                   display: flex;
                   align-items: center;
+                  max-width: ${compact ? '50vw' : 'unset'};
                 `}
               >
                 <LocalIdentityBadge


AquiGorka:
> Probably something we should fix in DataView

Can we have the conversation about not using `table` in `DataView`? If we were using a `grid` users (devs) could pass their own templates for `rows` and `columns` and it would be much easier to work with overflows and with width adjustments.

bpierre:
@AquiGorka Of course! The table layout algorithm has some important downsides and using CSS Grid is not impossible, but it would end up being more complicated IMO (I actually started prototyping with the two approaches at the beginning). In the end, what we want is really close to what the table layout is doing, which is not surprising since we want to render a table here :laughing:

But it also makes other things challenging, like with the expansion. About the overflow issue, I might be wrong but I don’t think it is a very difficult problem, it’s just that it requires to be fixed at the component level rather than in every app using it. Once solved, I don’t see any other issue with the current approach really.

An important issue is that CSS Grid doesn’t allow table-like structures, like this:

```jsx
<div>
  <div>cell 1</div>
  <div>cell 2</div>
</div>
<div>
  <div>cell 1</div>
  <div>cell 2</div>
</div>
```

And requires this kind of structure instead:

```jsx
<div>cell 1</div>
<div>cell 2</div>
<div>cell 1</div>
<div>cell 2</div>
```

Which would break accessibility among other issues (like making an entire row clickable which is something we want to add). A solution could be to use a grid structure inside of every row, but we would lose the dynamic width set at the table level by the table layout algorithm.

But we could experiment more with it, using ARIA to restore accessibility, and perhaps using JS to restore certain parts of the table layout algorithm. About letting users set the row / column templates, I would like a solution a bit higher level: I’d really like to avoid having cells in random order, only because the custom row / column template is not set properly.

I think the best would be to start working on a prototype using CSS Grid to see how it could work?

AquiGorka:
Pierre, you can easily setup a table using `grid` with such markup:

```html
<div class="table">
  <div class="row">
    <div>cell 1</div>
    <div>cell 2</div>
  </div>
  <div class="row">
    <div>cell 1</div>
    <div>cell 2</div>
  </div>
</div>
```

```css
.table { 
display: grid;
grid-template-columns: auto;
}
.row { 
display: grid;
grid-template-columns: auto auto;
}
```

bpierre:
@AquiGorka This is what I am talking about when I say that it would lose the dynamic width set by the table layout algorithm. The width of the second row is totally independent from the first one in your example, and it doesn’t behave like columns.

AquiGorka:
I don't know if I follow but it would give more control to the implementer as they could provide the templates and set the dimensions for rows and columns within the rows. Even in my example some rules can be set so that both cells within the rows behave the same, even with different content of different sizes (making it way better to control overflows and wrapping).
As simple as:

```css
.row {
display: grid;
  grid-template-columns: 50% 50%; /* repeat(numRows, 100/numRows) */
}
```

bpierre:
> I understand there are different scenarios but grid is very powerful, we can definitely start testing a prototype.

I am not disagreeing with the fact that Grid is powerful :smile: it could definitely be a solution here. I am just pointing the reasons that made me choose the table layout instead (which also has its advantages!), based on what we are trying to achieve.

> As simple as:
>
> ```css
>  .row {
>    display: grid;
>    grid-template-columns: 50% 50%;
>  }
> ```

This wouldn’t work, as the cell widths need to be dynamic.

AquiGorka:
Yeah, there is no one silver-bullet template for everything but for every specific case we can find the appropriate template (combination of fr's, %'s, auto's and what not) and the best, is we can provide some defaults but leave it open for the devs which are the best suited to decide the dynamic sizing for their content.
So yeah, let's prototype 👍

bpierre:
Yes! :rocket::rocket::rocket:

PR #1020, opened 2019-10-08
Voting: do not display the sync indicator with the empty state

bpierre:


bpierre:
The empty state also contains a loading indicator :smile:

bpierre:
> Although perhaps we should note this as a general guideline for the syncing indicator. The empty state should contain a syncing check, and it should "overrule" a syncing indicator.

Yes, we should probably add it to the component documentation for both components. /cc @dizzypaty

PR #1021, opened 2019-10-17
Voting: fix a description overflow issue

bpierre:
Before (notice the scrollbars):

image

After:

image

### DIFF
--- a/apps/voting/app/src/components/VoteCard/VoteCard.js
+++ b/apps/voting/app/src/components/VoteCard/VoteCard.js
@@ -100,7 +100,8 @@ const VoteCard = ({ vote, onOpen }) => {
         css={`
           overflow: hidden;
           ${textStyle('body1')};
-          height: ${27 * 3}px; // 27px = line-height with body1
+          line-height: ${27}px; // 27px = line-height of textstyle('body1')
+          height: ${27 * 3}px; // 27px * 3 = line-height * 3 lines
           display: -webkit-box;
           -webkit-box-orient: vertical;
           -webkit-line-clamp: 3;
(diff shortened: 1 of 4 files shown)


apps/voting/app/src/components/VoteCard/VoteCard.js
```diff
@@ -97,11 +97,18 @@ const VoteCard = ({ vote, onOpen }) => {
...
           -webkit-box-orient: vertical;
           -webkit-line-clamp: 3;
           overflow: hidden;
+
+          // overflow-wrap:anywhere and hyphens:auto are not
+          // supported yet by the latest versions of Webkit / Blink,
+          // which is why word-break:break-word has been added.
+          hyphens: auto;
+          overflow-wrap: anywhere;
+          word-break: break-word;
```
sohkai: Should we put this in `VoteText` instead?

bpierre: Yes maybe we could: `VoteText` is inline but it’s only used as a block, the structure must be coming from a need in the previous version. Changing it :+1:

bpierre: Done in https://github.com/aragon/aragon-apps/pull/1021/commits/e6e78b7fa843f10a1099bbaee3b9fdd4f42f3f8e

apps/voting/app/src/components/VoteCard/VoteCard.js
```diff
@@ -92,28 +92,20 @@ const VoteCard = ({ vote, onOpen }) => {
...
+        prefix={<span css="font-weight: bold">#{voteId}: </span>}
+        text={description || metadata}
+        title={`#${voteId}: ${description || metadata}`}
         css={`
+          overflow: hidden;
           ${textStyle('body1')};
-          /* lines per font size per line height */
-          /* shorter texts align to the top */
-          height: ${10.5 * GU}px;
+          height: ${27 * 3}px; // 27px = line-height with body1
```
sohkai: Maybe we can add a note to the comment about `3` being the number of lines we want

PR #1025, opened 2019-10-23
Voting: specify which newVote() override to use

sohkai:
Fixes https://github.com/aragon/aragon/issues/1127.

Requires a soon-to-be-released patch from aragon.js.


apps/voting/app/src/App.js
```diff
@@ -5,7 +5,6 @@ import {
   IconPlus,
   Main,
   SyncIndicator,
-  GU,
```
sohkai: Purely linting change :)

PR #1032, opened 2019-11-02
Update "Create vote" notice

john-light:
- A "yes" vote is no longer automatically cast when a vote is created


sohkai:
This is not necessarily true; it does automatically vote 'yes' if you have the direct permission to create a vote, or if the forwarder has voting tokens.

PR #1046, opened 2019-11-22
Tokens: refactor the side panel as a function component

bpierre:
And make use of `useSidePanelFocusOnReady()`.


apps/token-manager/app/src/components/Panels/TokenPanelContent.js
```diff
@@ -17,233 +18,279 @@ import LocalIdentitiesAutoComplete from '../LocalIdentitiesAutoComplete/LocalIde
...
+  )
+}
 
-          <Field label="Number of tokens">
-            <TextInput.Number
-              value={amountField.value}
-              onChange={this.handleAmountChange}
-              min={minTokenStep}
-              max={amountField.max}
-              disabled={accountMaxed}
```
sohkai: ⚠️ We should keep the logic for `accountMaxed`, or find a different mechanism to handle it (e.g. handling it during `updateAmount()`).

Right now, the form doesn't stop you from trying to assigning or removing too many tokens.

apps/token-manager/app/src/components/Panels/TokenPanelContent.js
```diff
@@ -17,233 +18,279 @@ import LocalIdentitiesAutoComplete from '../LocalIdentitiesAutoComplete/LocalIde
...
+function TokenPanelContent({
+  getHolderBalance,
+  holderAddress,
+  maxAccountTokens,
+  mode,
+  onUpdateTokens,
+  tokenDecimals,
+  tokenDecimalsBase,
+}) {
+  const holderInputRef = useSidePanelFocusOnReady()
```
sohkai: I think it should be fairly easy to do with the `useSidePanel()` hook, but it would be nice if we instead focused on the amount input if the `holderAddress` was already populated.

bpierre: Done in https://github.com/aragon/aragon-apps/pull/1046/commits/5535c0b17d9e9b4970a08680f60679366ed91383

apps/token-manager/app/src/components/Panels/TokenPanelContent.js
bpierre: Done in https://github.com/aragon/aragon-apps/pull/1046/commits/5535c0b17d9e9b4970a08680f60679366ed91383

apps/token-manager/app/src/App.js
```diff
@@ -212,5 +204,10 @@ export default () => {
   const theme = useTheme()
   const { layoutName } = useLayout()
 
+  const themeMode = useThemeMode()
```
sohkai: Should we remove this for now?

bpierre: Oops this was for a local demo, it shouldn’t have been committed :o

bpierre: Done https://github.com/aragon/aragon-apps/pull/1046/commits/03b2aaa47db934e20f206ed73c66269fc29369d1

PR #1048, opened 2019-11-22
Finance: MCD DAI migration

sohkai:
Based off #1045 

- Updates the verified token lists
- Turns the "token information fallback" into an override instead, to handle SAI's token name and symbol
- Migrates to using `tokenIconUrl()` while gracefully handling testnet tokens


apps/finance/app/src/lib/icon-utils.js
```diff
@@ -0,0 +1,13 @@
...
+import { ETHER_TOKEN_VERIFIED_BY_SYMBOL } from './verified-tokens'
+
+// Small shim on top of @aragon/ui's tokenIconUrl, to handle our testnet tokens
+export const tokenIconUrl = (tokenAddress, tokenSymbol, networkType) => {
+  if (networkType === 'main') {
+    return _tokenIconUrl(tokenAddress)
+  }
+
+  // On other networks, pretend any token with the same symbol is legit
+  const mainnetEquivalent = ETHER_TOKEN_VERIFIED_BY_SYMBOL.get(tokenSymbol)
```
bpierre: Couldn’t that be an (minor) issue for people testing an org on rinkeby with a symbol in the list? What about doing it the other way, by having a list of test tokens by network?

apps/finance/app/src/lib/token-utils.js
```diff
@@ -18,21 +19,22 @@ const PRESET_TOKENS = new Map([
...
-const KNOWN_TOKENS_FALLBACK = new Map([
+const KNOWN_TOKENS_OVERRIDE = new Map([
   [
     'main',
     new Map([
       [
-        DAI_MAINNET_TOKEN_ADDRESS,
-        { symbol: 'DAI', name: 'Dai Stablecoin v1.0', decimals: '18' },
+        SAI_MAINNET_TOKEN_ADDRESS,
+        { symbol: 'SAI', name: 'Sai Stablecoin v1.0', decimals: '18' },
```
bpierre: Good to see they fixed that in MCD DAI!

apps/finance/app/src/lib/token-utils.js
```diff
@@ -44,18 +46,18 @@ export const isTokenVerified = (tokenAddress, networkType) =>
     ? ETHER_TOKEN_VERIFIED_ADDRESSES.has(tokenAddress.toLowerCase())
     : true
 
-export const tokenDataFallback = (tokenAddress, fieldName, networkType) => {
-  // The fallback list is without checksums
+export const tokenDataOverride = (tokenAddress, fieldName, networkType) => {
```
bpierre: :100:

apps/finance/app/src/lib/token-utils.js
sohkai: Yes, thankfully. I still don't know how they got to that bytes32 abomination.

apps/finance/app/src/lib/icon-utils.js
sohkai: Yes... we could add an extra check to make sure that the given token address is in the list of test tokens?

In that case, it would **only** be our test tokens that look "legit" with icons displayed in test networks.

sohkai: > we could add an extra check to make sure that the given token address is in the list of test tokens?

Done in 70d65d4.

PR #1051, opened 2019-11-26
Finance: app improvements

bpierre:
## Panel

The new SidePanel now unmounts its content when closed, so there is no need to reset the state internally anymore.

Also use useLayoutEffect() rather than setTimeout() to focus the text field.

## ToggleContent

- Use IconUp rather than a custom image.
- Use ButtonBase rather than a button.
- Move to a function component.
- Add a pressed state.
- Use the css prop.

### DIFF
--- a/apps/finance/app/src/components/ToggleContent.js
+++ b/apps/finance/app/src/components/ToggleContent.js
@@ -1,59 +1,60 @@
-import React from 'react'
-import styled from 'styled-components'
-import { springs, GU } from '@aragon/ui'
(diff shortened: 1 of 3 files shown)


apps/finance/app/src/components/ToggleContent.js
```diff
@@ -1,59 +1,60 @@
-import React from 'react'
```
sohkai: 🙌🙌🙌 We should eventually move this to aragonUI :D

PR #1058, opened 2019-12-18
Agent: add ERC721Receiver support

### LINKED ISSUE #1036: Agent: cannot handle ERC721 NFTs
[ERC721 tokens](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md) require contracts handling them to expose a `onERC721Received()` method that returns a magic value `bytes4(keccak256("onERC721Received(address,address,uint256,bytes)"))` on success.

We should upgrade the Agent to support receiving ERC721s.

izqui:
Close #1036, see https://github.com/aragonone/product/issues/251

To do:

- [x] Tests
- [x] Decide on release strategy (SaintFame needs this ASAP)

We will be releasing from `next`. Live contract on mainnet: [`0x6b0fa0f07a2269b13a42c354ce94920d5e283ff4`](https://etherscan.io/address/0x6b0fa0f07a2269b13a42c354ce94920d5e283ff4)


apps/agent/contracts/Agent.sol
```diff
@@ -35,6 +35,7 @@ contract Agent is IERC165, ERC1271Bytes, IForwarder, IsContract, Vault {
     uint256 public constant PROTECTED_TOKENS_CAP = 10;
 
     bytes4 private constant ERC165_INTERFACE_ID = 0x01ffc9a7;
+    bytes4 private constant ERC721_RECEIVED = 0x150b7a02; // bytes4(keccak256("onERC721Received(address,address,uint256,bytes)"))
```
sohkai: Maybe rename the variable to `ERC721_RECEIVED_INTERFACE_ID` to be consistent with the `ERC165_INTERFACE_ID`.

apps/agent/contracts/Agent.sol
```diff
@@ -35,6 +35,7 @@ contract Agent is IERC165, ERC1271Bytes, IForwarder, IsContract, Vault {
     uint256 public constant PROTECTED_TOKENS_CAP = 10;
```
sohkai: Agent should probably subclass from `IERC721Receiver`

apps/agent/contracts/Agent.sol
```diff
@@ -197,6 +199,12 @@ contract Agent is IERC165, ERC1271Bytes, IForwarder, IsContract, Vault {
         emit SetDesignatedSigner(msg.sender, oldDesignatedSigner, _designatedSigner);
     }
 
+    function onERC721Received(address _operator, address _from, uint256 _tokenId, bytes _data) external returns (bytes4) {
+        emit ReceiveERC721(msg.sender, _operator, _from, _tokenId, _data);
```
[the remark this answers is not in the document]
izqui: I would avoid this to decrease gas usage to be honest. To many back and forth calls for no good reason, we can always not act on these logs if the frontend sees that `msg.sender` is not conformant

PR #1059, opened 2019-12-21
Tokens: fix dark mode header

sohkai:
We were using `useTheme()` before Main was declared, leading to the theme not being reactive to the `appearance` set.

In general, I wonder if we should make it a rule that Main is used ahead of any other app logic, to avoid cases like this? For example, both Finance and Voting could fall into this trap if they were to use `useTheme()` in their App.js.


sohkai:
> I’m wondering if we shouldn’t throw an error if useTheme is used outside of Theme? It would prevent cases like this one that can be hard to debug.

Yes! Or at least a warning :).

-------

I really like the connector function!

PR #1066, opened 2020-01-14
Voting: avoid crashing when external app details are not loadable

sohkai:
Things like an app's name come from IPFS, and may not be available all the time.

The importance of re-spec'ing the data structures involved with apps is becoming clearer, as it'll let us more clearly mark whether something is always going to be available or not from aragon.js.

### DIFF
--- a/apps/voting/app/src/hooks/useVotes.js
+++ b/apps/voting/app/src/hooks/useVotes.js
@@ -80,7 +80,9 @@ function useDecoratedVotes() {
         identifier,
         name,
       }))
-      .sort((a, b) => a.name.localeCompare(b.name))
+      .sort((a, b) => {
+        return a.name ? a.name.localeCompare(b.name) : 1
+      })
 
     return [decoratedVotes, executionTargets]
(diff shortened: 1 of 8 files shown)


apps/voting/app/src/hooks/useVotes.js
```diff
@@ -53,11 +51,12 @@ function useDecoratedVotes() {
       if (targetApp) {
         const { appAddress, icon, identifier, name } = targetApp
         executionTargetData = {
+          identifier,
           address: appAddress,
-          name,
+          // If the app name was not loaded, use the app's address
+          name: name || appAddress,
```
bpierre: Wondering if we should use the shortened version here?

sohkai:
> Agreed, do you think it should happen in the background script?

This will need to happen starting from `@aragon/wrapper` and will propagate to `@aragon/api`.

apps/voting/app/src/hooks/useVotes.js
sohkai: Yes, let's do that!

PR #1070, opened 2020-01-20
Voting: wrap attempt to fetch execution date with exponential retries

sohkai:
Missed this `web3Eth` when doing the original exponential backoff in https://github.com/aragon/aragon-apps/pull/934. Didn't realize that asking for block numbers could also fail, and they do so rather silently (only a `null` is returned, rather than erroring).

### DIFF
--- a/apps/voting/app/src/script.js
+++ b/apps/voting/app/src/script.js
@@ -356,10 +356,26 @@ function loadVoteSettings() {
     })
 }
 
-async function loadBlockTimestamp(blockNumber) {
-  const { timestamp } = await app.web3Eth('getBlock', blockNumber).toPromise()
-  // Adjust for solidity time (s => ms)
-  return timestamp * 1000
+function loadBlockTimestamp(blockNumber) {
+  // Wrap with retry in case the block is somehow not present
+  return retryEvery(() =>
+    app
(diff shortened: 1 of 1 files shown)


apps/voting/app/src/script.js
```diff
@@ -356,10 +356,26 @@ function loadVoteSettings() {
...
+        const { timestamp } = block || {}
+        if (!timestamp) {
+          throw new Error(`Block ${blockNumber} not yet mined`)
+        }
+
+        // Adjust for solidity time (s => ms)
+        return timestamp * 1000
+      })
+      .catch(err => {
+        console.log(`Error fetching block (${blockNumber})`, err)
```
bpierre: I’m thinking that it could be nice to have an API to report errors through aragonAPI one day, so that we could have a journal of all the errors (client and per app) somewhere in the client.

sohkai: Oh yes absolutely; we do detect top-level errors from the script and [(I think) `store()`](https://github.com/aragon/aragon.js/blob/master/packages/aragon-api/src/index.js#L543) but we should definitely ask app developers to wrap their entire app and have a global error handler that pipes messages out to the client.

PR #1073, opened 2020-01-28
Agent: Add documentation

macor161:
- Add technical documentation for the Agent app
- Add Agent link to main readme file

### DIFF
--- a/apps/agent/README.md
+++ b/apps/agent/README.md
@@ -1,0 +1,128 @@
+# Agent
+
+
+
+## 1. Mechanism
+
+## 2. Architecture
+
+### 2.1 Inheritance and implementation
+
+#### 2.1.1 Vault
+
+#### 2.1.2 Forwarder
+
+#### 2.1.3 ERC165
+
+#### 2.1.4 ERC1271Bytes
+
+#### 2.1.5 IsContract
+
+### 2.2 Libraries
+
+SignatureValidator
+
+## 3. Entry points
+
+## 4. External interface
+
+### 4.1 Events
+
+The following events are emitted by `Agent`:
+
+#### 4.1.1 Execute 
+ - **Name:** `Execute`
+ - **Args:** 
+   - **Sender:** Address of the original sender
+   - **Target:** Address of the target contract
+   - **ETH value:** 
+   - **Data:** Transaction data
+
+#### 4.1.2 Safe execute
+ - **Name:** `SafeExecute`
+ - **Args:** 
+   - **Sender:** Address of the sender
+   - **Target:** Address of the target contract
+   - **Data:** Transaction data
+
+#### 4.1.3 Add protected token 
+ - **Name:** `AddProtectedToken`
+ - **Args:** 
+   - **Token:** Token address
+
+#### 4.1.4 Remove protected token 
+ - **Name:** `RemoveProtectedToken`
+ - **Args:** 
+   - **Token:** Token address
+
+#### 4.1.5 Presign hash 
+ - **Name:** `PresignHash`
+ - **Args:** 
+   - **Sender:** Address of the sender
+   - **Hash:** Presigned hash
+
+#### 4.1.6 Set designated signer
+ - **Name:** `SetDesignatedSigner`
+ - **Args:** 
+   - **Sender:** Address of the sender
+   - **Old signer:** Address of the previous signer
+   - **New signer:** Address of the new signer
+
+### 4.2 Getters
+
+#### 4.2.1 Is presigned
+ - **Inputs:** 
+   - **Hash:** Hash
+ - **Pre-flight checks:** None
+ - **Outputs:**
+   - **Is presigned**: Whether the hash is presigned or not.
+
+#### 4.2.2 Designated signer
+ - **Inputs:** None
+ - **Pre-flight checks:** None
+ - **Outputs:**
+   - **Designated signer**: Address of the designated signer
+
+#### 4.2.3 Protected token
+ - **Inputs:** 
+   - **Index:** Token index
+ - **Pre-flight checks:** None
+ - **Outputs:**
+   - **Protected tokens**: Address of the protected token
+
+#### 4.2.4 Is forwarder
+ - **Inputs:** None
+ - **Pre-flight checks:** None
+ - **Outputs:**
+   - **Is forwarder**: Whether `Agent` is a `Forwarder`. Always `true`.
+
+#### 4.2.5 Can forward
+ - **Inputs:** 
+   - **Sender:** Address of the account intending to forward an action
+   - **EVM script**: Script to forward
+ - **Pre-flight checks:** None
+ - **Outputs:**
+   - **Can forward**: Whether the action can be forwarded or not
+
+#### 4.2.6 Protected tokens length
+ - **Inputs:** None
+ - **Pre-flight checks:** 
+   - Ensure the contract is initialized
+ - **Outputs:**
+   - **Protected tokens length**: Number of protected tokens
+
+#### 4.2.7 Supports interface
+ - **Inputs:** 
+   - **Interface ID:** Interface bytes to check
+ - **Pre-flight checks:** None
+ - **Outputs:**
+   - **Supports interface**: Whether the contract supports the interface
+
+#### 4.2.8 Is valid signature
+ - **Inputs:** 
+   - **Hash:** Data signed on the behalf of address(this)
+   - **Signature:** Signature byte array associated with the data
+ - **Pre-flight checks:** None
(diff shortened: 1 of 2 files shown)


apps/agent/README.md
```diff
@@ -0,0 +1,128 @@
+# Agent
+
+
+
+## 1. Mechanism
```
sohkai: Perhaps "mechanism" will make more sense once we have this section written, but we may find a better wording :).

apps/agent/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+## 2. Architecture
```
sohkai: I think I would put notes on the architecture later (perhaps in an appendix or etc. section), as I don't think this is as immediately interesting as the entry points and external interfaces.

apps/agent/README.md
```diff
@@ -0,0 +1,128 @@
+# Agent
+
+
+
+## 1. Mechanism
+
```
sohkai: It may be interesting to include a section on "security features" when describing the app, as the Agent specifically comes with some (`safeExecute()`).

apps/agent/README.md
```diff
@@ -0,0 +1,254 @@
...
+### 1.3. Vault
+
+The Agent app is a superset of the [Vault](../vault/contracts/Vault.sol) contract, which means that it can store or transfer the tokens and Ether of the DAO in which it is installed. Specifically, the following functions are available: 
+
+ - `deposit(address _token, uint256 _value)` 
+ - `transfer(address _token, address _to, uint256 _value)` - Protected by the `TRANSFER_ROLE` permission.
+ - `balance(address _token)`
+
+More importantly, this property means that Agent is capable of using the DAO's funds to interact seemlessly with any contracts that require the sender to possess Ether or ERC-20 tokens (e.g. Exchanging tokens on [Uniswap](https://uniswap.io), participating in a [PoolTogether](https://www.pooltogether.com) lottery).
```
sohkai: I would also add the case for Voting (e.g. Aragon Network Votes) as well as using apps that require the vault, e.g. Finance.

apps/agent/README.md
```diff
@@ -0,0 +1,254 @@
+# Agent
+
+Agent is an Aragon app that has the ability to perform arbitrary calls to external contracts. It can be thought of as the external interface of a DAO. See our [guide](https://hack.aragon.org/docs/guides-use-agent) for concrete examples on how to use the Agent app.
```
sohkai: We may want to also note that the Agent can hold assets, like ETH, ERC20 tokens, and ERC712 NFTs.

apps/agent/README.md
```diff
@@ -0,0 +1,254 @@
...
+
+ - `deposit(address _token, uint256 _value)` 
+ - `transfer(address _token, address _to, uint256 _value)` - Protected by the `TRANSFER_ROLE` permission.
+ - `balance(address _token)`
+
+More importantly, this property means that Agent is capable of using the DAO's funds to interact seemlessly with any contracts that require the sender to possess Ether or ERC-20 tokens (e.g. Exchanging tokens on [Uniswap](https://uniswap.io), participating in a [PoolTogether](https://www.pooltogether.com) lottery).
+
+### 1.4. Security
+
+Since the Agent app can manage tokens and Ether, it is crucial to always keep security in mind. In order to restrict unauthorized transfers and other abusive behaviors, a variant of the `execute` function is available that will prevent any attempt to transfer tokens included in a "protected tokens" list:
```
sohkai:
```suggestion
Since the Agent app can manage tokens and Ether, it is crucial to always keep security in mind. In order to restrict unauthorized transfers and other abusive behaviors, a variant of the `execute()` function is available, and protected by a separate role, that will prevent any attempt to transfer tokens included in a "protected tokens" list:
```

apps/agent/README.md
```diff
@@ -0,0 +1,254 @@
...
+
+- `safeExecute(address _target, bytes _data)` - Protected by the `SAFE_EXECUTE_ROLE` permission.
+
+Before executing the requested call, the function fetches in memory every balances from the protected tokens and compares them with the new balances after the call is executed. If a balance has changed, it reverts the call.
```
sohkai:
```suggestion
The function fetches balances for every protected token before and after executing the function, making sure they are equal. If a balance has changed, it reverts the call.
```

apps/agent/README.md
```diff
@@ -0,0 +1,254 @@
...
+- `addProtectedToken(address _token)` - Protected by the `ADD_PROTECTED_TOKEN_ROLE` permission.
+- `removeProtectedToken(address _token)` - Protected by the `REMOVE_PROTECTED_TOKEN_ROLE` permission.
+
+### 1.5. Forwarder
+
+
+## 2. Entry points
+
+### 2.1. Initialize
+ - **Actor:** External entity, ultimately calling the function through the `Kernel`
```
sohkai: > "ultimately calling the function through the `Kernel`"

This is not necessarily true, although it is common the Kernel will be the one calling this.

To be honest, I would consider moving the "initialization" of all apps into a separate function away from the entry points. Perhaps a "construction" or "initialization" section, that also details that you can't use normal constructors due to the proxy pattern.

apps/agent/README.md
```diff
@@ -0,0 +1,254 @@
...
+ - **State transitions:**
+   - Initialize the app
+
+### 2.2. Execute
+ - **Actor:** External entity
+ - **Inputs:** 
+   - **Target:** Address where the action is being executed
+   - **ETH value:** Amount of ETH (in Wei) sent with the action
+   - **Data:** Calldata for the action
+ - **Pre-flight checks:**
```
sohkai: I don't remember exactly, but there was some discussion around making sure `msg.data > 0` because otherwise you'd just use `.transfer()` to transfer ETH. We may want to highlight that behaviour.

apps/agent/README.md
```diff
@@ -0,0 +1,254 @@
...
+ - **State transitions:** None
+
+### 2.4. Add protected token
+ - **Actor:** External entity
+ - **Inputs:** 
+   - **Token:** Address of the protected token
+ - **Pre-flight checks:**
+   - Ensure the actor has the `ADD_PROTECTED_TOKEN_ROLE` permission with the following ACL parameters:
+     - Token address
+   - Ensure that the current number of protected tokens is inferior than the maximum allowed (hardcoded as a constant)
```
sohkai:
```suggestion
   - Ensure that the current number of protected tokens is less than the maximum allowed (hardcoded as a constant)
```

Maybe we should replace the "the maximum allowed" with the actual hardcoded number.

apps/agent/README.md
```diff
@@ -0,0 +1,254 @@
...
+ - **Outputs:**
+   - **Is valid signature bytes**: The `ERC-1271` magic value if the signature is valid
+
+
+## 4. Architecture
+
+### 4.1. Inheritance and implementation
+
+- [Vault](../vault/contracts/Vault.sol)
+- [Forwarder](https://github.com/aragon/aragonOS/blob/master/contracts/common/IForwarder.sol)
```
sohkai: Not a strong opinion; maybe linking to the hack.aragon.org documentation on forwarders would be better?

apps/agent/README.md
macor161: Good idea! It's indeed a major feature.

apps/agent/README.md
macor161: Oops, classic french to english translation mistake 😆

macor161:
I didn't have time to fix all the requested modifications, there are 2 or 3 left.

apps/agent/README.md
macor161: hmm.. I'm not sur I understand what you mean 🤔 Making sure `msg.data > 0` or making sure `_data > 0`?

apps/agent/README.md
macor161: I agree, I moved the function to the architecture section. We could eventually move it to a separate file when technical documentation is available for other Aragon apps.

apps/agent/README.md
sohkai: This was the change: https://github.com/aragon/aragon-apps/pull/625

And the currently deployed contract: https://github.com/aragon/aragon-apps/blob/master/apps/agent/contracts/Agent.sol#L71

I remembered we had a discussion about whether to enforce `execute` to have data, but we removed it in the end :).

apps/agent/README.md
macor161: I added the Finance example to the `Vault` section but added the Aragon Network example to the `Forwarder` section. Also mentioned interaction with other DAOs in `1.1. Arbitrary call execution`.

PR #1076, opened 2020-02-04
Tokens: tree shaking

bpierre:
Bundle size goes from 348.15 KB to 196.73 KB with this change.

### DIFF
--- a/apps/token-manager/app/package.json
+++ b/apps/token-manager/app/package.json
@@ -36,7 +36,7 @@
     "eslint-plugin-react": "^7.5.1",
     "eslint-plugin-standard": "^4.0.0",
     "jest": "^23.6.0",
-    "prettier": "^1.8.2"
+    "prettier": "^1.8.2",
     "parcel-bundler": "1.12.4",
     "parcel-plugin-bundle-visualiser": "^1.2.0",
   },
--- a/apps/token-manager/app/package.json
+++ b/apps/token-manager/app/package.json
@@ -38,7 +38,7 @@
     "jest": "^23.6.0",
     "prettier": "^1.8.2",
     "parcel-bundler": "1.12.4",
-    "parcel-plugin-bundle-visualiser": "^1.2.0",
+    "parcel-plugin-bundle-visualiser": "^1.2.0"
   },
   "scripts": {
     "test": "jest",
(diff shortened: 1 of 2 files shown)


apps/token-manager/app/src/web3-utils.js
```diff
@@ -1,4 +1,4 @@
-import { isAddress } from 'web3-utils'
+import { isAddress } from '@aragon/ui'
```
bpierre: [`isAddress()` from aragonUI](https://github.com/aragon/aragon-ui/blob/d668f801e8c3eea8ace21afdc4eeeaaedfaf4adf/src/utils/web3.js#L127-L137) is not as complete as [`isAddress()` from web3-utils](https://github.com/ethereum/web3.js/blob/1.x/packages/web3-utils/src/utils.js#L81-L99), but we are already using `js-sha3` in aragonUI so it should be easy to implement in a similar way.

We should ensure it passes these tests: https://github.com/ethereum/web3.js/blob/1.x/test/utils.isAddress.js

apps/token-manager/app/package.json
```diff
@@ -37,14 +36,15 @@
     "eslint-plugin-react": "^7.5.1",
     "eslint-plugin-standard": "^4.0.0",
     "jest": "^23.6.0",
-    "parcel-bundler": "1.9.7",
     "prettier": "^1.8.2"
```
sohkai:
```suggestion
    "prettier": "^1.8.2",
```

apps/token-manager/app/package.json
```diff
@@ -37,14 +36,15 @@
...
+    "parcel-bundler": "1.12.4",
+    "parcel-plugin-bundle-visualiser": "^1.2.0",
```
sohkai:
```suggestion
    "parcel-plugin-bundle-visualiser": "^1.2.0"
```

apps/token-manager/app/package.json
bpierre: Oops, manual hunk edit :see_no_evil:

sohkai:
@bpierre Wondering if we should close this for now, given the experimental state of treeshaking in aragonUI (and that the treeshaking wasn't working for Finance)?

bpierre:
Yes let’s close it for now, we’ll open another PR when it will be a bit more reliable.

PR #1079, opened 2020-02-07
Finance: update tokens

bpierre:


apps/finance/app/src/App.js
```diff
@@ -20,12 +20,11 @@ class App extends React.Component {
   static propTypes = {
     api: PropTypes.object,
     appState: PropTypes.object,
+    compactMode: PropTypes.bool,
+    isSyncing: PropTypes.bool,
   }
   static defaultProps = {
     isSyncing: true,
-    balances: [],
```
sohkai: Note to self: these aren't passed in anymore, as we now fetch the state from `api-react`'s `appState`.

Issue #1065, opened 2020-01-10
Finance: add thousands separators to token balances module


izqui:
Quite hard to read at a glance otherwise:

<img width="666" alt="Screen Shot 2020-01-10 at 5 06 08 PM" src="https://user-images.githubusercontent.com/447328/72167498-828edd80-33cb-11ea-8807-52b2879ebd55.png">

Evalir:
This seems to have been fixed already. :)

Finance (on network dao):
![Screen Shot 2020-04-06 at 5.25.33 PM.png](https://images.zenhubusercontent.com/5931bd7e51aa2c3057804854/424c981a-b6e7-41d4-8b2f-a168a089d9a3)

sohkai:
I think that might've been off of the PR fixing this?

It is currently not fixed on production for me:
![Screen Shot 2020-04-07 at 2.19.58 PM.png](https://images.zenhubusercontent.com/569e0df9638e718c996030c9/7aac31d1-92e9-4e3b-af84-c4c4aaa47c3a)

This will require the API from https://github.com/aragon/aragon-ui/pull/728 :).

PR #1081, opened 2020-02-13
Finance: Thousands Separators

ECWireless:
Fixing issue #1065

### DIFF
--- a/apps/finance/app/src/lib/utils.js
+++ b/apps/finance/app/src/lib/utils.js
@@ -28,3 +28,8 @@ export function formatTokenAmount(
     formatDecimals(round(amount / Math.pow(10, decimals), rounding), 18)
   )
 }
+
+// Thousands Separator
+export function separateThousands(x) {
+  return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+}
--- a/apps/finance/app/src/lib/utils.js
+++ b/apps/finance/app/src/lib/utils.js
@@ -31,5 +31,7 @@ export function formatTokenAmount(
 
 // Thousands Separator
 export function separateThousands(x) {
-  return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
+  var parts = x.toString().split(".");
+  parts[0] = parts[0].replace(/\B(?=(\d{3})+(?!\d))/g, ",");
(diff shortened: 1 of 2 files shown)


apps/finance/app/src/lib/utils.js
```diff
@@ -28,3 +28,8 @@ export function formatTokenAmount(
     formatDecimals(round(amount / Math.pow(10, decimals), rounding), 18)
   )
 }
+
+// Thousands Separator
+export function separateThousands(x) {
+  return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",");
```
sohkai: This doesn't work if the string has a decimal, e.g. (`'1.23456'`).

We should do the replace on only the whole number ahead of the decimal.

ECWireless: Okay, I just made a new commit that should fix this.

sohkai:
Actually @ECWireless, @bpierre and I were discussing this, and realized from [Ethers that the regex may not work well everywhere](https://github.com/ethers-io/ethers.js/blob/061b0eae1d4c570aedd9bee1971afa43fcdae1a6/src.ts/utils/units.ts#L69-L107).

We started talking about including some formatting utilities in aragonUI instead to do this: https://github.com/aragon/aragon-ui/issues/719

Evalir:
Considering we will be adding formatting utilities in aragonUI, should we close this? @sohkai @bpierre

sohkai:
Let's keep it here and update it when a new aragonUI version is released with the utility.

ECWireless:
@sohkai Is there a timeline for when the new AragonUI version is supposed to be released? It looks like a new one is released roughly every 2 months, but figured I'd ask...

sohkai:
@ECWireless 

> Is there a timeline for when the new AragonUI version is supposed to be released? It looks like a new one is released roughly every 2 months, but figured I'd ask...

Haha yes... we have not been very good at making incremental releases lately but we've included a bunch of updates to aragonUI and will be releasing soon.

@bpierre and I were discussing the formatting API on Friday, so we might wait just a bit longer on this—let's see.

ECWireless:
Closing in favor of #1145, since this PR uses a forked repo that has been deleted.

PR #1086, opened 2020-03-03
Finance: Add proper DataView empty state

### LINKED ISSUE #1085: Finance: use DataView empty state
Rather than a [custom component](https://github.com/aragon/aragon-apps/pull/1082#discussion_r386157377).

Example with Agent: https://github.com/aragon/aragon-apps/commit/2ded25b2904dc919ef513e91452df48fb6e3f745

Evalir:
Fixes #1085 .

Also shifts the `Syncindicator` to not appear under the HelpScout button.


apps/finance/app/src/components/Transfers.js
```diff
@@ -199,13 +202,28 @@ const Transfers = React.memo(({ tokens, transactions }) => {
     [filteredTransfers, compareDesc]
   )
 
-  if (!transactions.length) {
-    return <EmptyTransactions />
-  }
+  const dataViewStatus = useMemo(() => {
+    if (emptyResultsViaFilters && transactions.length > 0) {
```
Evalir: Here I noticed we actually have to check for the second condition, because if not, when the DataView is empty we'll get a non-accurate "empty-filters" screen, which is not right.

apps/finance/app/src/components/EmptyTransactions.js
```diff
@@ -1,56 +0,0 @@
-import React from 'react'
-import { useAragonApi } from '@aragon/api-react'
-import { Box, LoadingRing, GU, useTheme } from '@aragon/ui'
-import noTransfersPng from './assets/no-transfers.png'
```
bpierre: We should also remove this asset :-)

Evalir: Aaaah sure!

apps/finance/app/src/components/Transfers.js
```diff
@@ -199,13 +202,28 @@ const Transfers = React.memo(({ tokens, transactions }) => {
...
+      return 'empty-filters'
+    }
+    if (appState.isSyncing) {
+      return 'loading'
+    }
+    return 'default'
+  }, [isSyncing, emptyResultsViaFilters])
```
bpierre: We should add `transactions` here.

apps/finance/app/src/components/Transfers.js
bpierre: Ah yes you’re right!

apps/finance/app/src/components/Transfers.js
```diff
@@ -227,17 +245,19 @@ const Transfers = React.memo(({ tokens, transactions }) => {
...
+                <Button onClick={handleDownload}>
+                  <IconExternal
+                    css={`
+                      margin-right: ${1.5 * GU}px;
+                      color: ${theme.surfaceIcon};
+                    `}
+                  />{' '}
+                  Export
+                </Button>
+              </div>
```
bpierre: We could refresh this part a bit since we are here:

```suggestion
              <div>
                <Button
                  icon={<IconExternal />}
                  label="Export"
                  onClick={handleDownload}
                />
              </div>
```

apps/finance/app/src/App.js
```diff
@@ -96,7 +96,7 @@ class App extends React.Component {
         onResolve={this.handleResolveLocalIdentity}
         onShowLocalIdentityModal={this.handleShowLocalIdentityModal}
       >
-        <SyncIndicator visible={isSyncing} />
+        <SyncIndicator visible={isSyncing} shift={50} />
```
bpierre: We should bump the aragonUI’s version to get the new SyncIndicator :-)

Evalir: Oooh seems I have it linked 🤦‍♂️

sohkai: Wondering if we should apply this change in the other apps as well?

PR #1102, opened 2020-03-19
Finance: port agent filtering to finance

### LINKED ISSUE #1100: Finance: port Agent filtering
The finance app should use the revamped version of the filtering that was made for the new Agent UI.

Evalir:
Closes #1100 .

Ports agent filtering to finance, using hooks and consolidating all state regarding the filters in the `useFilteredTransfers` hook. Minor modifications were made due to differences in the `script.js`, mainly in detecting the transaction type. (`Incoming`, or `Outgoing`)

### DIFF
--- a/apps/finance/app/src/components/Transfers.js
+++ b/apps/finance/app/src/components/Transfers.js
@@ -1,12 +1,6 @@
-import React, { useMemo, useState, useCallback, useContext } from 'react'
+import React, { useMemo, useCallback, useContext } from 'react'
 import PropTypes from 'prop-types'
-import {
-  compareDesc,
-  endOfDay,
-  format,
-  isWithinInterval,
-  startOfDay,
-} from 'date-fns'
+import { compareDesc, format } from 'date-fns'
 import {
   Button,
   ContextMenu,
@@ -29,54 +23,16 @@ import {
   useNetwork,
 } from '@aragon/api-react'
 import { saveAs } from 'file-saver'
-import * as TransferTypes from '../transfer-types'
+import { READABLE_TRANSFER_TYPES } from '../transfer-types'
 import { addressesEqual, toChecksumAddress } from '../lib/web3-utils'
 import { formatTokenAmount } from '../lib/utils'
 import TransfersFilters from './TransfersFilters'
 import { useIdentity, IdentityContext } from './IdentityManager/IdentityManager'
 import LocalIdentityBadge from './LocalIdentityBadge/LocalIdentityBadge'
+import useFilteredTransfers from './useFilteredTransfers'
 
-const UNSELECTED_TOKEN_FILTER = -1
-const UNSELECTED_TRANSFER_TYPE_FILTER = -1
-const INITIAL_DATE_RANGE = { start: null, end: null }
-const TRANSFER_TYPES = [
-  TransferTypes.All,
-  TransferTypes.Incoming,
-  TransferTypes.Outgoing,
-]
-const TRANSFER_TYPES_STRING = TRANSFER_TYPES.map(TransferTypes.convertToString)
 const formatDate = date => format(date, 'dd/MM/yy')
-const getTokenDetails = (details, { address, decimals, symbol }) => {
-  details[toChecksumAddress(address)] = {
-    decimals,
-    symbol,
-  }
-  return details
-}
-// Filter transfer based on the selected filters
-const getFilteredTransfers = ({
-  transactions,
-  selectedToken,
-  selectedTransferType,
-  selectedDateRange,
-}) => {
-  const transferType = TRANSFER_TYPES[selectedTransferType]
-  return transactions.filter(
-    ({ token, isIncoming, date }) =>
-      (!selectedDateRange.start ||
-        !selectedDateRange.end ||
-        isWithinInterval(new Date(date), {
-          start: startOfDay(selectedDateRange.start),
-          end: endOfDay(selectedDateRange.end),
-        })) &&
-      (selectedToken === null ||
-        addressesEqual(token, selectedToken.address)) &&
-      (transferType === TransferTypes.All ||
-        selectedTransferType === UNSELECTED_TRANSFER_TYPE_FILTER ||
-        (transferType === TransferTypes.Incoming && isIncoming) ||
-        (transferType === TransferTypes.Outgoing && !isIncoming))
-  )
-}
+
 const getDownloadData = async (transfers, tokenDetails, resolveAddress) => {
   const mappedData = await Promise.all(
     transfers.map(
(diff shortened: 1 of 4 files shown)


apps/finance/app/src/components/useFilteredTransfers.js
```diff
@@ -0,0 +1,106 @@
...
+  }, [])
+
+  const tokensToFilter = useMemo(() => [{ symbol: 'All tokens' }, ...tokens], [
+    tokens,
+  ])
+
+  const filteredTransfers = useMemo(
+    () =>
+      transactions.filter(({ date, isIncoming, token }) => {
+        const type = isIncoming ? Incoming : Outgoing
```
Evalir: I'm not feeling super good about this particular part; it seems fine for what it does, but maybe there is a better way to get the type, as for the transactions we only get an `isIncoming` boolean.

apps/finance/app/src/transfer-types.js
```diff
@@ -1,21 +1,12 @@
-export const All = Symbol('All')
-export const Incoming = Symbol('Incoming')
-export const Outgoing = Symbol('Outgoing')
+export const All = 'ALL_TRANSFER'
```
sohkai, line 1: In this case, we may still want to use symbols (prefer them over strings when possible).

There's another comment below about whether we could use these in the script, but otherwise, let's revert this back to being a symbol.

apps/finance/app/src/components/Transfers.js
```diff
@@ -259,13 +196,12 @@ const Transfers = React.memo(({ tokens, transactions }) => {
...
-              tokenFilter={selectedToken}
               onTokenChange={handleTokenChange}
-              transferTypeFilter={selectedTransferType}
               onTransferTypeChange={handleTransferTypeChange}
-              compactMode={compactMode}
-              symbols={['All tokens', ...symbols]}
-              transferTypes={TRANSFER_TYPES_STRING}
+              tokenFilter={selectedToken}
+              transferTypeFilter={selectedTransferType}
+              transferTypes={READABLE_TRANSFER_TYPES}
```
sohkai, line 203: Perhaps this value should be controlled by `useFilteredTransfers()`, to encapsulate all the filtering logic to be within it.

It would be bad if we re-ordered the list but forgot to update it here or in that hook, which suggests we should encapsulate the concepts :).

apps/finance/app/src/transfer-types.js
```diff
@@ -1,21 +1,12 @@
...
 }
 
-export function convertFromString(str) {
-  return symbolMapping[str]
-}
-export function convertToString(symbol) {
-  return stringMapping[symbol]
-}
+export const TRANSFER_TYPES = Object.keys(TRANSFER_TYPES_LABELS)
+export const READABLE_TRANSFER_TYPES = Object.values(TRANSFER_TYPES_LABELS)
```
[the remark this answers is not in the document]
sohkai: Perhaps `TRANSFER_TYPES_LABELS`? It looks like we don't need to export the mapping.

apps/finance/app/src/components/useFilteredTransfers.js
```diff
@@ -0,0 +1,106 @@
...
+import { endOfDay, isWithinInterval, startOfDay } from 'date-fns'
+import { TRANSFER_TYPES, Incoming, Outgoing } from '../transfer-types'
+import { addressesEqual } from '../lib/web3-utils'
+
+const UNSELECTED_TOKEN_FILTER = -1
+const UNSELECTED_TRANSFER_TYPE_FILTER = -1
+const UNSELECTED_DATE_RANGE_FILTER = { start: null, end: null }
+
+function useFilteredTransfers({ transactions, tokens }) {
+  const [page, setPage] = useState(0)
```
sohkai, line 11: Just wondering, do we control the page like this in Agent too?

It's kind of related... but it also feels like it doesn't really belong here since this hook should just be controlling the filtering.

apps/finance/app/src/components/useFilteredTransfers.js
```diff
@@ -0,0 +1,106 @@
...
+  const [selectedDateRange, setSelectedDateRange] = useState(
+    UNSELECTED_DATE_RANGE_FILTER
+  )
+  const [selectedTransferType, setSelectedTransferType] = useState(
+    UNSELECTED_TRANSFER_TYPE_FILTER
+  )
+  const [selectedToken, setSelectedToken] = useState(UNSELECTED_TOKEN_FILTER)
```
sohkai, line 18: Nitpick: let's add a space before the `useCallback()`s here, to delineate the state and callbacks.

apps/finance/app/src/components/useFilteredTransfers.js
```diff
@@ -0,0 +1,106 @@
...
+  const handleTokenChange = useCallback(index => {
+    setPage(0)
+    setSelectedToken(index || UNSELECTED_TOKEN_FILTER)
+  }, [])
+  const handleTransferTypeChange = useCallback(index => {
+    setPage(0)
+    setSelectedTransferType(index || UNSELECTED_TRANSFER_TYPE_FILTER)
+  }, [])
+  const handleClearFilters = useCallback(() => {
+    setPage(0)
```
sohkai, line 32: Rather than repeating `setPage()` here, it would likely be nicer to run as a `useEffect()` instead—any time the filters change, we reset the page back to 0.

(Assuming we keep `page` here)

apps/finance/app/src/components/useFilteredTransfers.js
sohkai: Ah, where do we get `isIncoming`? Is it from the script? If so, we could use the type directly there or marshall it in the `app-state-reducer` so we avoid this.

apps/finance/app/src/components/useFilteredTransfers.js
```diff
@@ -0,0 +1,106 @@
...
+  ])
+
+  const filteredTransfers = useMemo(
+    () =>
+      transactions.filter(({ date, isIncoming, token }) => {
+        const type = isIncoming ? Incoming : Outgoing
+        // Exclude by transaction type
+        if (
+          selectedTransferType !== -1 &&
+          TRANSFER_TYPES[selectedTransferType] !== type
```
sohkai, line 49: Is `ALL` handled correctly here?

apps/finance/app/src/components/useFilteredTransfers.js
```diff
@@ -0,0 +1,106 @@
...
+        ) {
+          return false
+        }
+
+        // Exclude by date range
+        if (
+          // We're not checking for an end date because we will always
+          // have a start date for defining a range.
+          selectedDateRange.start &&
```
sohkai, line 58: I mentioned this last time in Agent too, but I think we should make this test more robust to be able to handle any type of input with the date ranges (e.g. has one, has both, has none).

We should treat this hook as being independent with the view, and it will be better practice to handle this state correctly from the get go rather than forgetting to change this (very easy to do!) if we ever change the date range picker's behaviour.

apps/finance/app/src/components/Transfers.js
Evalir: > Perhaps this value should be controlled by useFilteredTransfers(), to encapsulate all the filtering logic to be within it.

Now that I think about it, it makes sense, as `symbols` is handled by `useFilteredTransfers()` itself. Now, would we just import the labels and export them back with another name?

apps/finance/app/src/components/useFilteredTransfers.js
Evalir: Yes, it's done this way as well. I took this decision because we have to reset the page if any of the filtering options change; originally this was done on the component holding the DataView, but as we had to handle each case, it felt like it was leaking filtering logic into the DataView component.

apps/finance/app/src/components/useFilteredTransfers.js
Evalir: Let's see how this looks inside the hook!

apps/finance/app/src/components/useFilteredTransfers.js
Evalir: Yes; the `handleTransferTypeChange` makes `selectedTransferType` become `-1` when it gets a `0`. I'm not liking how this is using implicit knowledge about `0` being `falsy` now though, so I'll rewrite the handlers to use a ternary to explicitly assign the value.

apps/finance/app/src/components/useFilteredTransfers.js
Evalir: Ah yes! This in fact was changed on Agent; this PR was just not updated.

apps/finance/app/src/components/useFilteredTransfers.js
Evalir: Marshalling it in the `app-state-reducer` sounds good 👍 , let's do this in another PR.

apps/finance/app/src/components/Transfers.js
sohkai: Yep, that's what I would've done!

apps/finance/app/src/components/useFilteredTransfers.js
sohkai: Yes, the alternative would be running it as an effect in the component using this hook, such that pages is only dependent on the transaction list changing.

apps/finance/app/src/components/useFilteredTransfers.js
```diff
@@ -60,19 +60,35 @@ function useFilteredTransfers({ transactions, tokens }) {
...
+        // Case 2: we have a start date, but not an end date.
+        if (
+          selectedDateRange.start &&
+          !selectedDateRange.end &&
+          !isEqual(new Date(date), selectedDateRange.start)
+        ) {
+          return false
+        }
+
+        // Case 3: we don't have a start date, but do have an end date.
```
sohkai, line 85: I'm not sure I get these changes...

What I might do instead is filter separately by start date and end date, with `isAfter()` and `isBefore()`.

E.g.

```js
if (selectedDateRange.start && isBefore(new Date(date), startOfDay(selectedDateRange.start)) {
  return false
}
// and vice-versa for the end date
```

apps/finance/app/src/transfer-types.js
```diff
@@ -2,11 +2,16 @@ export const All = Symbol('ALL_TRANSFER')
...
 
-export const TRANSFER_TYPES = Object.keys(READABLE_TRANSFER_TYPES)
-export const TRANSFER_TYPES_LABELS = Object.values(READABLE_TRANSFER_TYPES)
+// As we're using symbols as object keys,
+// we need to use Reflect.ownKeys to properly retrieve and iterate
+// over the object values.
+export const TRANSFER_TYPES = Reflect.ownKeys(READABLE_TRANSFER_TYPES)
+export const TRANSFER_TYPES_LABELS = Reflect.ownKeys(
+  READABLE_TRANSFER_TYPES
+).map(type => READABLE_TRANSFER_TYPES[type])
```
sohkai, line 17: Rather than having to do this, we can instead use an array of tuples and work on that instead.

E.g.

```js
const AVAILABLE_TRANSFER_TYPES = [
  [All, 'All'], [Incoming, 'Incoming'], [Outgoing, 'Outgoing']
]

export const TRANSFER_TYPES = AVAILABLE_TRANSFER_TYPES.map(([type]) => type)
export const TRANSFER_TYPES_LABELS = AVAILABLE_TRANSFER_TYPES.map(([_, label]) => label)
```

apps/finance/app/src/components/useFilteredTransfers.js
```diff
@@ -32,10 +32,12 @@ function useFilteredTransfers({ transactions, tokens }) {
     setSelectedDateRange(range)
   }, [])
   const handleTokenChange = useCallback(index => {
-    setSelectedToken(index || UNSELECTED_TOKEN_FILTER)
+    const tokenIndex = index === 0 ? -1 : index
```
sohkai, line 35:
```suggestion
    const tokenIndex = index === 0 ? UNSELECTED_TOKEN_FILTER : index
```

Maybe this is more clear and uses the constant.

apps/finance/app/src/transfer-types.js
Evalir: Looks way cleaner!

PR #1103, opened 2020-03-19
Finance: add useDownloadData hook

### LINKED ISSUE #1087: Finance: escape CSV data + build CSV on download
See https://github.com/aragon/aragon-apps/pull/1082/commits/c8ac5695f86b0dc2db7659693d2ddeeb177e14f7 and https://github.com/aragon/aragon-apps/pull/1082/commits/813d22f8b5fe85c62fc35bdc2edb7d069670b363 from the Agent app implementation.

Evalir:
Builds on #1102 (should be merged first) , closes #1087 .

Adds an useDownloadData hook that builds the CSV on demand, properly formatting it to fit the finance usage. Ported over from the Agent implementation.

### DIFF
--- a/apps/finance/app/src/components/Transfers.js
+++ b/apps/finance/app/src/components/Transfers.js
@@ -14,7 +14,6 @@ import {
   textStyle,
   useLayout,
   useTheme,
-  useToast,
 } from '@aragon/ui'
 import {
   useAragonApi,
@@ -35,7 +34,6 @@ const Transfers = React.memo(({ tokens, transactions }) => {
   const connectedAccount = useConnectedAccount()
   const { layoutName } = useLayout()
   const theme = useTheme()
-  const toast = useToast()
 
   const {
     emptyResultsViaFilters,
@@ -66,7 +64,6 @@ const Transfers = React.memo(({ tokens, transactions }) => {
   const { handleDownload } = useDownloadData({
     filteredTransfers,
     selectedDateRange,
-    toast,
     tokenDetails,
   })
 
--- a/apps/finance/app/src/components/Transfers.js
+++ b/apps/finance/app/src/components/Transfers.js
@@ -1,4 +1,4 @@
-import React, { useMemo, useCallback, useContext } from 'react'
+import React, { useMemo, useCallback } from 'react'
 import PropTypes from 'prop-types'
 import { compareDesc, format } from 'date-fns'
 import {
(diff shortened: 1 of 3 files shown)


apps/finance/app/src/components/Transfers.js
```diff
@@ -109,24 +62,13 @@ const Transfers = React.memo(({ tokens, transactions }) => {
...
-      currentApp.appAddress,
-      selectedDateRange
-    )
-    saveAs(new Blob([data], { type: 'text/csv;charset=utf-8' }), filename)
-    toast('Transfers data exported')
-  }, [currentApp, filteredTransfers, tokenDetails, resolveAddress])
+  const { handleDownload } = useDownloadData({
+    filteredTransfers,
+    selectedDateRange,
+    toast,
```
sohkai, line 69: Rather than providing this, we should directly use `useToast()` in the `useDownload()` hook.

apps/finance/app/src/lib/date-utils.js
```diff
@@ -0,0 +1,8 @@
+import { format } from 'date-fns'
+
+export const MMDDYY_FUNC_FORMAT = 'MM/dd/yy'
+export const YYMMDD_LONG_FORMAT = 'yyyy-MM-dd'
+
+export function formatDate(date) {
```
sohkai, line 6: Not sure if there are other areas where we use `format()` directly, but we could settle on always importing `formatDate()` from this `date-utils` when we need to format dates.

Then we can add the format we'd like as a parameter, and default it to `MMDDYY_FUNC_FORMAT`.

apps/finance/app/src/components/useDownloadData.js
```diff
@@ -0,0 +1,94 @@
+import { useContext, useCallback } from 'react'
+import { format } from 'date-fns'
+import { saveAs } from 'file-saver'
+import { useCurrentApp } from '@aragon/api-react'
+import { IdentityContext } from './IdentityManager/IdentityManager'
+import { toChecksumAddress } from '../lib/web3-utils'
```
sohkai, line 6: Nitpick: let's alphabetize :).

apps/finance/app/src/components/useDownloadData.js
```diff
@@ -0,0 +1,94 @@
...
+  const handleDownload = useCallback(async () => {
+    const { appAddress: financeAddress } = currentApp
+    const downloadData = await getDownloadData(
+      filteredTransfers,
+      tokenDetails,
+      resolveAddress
+    )
+    const fileName = getDownloadFilename(financeAddress, selectedDateRange)
+
+    saveAs(
```
sohkai, line 84: Is this async? Should we await on it, so we only show the toast if it completes?

apps/finance/app/src/components/useDownloadData.js
```diff
@@ -0,0 +1,94 @@
...
+      new Blob([downloadData], { type: 'text/csv;charset=utf-8' }),
+      fileName
+    )
+    toast('Transfers data exported')
```
sohkai, line 88:
```suggestion
    toast('Transaction data exported')
```

We should also do this in Agent!

apps/finance/app/src/components/useDownloadData.js
```diff
@@ -0,0 +1,94 @@
...
+  ])
+}
+
+function useDownloadData({
+  filteredTransfers,
+  selectedDateRange,
+  toast,
+  tokenDetails,
+}) {
+  const { resolve: resolveAddress } = useContext(IdentityContext)
```
sohkai, line 72: Let's do this separately, but what do you think about migrating all the apps' `IdentityManager`s to [Voting's version](https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/identity-manager.js) which includes the utilities for resolving?

apps/finance/app/src/components/useDownloadData.js
Evalir: Nope, `getDownloadFilename` is not async; if this is called the CSV creation itself will have already completed as `getDownloadData` (called above) is async.

apps/finance/app/src/components/useDownloadData.js
Evalir: Yes! Will open the PR.

apps/finance/app/src/components/useDownloadData.js
Evalir: I agree, this version feels fairly cumbersome and not so easy to understand at first glance. Will open the issue so we can get to it right away!

apps/finance/app/src/components/useDownloadData.js
Evalir: You mean alphabetize by import names? 😃 AFAIK by the rules we have, this is alphabetized:
- npm packages first and alphabetized
- namespaced packages after npm packages, also alphabetized
- local files, alphabetized

sohkai: The lib imports are currently upside down :)

Evalir: Aaah gotcha!

apps/finance/app/src/components/useDownloadData.js
Evalir: This, along with the other notes are now [here](https://github.com/aragon/aragon-apps/issues/1120).

apps/finance/app/src/components/useDownloadData.js
```diff
@@ -0,0 +1,95 @@
...
+        const formattedAmount = formatTokenAmount(
+          amount,
+          isIncoming,
+          decimals,
+          true,
+          { rounding: 5 }
+        )
+        const { name = '' } = (await resolveAddress(entity)) || {}
+        return [
+          formatDate(date),
```
sohkai, line 51: This is currently using `MM/dd/yy`, but perhaps we should default the date formatting to `dd/mm/yy` instead? That's what it used to be: https://github.com/aragon/aragon-apps/pull/1103/files#diff-c6b357d815fc23c7aea312476aabe3beL33.

Or building on the offline discussion, maybe we should just settle on yyyy-mm-dd as the default. It's kind of weird we use that format for the file name but then not the dates inside.

dizzypaty: Yes, we agreed yesterday to implement the ISO international format `yyyy-mm-dd` by default everywhere.

Evalir: As a result of this, we need to update the tables and DataView; This will be done in a separate PR.

apps/finance/app/src/lib/date-utils.js
```diff
@@ -3,6 +3,6 @@ import { format } from 'date-fns'
 export const MMDDYY_FUNC_FORMAT = 'MM/dd/yy'
```
sohkai, line 3: Wondering if we can remove this now, and avoid exporting the other format?

Evalir: Yep, we can actually do this after standardizing to `yyyy-MM-dd`.

PR #1104, opened 2020-03-19
Finance: format token helper script section to match Agent's

### LINKED ISSUE #1091: Finance: update script.js to include "token helpers" section
[As mentioned here,](https://github.com/aragon/aragon-apps/pull/1082#discussion_r389054181) we should update the Finance app `script.js` to match the formatting of the Agent's `script.js` token helpers section.

Evalir:
Closes #1091 .

### DIFF
--- a/apps/finance/app/src/script.js
+++ b/apps/finance/app/src/script.js
@@ -294,9 +294,72 @@ async function newTransaction(
 
 /***********************
  *                     *
- *       Helpers       *
+ *    Token Helpers    *
  *                     *
  ***********************/
+
+function loadTokenBalance(tokenContract, tokenAddress, { ethToken, vault }) {
+  if (addressesEqual(tokenAddress, ethToken.address)) {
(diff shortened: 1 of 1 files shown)


apps/finance/app/src/script.js
```diff
@@ -294,10 +294,73 @@ async function newTransaction(
...
+    symbol = override || (await getTokenSymbol(app, tokenAddress))
+    tokenSymbols.set(tokenContract, symbol)
+  } catch (err) {
+    // symbol is optional
+    symbol = ''
+  }
+  return symbol
+}
+
 async function updateBalances(
```
sohkai, line 364: Could we add another "section header" starting from here for the balance / transaction helpers?

Evalir: Sounds good!

Evalir: Let's do it!

PR #1109, opened 2020-04-03
Finance: Move to buidler setup

0xGabi:
Note: Both coverage and gas report still needs to use ganache-cli until buidlerevm support it (soonish). To keep compatible with buidler the ganache-cli bash script of test-helpers need to change to [this one instead.](https://github.com/aragon/contract-helpers/pull/30/files#diff-38a72668a71945805c505ddf2efd7b26)

As a consequence of the above, this PR depends on https://github.com/aragon/contract-helpers/pull/30


apps/finance/app/package.json
```diff
@@ -46,11 +46,15 @@
...
   "scripts": {
     "lint": "eslint ./src",
-    "start": "npm run sync-assets && npm run watch:script & parcel serve index.html -p 3002 --out-dir build/",
-    "build": "npm run sync-assets && npm run build:script && parcel build index.html --no-cache --out-dir build/ --public-url \".\"",
-    "build:script": "parcel build src/script.js --out-dir build/",
-    "watch:script": "parcel watch src/script.js --out-dir build/ --no-hmr",
-    "sync-assets": "copy-aragon-ui-assets -n aragon-ui ./build && rsync -rtu ./public/ ./build",
+    "test": "jest",
+    "start": "npm run sync-assets && npm run watch:script & parcel serve index.html -p 3002 --out-dir ../dist/",
+    "watch": "npm run watch:script",
```
sohkai, line 51: Just wondering, do we need this for buidler?

apps/finance/scripts/buidler-hooks.js
```diff
@@ -0,0 +1,20 @@
+let vault
+
+module.exports = {
+  postDao: async function({ _experimentalAppInstaller }, bre) {
+    vault = await _experimentalAppInstaller('vault')
```
sohkai, line 5: Just wondering, are we also able to deploy off of a contract from a node module?

It would be nice to use the local Vault dependency here:

1. We want to keep Finance in sync with any local changes here
2. It's nice to avoid going to network if we don't have to

apps/finance/package.json
```diff
@@ -37,22 +35,22 @@
...
+    "@aragon/buidler-aragon": "^0.2.0-beta.2",
+    "@aragon/contract-test-helpers": "0.0.1",
+    "@nomiclabs/buidler": "^1.0.2",
+    "@nomiclabs/buidler-etherscan": "^1.2.0",
+    "@nomiclabs/buidler-solhint": "^1.2.0",
+    "@nomiclabs/buidler-truffle5": "^1.1.2",
+    "@nomiclabs/buidler-web3": "^1.1.2",
+    "bignumber.js": "^9.0.0",
+    "buidler-gas-reporter": "^0.1.3",
+    "eth-ens-namehash": "^2.0.8",
```
sohkai, line 52: Do we need this for the tests now?

apps/finance/package.json
```diff
@@ -3,22 +3,20 @@
   "version": "3.0.0",
   "main": "index.js",
   "scripts": {
-    "compile": "truffle compile",
+    "postinstall": "npm run compile && npm run copy-artifacts && npm run build-app",
```
sohkai, line 6: The reason we had the prepare script previously was to avoid having to build the app.

I wonder for the sake of sanity, we should just drop building the app altogether—if someone new wants to contribute, they can manually install the app they'd like to work on rather than every app in this repo.

If we do this, can we then remove the `build-app` script?

apps/finance/package.json
```diff
@@ -3,22 +3,20 @@
...
-    "ganache-cli:test": "./node_modules/@aragon/test-helpers/ganache-cli.sh",
-    "prepare": "npx apps-shared-scripts-prepare",
-    "install:frontend": "cd app && npm install",
-    "abi:extract": "truffle-extract --output abi/ --keys abi",
-    "prepublishOnly": "truffle compile --all && npm run abi:extract -- --no-compile"
+    "coverage": "buidler coverage --network buidlerevm",
+    "console": "buidler console"
   },
   "files": [
     "/abi",
```
sohkai, line 22: I think `/build` needs to be changed here to `/artifact`?

apps/finance/test/finance.js
```diff
@@ -1,990 +1,1608 @@
-const getBalance = require('@aragon/test-helpers/balance')(web3)
```
sohkai, line 1: This test file change is very difficult to review.

Please—it's easy to make multiple commits with git and we should take advantage of this to make reviewing these types of changes much easier than re-reading 2500 lines of code.

Can we revert the changes here, and then apply in separate commits:

1. Update formatting
2. Test changes

apps/finance/app/package.json
0xGabi: Yes, the scripts `serve`, `watch` and `sync-assets` are pre-requisites of the plugin `start` task if there is an app directory. 

I think this is a harsh restriction for the user at the moment. Maybe in a future version, we should give configurations to specify the scripts and don't make `sync-assets` a must-have.

apps/finance/app/package.json
```diff
@@ -46,11 +46,15 @@
...
+    "build": "npm run sync-assets && npm run build:app && npm run build:script",
+    "build:app": "parcel build index.html -d ../dist/ --public-url \".\" --no-cache",
+    "build:script": "parcel build src/script.js --out-dir ../dist/ --no-cache",
+    "serve": "parcel serve index.html --out-dir ../dist/ --no-cache",
```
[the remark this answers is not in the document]
0xGabi: Yes as I said above we need it. And I agree, it makes to use 3002 👍

apps/finance/package.json
0xGabi: Yes, that's a good point. I remember taking a super long time to install the whole repo.

apps/finance/package.json
0xGabi: No, good catch 👍

apps/finance/scripts/buidler-hooks.js
0xGabi: You are right. We didn't think about it. We should include it as a new feature. I'll add an issue to track it.

apps/finance/test/finance.js
0xGabi: @sohkai Apologies for that. It's ready to review now. I only included the changes to move to web3@1 now. I had a problem with one extra test than Ale. 

`sets the end of time correctly`

I'm including it to #1111

apps/finance/app/package.json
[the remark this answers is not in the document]
0xGabi: Doing further tests if we want to change to port we need to do it on `buidler.config` file otherwise the app don't load correctly.

0xGabi:
It's ready for a quick review before merging.

apps/finance/test/finance.js
```diff
@@ -176,9 +180,9 @@ contract('Finance App', ([root, owner, recipient]) => {
         await financeApp.mockIncreaseTime(100) // to make sure it overflows with MAX_UINT64 period length
         // initialize with MAX_UINT64 as period duration
         await financeApp.initialize(vault.address, MAX_UINT64)
-        const [isCurrent, start, end, firstTx, lastTx] = await financeApp.getPeriod(await financeApp.currentPeriodId())
+        const data = await financeApp.getPeriod(await financeApp.currentPeriodId())
 
-        assert.equal(end.valueOf(), MAX_UINT64.valueOf(), "should have set the period's end date to MAX_UINT64")
+        assert.equal(data.endTime.eq(MAX_UINT64), true, "should have set the period's end date to MAX_UINT64")
```
0xGabi, line 185: We use BN `eq` function to compare two BN numbers

apps/finance/buidler.config.js
```diff
@@ -0,0 +1,39 @@
...
+usePlugin('buidler-gas-reporter')
+usePlugin('solidity-coverage')
+
+module.exports = {
+  defaultNetwork: 'localhost',
+  networks: {
+    localhost: {
+      url: 'http://localhost:8545',
+      accounts: {
+        mnemonic: "explain tackle mirror kit van hammer degree position ginger unfair soup bonus"
```
0xGabi, line 15: We use the same mnemonic as in ganache-cli script

apps/finance/contracts/test/TestImports.sol
```diff
@@ -7,11 +7,10 @@ import "@aragon/os/contracts/factory/EVMScriptRegistryFactory.sol";
 
 import "@aragon/apps-vault/contracts/Vault.sol";
 
-import "@aragon/apps-shared-migrations/contracts/Migrations.sol";
```
0xGabi, line 10: We don't need migrations anymore.

apps/finance/app/package.json
sohkai: Definitely—we should also double check if they exist and warn if they don't rather than make it a hard requirement.

apps/finance/buidler.config.js
sohkai: Ah interesting, I made a note in one of the other apps, but maybe this is something we can export then.

This is just so the tests and the local development environment are using the same accounts, right? Technically this shouldn't have an impact? Or is the ganache-cli script from `contract-helpers` starting aragen?

apps/finance/test/finance.js
sohkai: Noted in Voting as well, but let's use the string serialization instead of `.eq()` to make the failure case more explicit in output logs :)

apps/finance/app/package.json
```diff
@@ -46,11 +46,13 @@
...
-    "build:script": "parcel build src/script.js --out-dir build/",
-    "watch:script": "parcel watch src/script.js --out-dir build/ --no-hmr",
-    "sync-assets": "copy-aragon-ui-assets -n aragon-ui ./build && rsync -rtu ./public/ ./build",
+    "start": "npm run sync-assets && npm run watch & parcel serve index.html -p 3002 --out-dir ../dist/",
+    "watch": "parcel watch src/script.js --out-dir ../dist/ --no-hmr",
+    "serve": "parcel serve index.html --out-dir ../dist/ --no-cache",
+    "build": "npm run sync-assets && npm run build:frontend && npm run build:script",
+    "build:frontend": "parcel build index.html -d ../dist/ --public-url \".\" --no-cache",
+    "build:script": "parcel build src/script.js --out-dir ../dist/ --no-cache",
+    "sync-assets": "copy-aragon-ui-assets -n aragon-ui ../dist && rsync -rtu ./public/ ../dist",
```
sohkai, line 55: Similar comments from #1119 apply here :)

apps/finance/app/package.json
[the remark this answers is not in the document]
sohkai: Would like to understand how this works a bit better. Ideally `npm run start` would be similar to how the buidler plugin starts the app.

If I understand correctly, `serve`, `watch`, and `sync-assets` are requirements for buidler, but `start` is not. `start` is only used if we'd like to run the frontend "headless" by itself? And the port settings for serving the frontend with buidler **must** be configured in the buidler config?

If these are true, I'm starting to wonder if we should make the frontend's `start` script provide a help text of sorts, to better guide the user on which to run in case they ran it by mistake.

apps/finance/scripts/buidler-hooks.js
sohkai: Let's also add a TODO note here to remind us to improve it :).

apps/finance/scripts/buidler-hooks.js
```diff
@@ -0,0 +1,20 @@
+let vault
+
+module.exports = {
+  postDao: async function({ _experimentalAppInstaller, log }, bre) {
+    vault = await _experimentalAppInstaller('vault')
+    log(`> Vault app installed: ${vault.address}`)
+  },
+
+  preInit: async function({ proxy, log }, bre) {
+    await vault.createPermission('TRANSFER_ROLE', proxy.address)
```
sohkai, line 10: Noted in the Agent app as well, but I think we need to hash this string.

apps/finance/test/finance.js
```diff
@@ -636,28 +643,28 @@ contract('Finance App', ([root, owner, recipient]) => {
...
+                const data = await finance.getPeriod(2)
 
-                assert.isFalse(isCurrent, 'shouldnt be current period')
-                assert.equal(start.valueOf(), PERIOD_DURATION * 2 + 1, 'should have correct start date')
-                assert.equal(end.valueOf(), PERIOD_DURATION * 3, 'should have correct end date')
-                assert.equal(firstTx, 0, 'should have empty txs')
-                assert.equal(lastTx, 0, 'should have empty txs')
+                assert.isFalse(data.isCurrent, 'shouldnt be current period')
+                assert.equal(data.startTime.toNumber(), PERIOD_DURATION * 2 + 1, 'should have correct start date')
+                assert.equal(data.endTime.toNumber(), PERIOD_DURATION * 3, 'should have correct end date')
```
sohkai, line 665: Just in case, I wonder if we can use `.toString()` here instead. We should in general prefer to serialize BNs as strings; JS will transparently handle the cast from number to string for the other values.

apps/finance/test/finance.js
```diff
@@ -636,28 +643,28 @@ contract('Finance App', ([root, owner, recipient]) => {
...
 
                 const receipt = await finance.send(sentWei, { gas: 3e5 })
                 const transactionId = getEventArgument(receipt, 'NewTransaction', 'transactionId')
-                const [periodId, amount, paymentId, paymentExecutionNumber, token, entity, incoming, date] = await finance.getTransaction(transactionId)
+                const data = await finance.getTransaction(transactionId)
 
-                assert.equal(amount, sentWei, 'app should have received ETH and sent it to vault')
-                assert.equal((await getBalance(vault.address)).valueOf(), prevVaultBalance + sentWei, 'app should have received ETH and sent it to vault')
+                assert.equal(data.amount, sentWei, 'app should have received ETH and sent it to vault')
+                assert.equal(parseInt(await getBalance(vault.address), 10), prevVaultBalance + sentWei, 'app should have received ETH and sent it to vault')
```
sohkai, line 655: Mentioned elsewhere, but let's avoid `parseInt()` in tests when we get BNs back.

apps/finance/test/finance.js
```diff
@@ -32,6 +32,9 @@ const tokenTestGroups = [
     },
 ]
 
+// Hack to make contract-test-helpers work with Web3 1.x.
+web3.BigNumber = web3.utils.BN
```
sohkai, line 36: Interesting... where do we need this in contract-helpers?

If we do intend to keep this, let's add a big TODO that we can fix immediately.

apps/finance/app/package.json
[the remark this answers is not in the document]
0xGabi: Yes, all those you mentioned are true. You are right about the message, maybe also using another name instead of start?

apps/finance/buidler.config.js
0xGabi: No, `contract-helpers` was not starting aragen. I was having this behavior: https://github.com/trufflesuite/truffle/issues/1062

Actually doing further tests it only happens if `ganache-cli` doesn't initialize a mnemonic. So the account configuration on `buidler.config` is not necessary. Regardless I wonder if it's a restriction on the way that buidler setup the JSON-RPC networks, I'll ask them 🤔

PR #1110, opened 2020-04-03
Voting: add support / approval copy suggestion

Evalir:
Spring cleaning 🌸, closing [#1346](https://github.com/aragon/aragon/issues/1346) over at `aragon/aragon`.

> Your affinity for fixes is apparent. I'm confident you'll find a typo there as well.

Adds suggestion for the voting copy regarding approval & support by @john-light . Maybe we want a bit more discussion around it?

### DIFF
--- a/apps/voting/app/src/screens/VoteDetail.js
+++ b/apps/voting/app/src/screens/VoteDetail.js
@@ -216,7 +216,7 @@ function VoteDetail({ vote, onBack, onVote, onExecute }) {
                     color: ${theme.surfaceContentSecondary};
                   `}
                 >
-                  ({round(supportRequired * 100, 2)}% support needed)
+                  (>{round(supportRequired * 100, 2)}% support needed)
                 </span>
               </div>
               <SummaryBar
@@ -239,7 +239,7 @@ function VoteDetail({ vote, onBack, onVote, onExecute }) {
                     color: ${theme.surfaceContentSecondary};
                   `}
                 >
-                  ({round(minAcceptQuorum * 100, 2)}% approval needed)
+                  (>{round(minAcceptQuorum * 100, 2)}% approval needed)
                 </span>
               </div>
               <SummaryBar


Evalir:
Ah yes, I'm actually putting up screenshots for these changes; may have missed posting these. I'll edit the original PR message to have them.

sohkai:
Right... so in this case, we can see that the title is now overflowing into two lines and is most likely not what we'd like:

Screen Shot 2020-04-07 at 9 01 22 PM

To get that character back... maybe we could rephrase to `(needs >50% support)`?

Evalir:
Seems that this would work!
Screen Shot 2020-04-07 at 3 08 13 PM

I feel the copy is clear enough, however, if we want to keep the `>` we don't have much room for rephrasing.

sohkai:
Another option is to drop the "support" or "quorum" from the copy, given that it's already in context due to the Box's title.

Evalir:
This makes sense. Something like `99.62%` (needs >50%)

I'd say we only need to wait for @dizzypaty 's thoughts and we can do a quick PR.

john-light:
@Evalir while we're modifying this section's copy, could we change `Relative Support %` in the box's title to `Support %`? I'm not sure what "Relative" is supposed to mean here, and anyways AFAIK we don't call it "Relative Support" anywhere else (neither in the Permissions for this parameter nor the setup flow when first setting this parameter).

Evalir:
I believe we can @john-light ; we indeed are not using the "relative term" anywhere else, even in the internal contract documentation.

sohkai:
We use "relative percentage" when [describing support during onboarding](https://github.com/aragon/aragon/blob/master/src/templates/kit/screens/VotingScreen.js#L173-L176):

> Support is the relative percentage of tokens that are required to vote “Yes” for a proposal to be approved

This was meant to contrast the _absolute_ nature of calculating quorum (support is relative to current voters, quorum is absolute against all possible voters).

But I agree, we should be consistent with "support". Perhaps we can lift the Help tooltips directly from onboarding?

PR #1112, opened 2020-04-06
Voting: make amount of tokens always display 3 decimals

Evalir:
Spring cleaning 🌸, closing [#589](https://github.com/aragon/aragon/issues/589).

> _Heavy metal intensifies_

Makes the number of tokens shown in the vote detail always have 3 decimals; this amount was chosen due to anj.aragon.org always displaying this amount of decimals, in an effort to make it more consistent across all products (we may want some discussion over this, as other apps are using 2 decimals). The decimal change was not made at the top (script) level to avoid any precision issues elsewhere. Some screenshots:

Before:
Screen Shot 2020-04-06 at 5 15 07 PM

After:
Screen Shot 2020-04-06 at 5 14 24 PM

### DIFF
--- a/apps/voting/app/src/components/SummaryRows.js
+++ b/apps/voting/app/src/components/SummaryRows.js
@@ -18,14 +18,14 @@ function SummaryRows({ yea, nay, symbol, connectedAccountVote }) {
         color={theme.positive}
         label="Yes"
         pct={yea.pct}
-        token={{ amount: yea.amount, symbol }}
+        token={{ amount: yea.amount.toFixed(3), symbol }}
         youVoted={connectedAccountVote === VOTE_YEA}
       />
       <SummaryRow
         color={theme.negative}
         label="No"
         pct={nay.pct}
-        token={{ amount: nay.amount, symbol }}
+        token={{ amount: nay.amount.toFixed(3), symbol }}
         youVoted={connectedAccountVote === VOTE_NAY}
       />
     </div>


bpierre:
Since the number of decimals is arbitrary, I think it might be a good idea to use 2 or 3 as a base, but also detect cases where we need to display more.

In the context of a vote, I think the only important thing is to ensure that we don’t have two values that appear equal while they are not.

A way to do this would be, after checking that they are indeed different, to keep adding decimals until the two values are different. Then we could use the same amount of decimals for the two.

Also I wonder if we should also display somewhere (e.g. in the label with `<Help />`) that the values are rounded? /cc @dizzypaty @sohkai

Evalir:
> A way to do this would be, after checking that they are indeed different, to keep adding decimals until the two values are different. Then we could use the same amount of decimals for the two.

What does being "different" mean? How would this work?

> Also I wonder if we should also display somewhere (e.g. in the label with <Help />) that the values are rounded?

I definitely think this should be the case, for having better transparency.

bpierre:
> What does being "different" mean? How would this work?

What I meant is that we could start by testing the equality of the two values before detecting the number of decimals needed:

```jsx
function displayDecimals(valueA, valueB) {

  if (valueA === valueB) {
    return 2 // we can stop here if the values are identical
  }

  // detect and return the required number of display decimals
}
```

sohkai:
> The decimal change was not made at the top (script) level to avoid any precision issues elsewhere

This is the correct choice, but not for the reason stated—unless there is good reason not to, the script should be fully precise let the responsibility of "formatting" go to the frontend.

> In the context of a vote, I think the only important thing is to ensure that we don’t have two values that appear equal while they are not.

Yes! Agree strongly.

sohkai:
@Evalir Should we close this for https://github.com/aragon/aragon-apps/pull/1113, since `token-amount` has the ability to do this natively?

Evalir:
Yup, makes sense as we've got the utility now. 👍

PR #1113, opened 2020-04-08
Voting: Wrong token amount used to vote 

rperez89:
NEED TO TEST IT!!

I don't know how to run an aragon app as an standalone app to test it.

This pr closes https://github.com/aragon/aragon/issues/1347

### DIFF
--- a/apps/voting/app/src/hooks/useExtendedVoteData.js
+++ b/apps/voting/app/src/hooks/useExtendedVoteData.js
@@ -1,6 +1,6 @@
 import { useMemo } from 'react'
 import { useAragonApi } from '@aragon/api-react'
-import { getUserBalanceAt, getUserBalanceNow } from '../token-utils'
+import { getUserBalanceAt, getUserBalanceNow, formatBalance } from '../token-utils'
 import { getCanExecute, getCanVote } from '../vote-utils'
 import useTokenContract from './useTokenContract'
 import usePromise from './usePromise'
@@ -28,12 +28,14 @@ export default function useExtendedVoteData(vote) {
     if (!vote) {
       return -1
     }
-    return getUserBalanceAt(
+    const userBalance =  getUserBalanceAt(
       connectedAccount,
       vote.data.snapshotBlock,
       tokenContract,
-      tokenDecimals
     )
+    
+    return formatBalance(userBalance, tokenDecimals)
+
   }, [connectedAccount, tokenContract, tokenDecimals, vote])
   const userBalance = usePromise(userBalancePromise, [], -1)
 
--- a/apps/voting/app/src/token-utils.js
+++ b/apps/voting/app/src/token-utils.js
@@ -1,8 +1,7 @@
 export async function getUserBalanceAt(
   connectedAccount,
   snapshotBlock,
-  tokenContract,
-  tokenDecimals
+  tokenContract
 ) {
   if (!tokenContract || !connectedAccount) {
     return -1
@@ -12,7 +11,7 @@ export async function getUserBalanceAt(
     .balanceOfAt(connectedAccount, snapshotBlock)
     .toPromise()
 
-  return Math.floor(parseInt(balance, 10) / Math.pow(10, tokenDecimals))
+  return balance
 }
 
 export async function getUserBalanceNow(
@@ -27,4 +26,27 @@ export async function getUserBalanceNow(
   const balance = await tokenContract.balanceOf(connectedAccount).toPromise()
 
   return Math.floor(parseInt(balance, 10) / Math.pow(10, tokenDecimals))
+}
+
+/**
+ * Format the balance to a fixed number of decimals
+ *
+ * @param {BN} amount the total amount
+ * @param {BN} base the decimals base
+ * @param {number} precision number of decimals to format
+ * @return {string} formatted balance
+ */
+export function formatBalance(amount, base, precision = 2) {
+  const baseLength = base.toString().length
+
+  const whole = amount.div(base).toString()
+  let fraction = amount.mod(base).toString()
+  const zeros = '0'.repeat(Math.max(0, baseLength - fraction.length - 1))
+  fraction = `${zeros}${fraction}`.replace(/0+$/, '').slice(0, precision)
+
+  if (fraction === '' || parseInt(fraction, 10) === 0) {
+    return whole
+  }
+
+  return `${whole}.${fraction}`
 }


Evalir:
Hey Rodri! We have this [frontend setup](https://github.com/aragon/aragon/blob/master/docs/FRONTEND_SETUP.md#developing-the-apps) guide to get you started on developing / testing aragon apps. :) In a nutshell, running `npm start` on the voting app & `ARAGON_APP_LOCATOR=local npm run start:rinkeby` should do. ;) note that mainnet also works seamlessly!

apps/voting/app/src/token-utils.js
```diff
@@ -28,3 +27,26 @@ export async function getUserBalanceNow(
 
   return Math.floor(parseInt(balance, 10) / Math.pow(10, tokenDecimals))
```
rperez89, line 28: ahhh we have another function using the same logic than before here

rperez89:
> Actually, I wonder if we should just pause this until we get to use this from aragonUI: [aragon/aragon-ui#728](https://github.com/aragon/aragon-ui/pull/728)

Not sure, i mean is only one function haha

sohkai:
Haha yeah, but better to use the same utilities everywhere :).

bpierre:
It is available now! :point_right: https://github.com/aragon/aragon-ui/blob/bc35fec2f8165c370ff8c2760f7a0f1073b5db71/src/utils/format.js#L53-L67

bpierre:
I added some changes to remove the `Number`-based calculations from the app, and rely on `formatTokenAmount()` to format token amounts.

Changes:

- Add a new `TokenAmount` class.
- Add a new `Percentage` class.
- Add `bnPercentageToNumber()`.
- Add `formatBnPercentage()`.
- Format amounts / percentages from the `TokenAmount` / `Percentage` classes.
- Convert the calculations to BN.js.
- Remove `numData` from the vote data structure.

apps/voting/app/src/components/SummaryRows.js
```diff
@@ -1,17 +1,16 @@
 import React from 'react'
 import styled from 'styled-components'
-import { GU, textStyle, useTheme } from '@aragon/ui'
+import { GU, textStyle, formatTokenAmount, useTheme } from '@aragon/ui'
```
Evalir, line 3: We can move this behind `textStyle so it's alphabetized` 🙏

apps/voting/app/src/app-state-reducer.js
```diff
@@ -1,5 +1,6 @@
 import BN from 'bn.js'
 import { hasLoadedVoteSettings } from './vote-settings'
+import { Percentage, TokenAmount } from './math-utils'
```
Evalir, line 3: Hmm, we don't have to do it _now_, but I think we could use our newly released [TokenAmount](https://github.com/aragon/token-amount) lib!

apps/voting/app/src/hooks/useExtendedVoteData.js
```diff
@@ -1,4 +1,6 @@
 import { useMemo } from 'react'
+import BN from 'bn.js'
+import { formatTokenAmount } from '@aragon/ui'
 import { useAragonApi } from '@aragon/api-react'
```
Evalir, line 4: Now that I'm seeing the order, shouldn't we move `@aragon/ui` below `@aragon/api-react`?

apps/voting/app/src/vote-utils.js
```diff
@@ -43,22 +55,31 @@ export function getVoteStatus(vote, pctBase) {
 }
 
 export function getVoteSuccess(vote, pctBase) {
-  const { yea, minAcceptQuorum, nay, supportRequired, votingPower } = vote.data
+  const { minAcceptQuorum, nay, supportRequired, votingPower, yea } = vote.data
 
-  const totalVotes = yea.add(nay)
+  const totalVotes = yea.value().add(nay.value())
```
Evalir, line 60: Just realized this probably could equal `-2` if both `yea` and `nay` get passed in as `-1`, but, AFAIK, this can't really happen as they're not being set at `-1` at any moment.

apps/voting/app/src/app-state-reducer.js
```diff
@@ -44,20 +40,12 @@ function appStateReducer(state) {
...
-              minAcceptQuorum: parseInt(data.minAcceptQuorum, 10) / pctBaseNum,
-              nay: parseInt(data.nay, 10) / tokenDecimalsBaseNum,
-              supportRequired: parseInt(data.supportRequired, 10) / pctBaseNum,
-              votingPower:
-                parseInt(data.votingPower, 10) / tokenDecimalsBaseNum,
-              yea: parseInt(data.yea, 10) / tokenDecimalsBaseNum,
+              supportRequired: new Percentage(data.supportRequired, pctBaseBn),
+              votingPower: new TokenAmount(data.votingPower, tokenDecimals),
+              yea: new TokenAmount(data.yea, tokenDecimals),
+              nay: new TokenAmount(data.nay, tokenDecimals),
```
sohkai, line 48: What do you think about also including the token symbol in the `TokenAmount`?

apps/voting/app/src/components/SummaryRows.js
```diff
@@ -1,17 +1,16 @@
...
 import You from './You'
-import { formatNumber } from '../math-utils'
 import { VOTE_NAY, VOTE_YEA } from '../vote-types'
 
 function SummaryRows({ yea, nay, symbol, connectedAccountVote }) {
```
sohkai, line 7: Wanted to double check—what type is `yea` and `nay`? We're accessing `yea.pct` but I'm not sure this exists on the object anymore.

apps/voting/app/src/components/VoteCard/VoteOptions.js
```diff
@@ -1,11 +1,28 @@
...
 const ANIM_DELAY_MIN = 100
 const ANIM_DELAY_MAX = 800
 
+function optionPercentage(optionPower, totalVotingPower) {
+  const precision = 10 ** 9
+  const precisionBn = new BN(precision)
+
+  return (
+    parseInt(
+      divideRoundBigInt(
```
sohkai, line 17: It's starting to look like we may want another utility :).

Maybe `divideBN()` and we can give it a `decimalsPrecision` option?

apps/voting/app/src/math-utils.js
```diff
@@ -38,10 +39,88 @@ export function formatNumber(num, decimals = 2, { truncate = true } = {}) {
...
+export function formatBnPercentage(
+  value,
+  pctBase,
+  { digits = 2, suffix = '%' } = {}
+) {
+  let basePrecision = 10 ** digits
+
+  // Tolerate having too many digits by correcting the value.
+  if (basePrecision > Number.MAX_SAFE_INTEGER) {
+    digits = Math.floor(Math.log(Number.MAX_SAFE_INTEGER) / Math.log(10))
```
sohkai, line 57: Not sure about this:

1. I think we should make this max precision value be a constant
2. This value is actually only `15`, and and I think we should at least support 18 decimals (since that's what tokens use, as well as our contract's `pctBase` exponent).

It's unlikely someone will want something this large, but I'm also wondering if we shouldn't just cap the precision at the `pctBase`'s exponent (since it'll just be 0s after that anyway).

Evalir: > I think we should make this max precision value be a constant

Why? This is already on the Javascript `Number` object, and so having this as a constant seems a bit redundant to me.

> This value is actually only 15, and and I think we should at least support 18 decimals (since that's what tokens use, as well as our contract's pctBase exponent).

This is true, and relates to #1177 in terms of having an utility to safely scale up normal numbers to Bigints.

sohkai: > Why? This is already on the Javascript Number object, and so having this as a constant seems a bit redundant to me.

Because `Math.floor(Math.log(Number.MAX_SAFE_INTEGER) / Math.log(10))` is pure and doesn't need to be calculated every time

eliobricenov:
hey, I continued the last changes you requested here https://github.com/eliobricenov/aragon-apps/commits/voting-token-amount-fix-v2 

Should I create a new PR with these changes?

C.C @nivida

PR #1114, opened 2020-04-08
Voting: use local date in warnings about tokens holded

rperez89:
This pr closes https://github.com/aragon/aragon/issues/1150.

Rather than aligning the dates to UTC, we've decided to keep the times local, as all the other filters (as well as apps, e.g. Finance) use local times.

Screen Shot 2020-04-09 at 7 19 34 PM

### DIFF
--- a/apps/voting/app/src/components/VoteActions.js
+++ b/apps/voting/app/src/components/VoteActions.js
@@ -1,17 +1,19 @@
 import React, { useEffect, useState, useCallback } from 'react'
 import styled from 'styled-components'
 import {
+  blockExplorerUrl,
   Button,
   GU,
   IconCheck,
   IconConnect,
   IconCross,
   Info,
+  Link,
   RADIUS,
   textStyle,
   useTheme,
 } from '@aragon/ui'
-import { useAppState, useConnectedAccount } from '@aragon/api-react'
+import { useAppState, useConnectedAccount, useNetwork } from '@aragon/api-react'
 import useExtendedVoteData from '../hooks/useExtendedVoteData'
 import { noop, formatDate } from '../utils'
 import { VOTE_NAY, VOTE_YEA } from '../vote-types'
@@ -223,7 +225,8 @@ const TokenReference = ({
     <strong>
       {userBalance} {tokenSymbol}
     </strong>{' '}
-    due to the snapshot taken at block <strong>{snapshotBlock}</strong> at{' '}
+    due to the snapshot taken at block{' '}
+    <BlockNumber blockNumber={snapshotBlock} /> at{' '}
     <strong>{formatDate(startDate)}</strong>.{' '}
     {userBalance !== userBalanceNow ? (
       <span>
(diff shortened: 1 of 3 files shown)


apps/voting/app/src/hooks/useVotes.js
```diff
@@ -12,7 +12,8 @@ function useDecoratedVotes() {
   const installedApps = useInstalledApps()
 
   return useMemo(() => {
-    if (!votes) {
+    // Make sure we have loaded information about the current app and other installed apps before showing votes
+    if (!(votes && currentApp && installedApps)) {
```
bpierre, line 16: Should this be part of this PR?

apps/voting/app/src/components/VoteActions.js
```diff
@@ -239,6 +242,23 @@ const TokenReference = ({
...
 
+function BlockNumber({ blockNumber }) {
+  const network = useNetwork()
+
+  return network ? (
+    <Link
+      href={blockExplorerUrl('block', blockNumber, {
+        networkType: network.type,
+      })}
+      title={blockNumber}
```
bpierre, line 253: We could remove this since the number is not truncated, or maybe replace it by something like this?

```jsx
title={`See block ${blockNumber} on Etherscan`}
```

sohkai:
@rperez89 Let's also add some screenshots!

apps/voting/app/src/hooks/useVotes.js
rperez89: ahh no, i needed that in order to test it, we need to get merged https://github.com/aragon/aragon-apps/pull/1116

apps/voting/app/src/components/VoteActions.js
rperez89: i removed it :)

rperez89:
> @rperez89 Let's also add some screenshots!

done!

apps/voting/app/src/components/VoteActions.js
```diff
@@ -223,7 +225,8 @@ const TokenReference = ({
     <strong>
       {userBalance} {tokenSymbol}
     </strong>{' '}
-    due to the snapshot taken at block <strong>{snapshotBlock}</strong> at{' '}
+    due to the snapshot taken at block{' '}
```
sohkai, line 228:
```suggestion
    as that was your balance during the snapshot taken at block{' '}
```

apps/voting/app/src/components/VoteActions.js
```diff
@@ -223,7 +225,8 @@ const TokenReference = ({
...
+    <BlockNumber blockNumber={snapshotBlock} /> at{' '}
```
sohkai, line 229:
```suggestion
    <BlockNumber blockNumber={snapshotBlock} /> (mined at <date>).
```

apps/voting/app/src/components/VoteActions.js
```diff
@@ -241,7 +241,7 @@ const TokenReference = ({
     )}
   </Info>
 )
-
+// Voting with <tokens>. This was your balance when the vote started (block <block>, mined at <date>).
```
sohkai, line 244: Haha, should we remove this?

PR #1116, opened 2020-04-09
Voting: fix race condition with currentApp address being null

rperez89:
We were getting an error in the `LocalIdentityProvider` inside the wrapper because we had a race condition where the `currentApp` or `installedApps` were not resolved by the time we were rendering initial votes (and sending a null address into the local identity badges in)

### DIFF
--- a/apps/voting/app/src/hooks/useVotes.js
+++ b/apps/voting/app/src/hooks/useVotes.js
@@ -12,6 +12,7 @@ function useDecoratedVotes() {
   const installedApps = useInstalledApps()
 
   return useMemo(() => {
+    // Make sure we have loaded information about the current app and other installed apps before showing votes
     if (!(votes && currentApp && installedApps)) {
       return [[], []]
     }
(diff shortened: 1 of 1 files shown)


apps/voting/app/src/hooks/useVotes.js
```diff
@@ -68,6 +68,7 @@ function useDecoratedVotes() {
     })
 
     // Reduce the list of installed apps to just those that have been targetted by apps
+    console.log('installed apps ', installedApps)
```
Evalir, line 71: Let's remove this. 😄

apps/voting/app/src/hooks/useVotes.js
```diff
@@ -12,7 +12,7 @@ function useDecoratedVotes() {
   const installedApps = useInstalledApps()
 
   return useMemo(() => {
-    if (!votes) {
+    if (!(votes && currentApp && installedApps)) {
```
sohkai, line 15: Let's add a comment here:

> // Make sure we have loaded information about the current app and other installed apps before showing votes

apps/voting/app/src/hooks/useVotes.js
rperez89: oops

PR #1117, opened 2020-04-09
Voting: quorum and support hint copy updates

john-light:
- Users are more likely to set `50%` support for simply majority rule, rather than the previous `51%`, so I'm changing this to avoid confusion among users about how to implement a simple majority rule.
- "token holders" -> "tokenholders" (which is the terminology we use in the user guide at help.aragon.org)
- Fix Minimum Approval description


sohkai:
> "token holders" -> "tokenholders" (which is the terminology we use in the user guide at help.aragon.org)

There are a number of places where we use "token holder" in Tokens as well—will fix.

PR #1119, opened 2020-04-09
Tokens: Move to buidler setup

0xGabi:
All tests kept working 🙌


apps/token-manager/app/package.json
```diff
@@ -41,13 +41,14 @@
     "prettier": "^1.8.2"
   },
   "scripts": {
-    "test": "jest",
     "lint": "eslint ./src",
-    "start": "npm run sync-assets && npm run watch:script & parcel serve index.html -p 3003 --out-dir build/",
-    "build": "npm run sync-assets && npm run build:script && parcel build index.html --out-dir build/ --public-url \".\"",
-    "build:script": "parcel build src/script.js --out-dir build/",
-    "watch:script": "parcel watch src/script.js --out-dir build/ --no-hmr",
```
0xGabi, line 49: I decided to remove this script and only leave `watch`.

apps/token-manager/package.json
```diff
@@ -37,22 +35,23 @@
...
+    "@aragon/contract-test-helpers": "0.0.1",
+    "@aragon/minime": "1.0.0",
+    "@nomiclabs/buidler": "^1.0.2",
+    "@nomiclabs/buidler-etherscan": "^1.2.0",
+    "@nomiclabs/buidler-solhint": "^1.2.0",
+    "@nomiclabs/buidler-truffle5": "^1.1.2",
+    "@nomiclabs/buidler-web3": "^1.1.2",
+    "bignumber.js": "^9.0.0",
+    "buidler-gas-reporter": "^0.1.3",
+    "eth-ens-namehash": "^2.0.8",
```
0xGabi, line 53: I'll move this one to contracts-helpers.

apps/token-manager/test/tokenmanager.js
```diff
@@ -14,10 +13,13 @@ const DAOFactory = artifacts.require('DAOFactory')
 const EVMScriptRegistryFactory = artifacts.require('EVMScriptRegistryFactory')
 const EtherTokenConstantMock = artifacts.require('EtherTokenConstantMock')
 
-const n = '0x00'
+const ZERO_ADDR = '0x0000000000000000000000000000000000000000'
```
0xGabi, line 16: I thought these are good candidates to go into contracts-helper.

apps/token-manager/scripts/buidler-hooks.js
```diff
@@ -0,0 +1,44 @@
...
+
+const ZERO_ADDRESS = '0x0000000000000000000000000000000000000000'
+
+module.exports = {
+  postDao: async function({ _experimentalAppInstaller, log }, bre) {
+    const bigExp = (x, y) =>
+      bre.web3.utils
+        .toBN(x)
+        .mul(bre.web3.utils.toBN(10).pow(web3.utils.toBN(y)))
+    const pct16 = (x) => bigExp(x, 16)
```
0xGabi, line 11: Good candidate to move into contracts-helper.

apps/token-manager/app/package.json
```diff
@@ -41,13 +41,14 @@
     "prettier": "^1.8.2"
   },
   "scripts": {
-    "test": "jest",
     "lint": "eslint ./src",
```
sohkai, line 44: Similar to the app's package.json, let's move the `lint` script to be just before `build`

apps/token-manager/app/package.json
sohkai: Sounds good!

apps/token-manager/app/package.json
```diff
@@ -41,13 +41,14 @@
...
-    "start": "npm run sync-assets && npm run watch:script & parcel serve index.html -p 3003 --out-dir build/",
-    "build": "npm run sync-assets && npm run build:script && parcel build index.html --out-dir build/ --public-url \".\"",
-    "build:script": "parcel build src/script.js --out-dir build/",
-    "watch:script": "parcel watch src/script.js --out-dir build/ --no-hmr",
-    "sync-assets": "copy-aragon-ui-assets -n aragon-ui ./build && rsync -rtu ./public/ ./build",
+    "start": "npm run sync-assets && npm run watch & parcel serve index.html -p 3003 --out-dir ../dist/",
```
sohkai, line 45:
```suggestion
    "start": "npm run sync-assets && npm run watch & npm run serve",
```

apps/token-manager/app/package.json
```diff
@@ -41,13 +41,14 @@
...
+    "watch": "parcel watch src/script.js --out-dir ../dist/ --no-hmr",
+    "serve": "parcel serve index.html --out-dir ../dist/ --no-cache",
```
sohkai, line 47:
```suggestion
    "serve": "parcel serve index.html --out-dir ../dist/",
```

Let's see how this goes... I think it should be fine.

apps/token-manager/app/package.json
```diff
@@ -41,13 +41,14 @@
...
+    "build": "npm run sync-assets && npm run build:frontend && npm run build:script",
+    "build:frontend": "parcel build index.html -d ../dist/ --public-url \".\" --no-cache",
```
sohkai, line 49:
```suggestion
    "build:frontend": "parcel build index.html --out-dir ../dist/ --public-url \".\" --no-cache",
```

Might as well keep this consistent :)

apps/token-manager/app/package.json
```diff
@@ -41,13 +41,14 @@
...
+    "build:script": "parcel build src/script.js --out-dir ../dist/ --no-cache",
+    "sync-assets": "copy-aragon-ui-assets -n aragon-ui ../dist && rsync -rtu ./public/ ../dist",
```
sohkai, line 51:
```suggestion
    "sync-assets": "copy-aragon-ui-assets -n aragon-ui ../dist/ && rsync -rtu ./public/ ../dist/",
```

apps/token-manager/buidler.config.js
```diff
@@ -0,0 +1,39 @@
...
+usePlugin('buidler-gas-reporter')
+usePlugin('solidity-coverage')
+
+module.exports = {
+  defaultNetwork: 'localhost',
+  networks: {
+    localhost: {
+      url: 'http://localhost:8545',
+      accounts: {
+        mnemonic: "explain tackle mirror kit van hammer degree position ginger unfair soup bonus"
```
sohkai, line 15: I wonder if this is something we would want to expose as part of the buidler plugin, e.g. as a default mnemonic?

Does it matter if we use this (e.g. to get the correct contract deployments)?

apps/token-manager/test/tokenmanager.js
```diff
@@ -14,10 +13,13 @@ const DAOFactory = artifacts.require('DAOFactory')
...
 const ANY_ADDR = '0xffffffffffffffffffffffffffffffffffffffff'
+const withdrawAddr = '0x0000000000000000000000000000000000001234'
```
sohkai, line 18:
```suggestion
const WITHDRAW_ADDR = '0x0000000000000000000000000000000000001234'
```

apps/token-manager/test/tokenmanager.js
```diff
@@ -291,13 +293,13 @@ contract('Token Manager', ([root, holder, holder2, anyone]) => {
 
             it("cannot call proxyPayment() from outside of the token's context", async () => {
                 const value = 10
-                const prevTokenManagerBalance = (await getBalance(tokenManager.address)).toNumber()
+                const prevTokenManagerBalance = parseInt(await getBalance(tokenManager.address), 10)
```
sohkai, line 296: We should prefer serializing and comparing as strings, as otherwise a large value here will be imprecise and result in flaky tests.

apps/token-manager/package.json
```diff
@@ -37,22 +35,23 @@
...
-  },
   "dependencies": {
-    "@aragon/apps-shared-minime": "1.0.0",
+    "@aragon/minime": "1.0.0",
     "@aragon/os": "4.2.0"
+  },
+  "devDependencies": {
+    "@aragon/buidler-aragon": "^0.2.2",
+    "@aragon/contract-test-helpers": "0.0.1",
+    "@aragon/minime": "1.0.0",
```
sohkai, line 45: I think we can remove this now

apps/token-manager/package.json
sohkai: Is this only used by contract-helpers internal logic?

apps/token-manager/package.json
```diff
@@ -37,22 +35,23 @@
...
+    "@nomiclabs/buidler": "^1.0.2",
+    "@nomiclabs/buidler-etherscan": "^1.2.0",
+    "@nomiclabs/buidler-solhint": "^1.2.0",
+    "@nomiclabs/buidler-truffle5": "^1.1.2",
+    "@nomiclabs/buidler-web3": "^1.1.2",
+    "buidler-gas-reporter": "^0.1.3",
+    "eth-ens-namehash": "^2.0.8",
+    "ganache-cli": "^6.9.1",
+    "solidity-coverage": "^0.7.0-beta.3",
```
sohkai, line 54: We can fix this later, but let's move up to the latest version :).

apps/token-manager/package.json
0xGabi: No, it's used in some tests on a couple of apps.

0xGabi:
Oh, you are right, what probably happened is that I push a commit to the vault that included the changes to the Token manager in a previous commit. I'll try to fix the commit history.

apps/token-manager/buidler.config.js
0xGabi: As mention above it's not necessary actually.

Issue #756, opened 2019-04-01
Finance: Use BN for amounts when exporting to CSV


2color:
## Problems
- We use [JS numbers](https://github.com/aragon/aragon-apps/blob/master/apps/finance/app/src/components/Transfers.js#L110) when exporting the CSV

## Assumptions 
- JS numbers have precision issues
- Generally we use BN.js whenever precision is important
- It's reasonable to assume we want high precision in the exported CSV 

## Suggestion solution 
Use BN.js to generate sums with no precision loss

## Open questions
What types does `formatTokenAmount` take? (would be nice to add some tests)

It looks like integers (which have yet to loose precision due to their representation) and return a formatted string (which has precision issues)?

If so, do you suggest we instantiate a BN.js object and then do the calculation, i.e. 
`new BN(amount).div(new BN(decimals))`?  

How do we normally convert BN objects to strings like in this case?

## Original source for issue
See [comment](https://github.com/aragon/aragon-apps/pull/730/files/d9d39644b1eede6e1b233b30789600d95719f127#diff-c6b357d815fc23c7aea312476aabe3be) in original PR for exporting functionality.

sohkai:
> What types does `formatTokenAmount` take? (would be nice to add some tests)

Normal JS numbers. Agreed on the tests; a lot of these math utils we should move into a shared lib in aragon-apps and add some tests 😄.

> How do we normally convert BN objects to strings like in this case?

I believe we either still have, or used to have, some BN utilities for conversions into strings with rounding. Would have to look around in some of the apps' utils to see (but in any case, see below).

I might've modified all of the previous BN calculations to use normal JS numbers, since it's not as important for them to be 100% accurate when shown in a UI (or used as part of an animation).

---------

In this case, we just need to manually apply the rounding, which we can do by applying the desired precision with `divRound()`:

```js
const amountBN = new BN(amount)
const decimalBase = new BN(10).pow(new BN(decimals)) // 10e(decimals)

// Split into integer and fractional parts
const intAmount = amountBN.div(decimalBase)
const fractionAmount = amountBN.mod(decimalBase)

// Apply rounding to fractional part
const roundedFractionAmount = fractionAmount.divRound(new BN(decimals - precision))

const amountStr = `${intAmount.toString()}.${roundedFractionAmount.toString()}`
```

PR #1121, opened 2020-04-09
Finance: update BN.js usage, use `formatTokenAmount()`, add rounded balances tooltip

rperez89:
Fixes https://github.com/aragon/aragon/issues/506, https://github.com/aragon/aragon-apps/issues/756, https://github.com/aragon/aragon-apps/issues/1065


apps/finance/app/src/components/BalanceToken.js
```diff
@@ -63,9 +69,21 @@ const BalanceToken = ({
...
+          {splitAmount(integerAmount, fractionalAmount)}
+          {amountWasRounded && (
+            <div
+              css={`
+                display: flex;
+                align-items: center;
+                margin-left: ${GU}px;
+              `}
+            >
+              <Help>{amount}</Help>
```
bpierre, line 84: We should also add a [hint](https://github.com/aragon/aragon-ui/blob/master/src/components/Help/Help.js#L67) on `Help`. Perhaps something like “This is an approximation, see the complete amount”? This is what gets used as a `title` attribute for the button. /cc @dizzypaty @sohkai

sohkai: > “This is an approximation, see the complete amount”

Agreed!

bpierre:
@rperez89 @sohkai I added a few commits, the main thing being the move to aragonUI’s `formatTokenAmount()`. Let me know what you think!

apps/finance/app/src/components/BalanceToken.js
```diff
@@ -63,23 +57,70 @@ const BalanceToken = ({
...
+BalanceToken.defaultProps = {
+  convertedAmount: -1,
+}
+
+BalanceToken.propTypes = {
+  address: PropTypes.string.isRequired,
+  amount: PropTypes.instanceOf(BN),
+  compact: PropTypes.bool.isRequired,
+  convertedAmount: PropTypes.instanceOf(BN),
+  decimals: PropTypes.instanceOf(BN),
```
sohkai, line 103: I wonder if this should also be required?

apps/finance/app/src/app-state-reducer.js
```diff
@@ -21,13 +21,6 @@ function appStateReducer(state) {
           ...balance,
           amount: new BN(balance.amount),
           decimals: new BN(balance.decimals),
```
sohkai, line 23: We can do this later, but now I'm wondering if we should prefer leaving `decimals` as a number (e.g. in the `use-token` data types).

It's usually encoded as `uint8`, so it can never be imprecise, and is _always_ a formatting tool (its only purpose is to be one). It may still be operated on, e.g. with uniswap/conversion APIs, but always just to calculate how many places a decimal should be moved left/right.

apps/finance/app/src/components/Balances.js
```diff
@@ -1,152 +1,170 @@
...
+                symbol,
+                verified,
+              }) => (
+                <li
+                  key={address}
+                  css={`
+                    flex-shrink: 0;
+                    display: block;
+                    min-width: ${20 * GU}px;
+                    padding-right: ${3 * GU}px;
```
sohkai, line 143:
```suggestion
                    padding-right: ${4 * GU}px;
```

I was looking at the A1 org, and for larger token balances, the additional 8px is nice (since we have the min-width applied as well, this doesn't really affect smaller token balances).

bpierre: Agreed, especially with the new `<Help />` :+1:

apps/finance/app/src/components/Balances.js
```diff
@@ -114,12 +125,22 @@ class Balances extends React.Component {
...
+                  symbol,
+                  verified,
+                }) => (
                   <li
                     key={address}
                     css={`
+                      flex-shrink: 0;
                       display: block;
                       min-width: ${20 * GU}px;
+                      padding-right: ${3 * GU}px;
```
sohkai, line 142:
```suggestion
```

😄

rperez89: this padding was not necessary ? 🐙

apps/finance/app/src/components/Balances.js
```diff
@@ -80,88 +80,89 @@ function useBalanceItems(balances) {
 
 function Balances({ balances, compactMode }) {
   const theme = useTheme()
+  const { layoutName } = useLayout()
```
rperez89, line 83: ahh now we use `layoutName` instead of `compact = bellow(...)` ?

apps/finance/app/src/app-state-reducer.js
bpierre: Yes, I guess the problem will disappear when we will only pass it once to the `TokenAmount` constructor.

I tend to use a `BigInt` (or here a `BN`) by default (e.g. in [`formatTokenAmount()`](https://github.com/aragon/aragon-ui/blob/2c63df362af123a34675f679a8f5b758fa11991a/src/utils/format.js#L69)), but only because the amount is also a `BigInt`, and we can’t calculate a `Number` with a `BigInt`.

apps/finance/app/src/components/Balances.js
bpierre: I messed up during the merge, there is another `padding-right` [right below](https://github.com/aragon/aragon-apps/pull/1121/files/1dfd55665a673af8007b3365b1dbeff62f87df0b#diff-e842cf2b89a742592b21eeebff9f7086R143) :see_no_evil:

apps/finance/app/src/components/Balances.js
bpierre: We can yes, but I am thinking about merging the two because it’s a bit confusing.

apps/finance/app/src/app-state-reducer.js
sohkai: > Yes, I guess the problem will disappear when we will only pass it once to the TokenAmount constructor.

Yes, agreed!

I mention it here because this app is the only one of ours that uses only numbers for the decimals in some places (in the `tokens` data).

apps/finance/app/src/components/Balances.js
```diff
@@ -1,152 +1,170 @@
...
+        }
+      } catch (err) {
+        // The !cancelled check is needed in case:
+        //  1. The fetch() request is ongoing.
+        //  2. The component gets unmounted.
+        //  3. An error gets thrown.
+        //
+        //  Assuming the fetch() request keeps throwing, it would create new
+        //  requests even though the useEffect() got cancelled.
+        if (!cancelled) {
```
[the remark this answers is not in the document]
bpierre: I think it could be important to keep trying, e.g. if the service is temporarily down. What do you think of adding more delay one every attempt? https://github.com/aragon/aragon-apps/pull/1121/commits/2ebcf8978d41b2d6d62fe0f50ed830c8e9431822

apps/finance/app/src/components/Balances.js
```diff
@@ -1,99 +1,119 @@
...
+      decimals,
+      symbol,
+      verified,
+    }))
+  }, [balances, convertRates])
+
+  return balanceItems
+}
+
+function Balances({ balances, compactMode }) {
```
sohkai, line 90: Not a big deal, but if we're using `useLayout()` here anyway, should we avoid drilling `compactMode` down?

bpierre: Yes! We can even remove it from `App` entirely: https://github.com/aragon/aragon-apps/pull/1121/commits/5de1e668a1e15b2cd69478a76b5566206b251f43

bpierre:
@rperez89 :tada::partying_face::man_dancing:

PR #1124, opened 2020-04-10
Tokens: Fix hidden you pill on sidebar

### LINKED ISSUE #1008: Token Manager: "you" pill is partially hidden in sidebar
in the ownership distribution tab, the pill indicating the current unlocked account is partially hidden behind the percentage

uierror

delfipolito:
This PR  closes  #1008 
image

With long custom labels it looks like this: 
image

### DIFF
--- a/apps/token-manager/app/src/components/InfoBoxes.js
+++ b/apps/token-manager/app/src/components/InfoBoxes.js
@@ -115,8 +115,9 @@ function InfoBoxes({
                 <LocalIdentityBadge
                   entity={account}
                   connectedAccount={isCurrentUser}
+                  label={isCurrentUser && 'YOU'}
+                  labelStyle={isCurrentUser && `color: ${theme.tagIndicatorContent};`}
                 />
-                {isCurrentUser && <You />}
               </div>
             )
           }}
--- a/apps/token-manager/app/src/components/InfoBoxes.js
+++ b/apps/token-manager/app/src/components/InfoBoxes.js
@@ -111,15 +111,18 @@ function InfoBoxes({
           items={stakes}
           renderLegendItem={({ item: account }) => {
             const isCurrentUser = addressesEqual(account, connectedAccount)
-            const [label, showLocalIdentityModal] = useIdentity(account)
-            
+            const [label] = useIdentity(account)
+
+            const identityBadgeProps = {
+              defaultLabel: !Boolean(label) && isCurrentUser ? 'YOU' : '',
+              labelStyle: isCurrentUser ? '' + (Boolean(label) ? 'max-width: 83px;' : `color: ${theme.tagIndicatorContent};` ) : 'max-width: 117px;'
+            }
             return (
               <div>
                 <LocalIdentityBadge
                   entity={account}
                   connectedAccount={isCurrentUser}
-                  customLabel={!Boolean(label) && isCurrentUser && 'YOU'}
-                  labelStyle={(!Boolean(label) && isCurrentUser ? `color: ${theme.tagIndicatorContent};` : '')}
+                  {...identityBadgeProps}
                 />
                 {isCurrentUser && Boolean(label) && <You />}
               </div>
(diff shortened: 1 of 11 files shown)


apps/token-manager/app/src/components/InfoBoxes.js
```diff
@@ -110,13 +111,17 @@ function InfoBoxes({
           items={stakes}
           renderLegendItem={({ item: account }) => {
             const isCurrentUser = addressesEqual(account, connectedAccount)
+            const [label, showLocalIdentityModal] = useIdentity(account)
```
sohkai, line 114: We generally leave the "trailing" arguments if they're not used:

```suggestion
            const [label] = useIdentity(account)
```

apps/token-manager/app/src/components/InfoBoxes.js
```diff
@@ -110,13 +111,17 @@ function InfoBoxes({
...
             return (
               <div>
                 <LocalIdentityBadge
                   entity={account}
                   connectedAccount={isCurrentUser}
+                  customLabel={!Boolean(label) && isCurrentUser && 'YOU'}
+                  labelStyle={(!Boolean(label) && isCurrentUser ? `color: ${theme.tagIndicatorContent};` : '')}
                 />
-                {isCurrentUser && <You />}
+                {isCurrentUser && Boolean(label) && <You />}
```
sohkai, line 124: Ah, I wonder if we still need to show this? I understood the change as always showing just the identity badge.

apps/token-manager/app/src/components/InfoBoxes.js
```diff
@@ -110,13 +111,17 @@ function InfoBoxes({
...
             const isCurrentUser = addressesEqual(account, connectedAccount)
+            const [label, showLocalIdentityModal] = useIdentity(account)
+            
             return (
               <div>
                 <LocalIdentityBadge
                   entity={account}
                   connectedAccount={isCurrentUser}
+                  customLabel={!Boolean(label) && isCurrentUser && 'YOU'}
+                  labelStyle={(!Boolean(label) && isCurrentUser ? `color: ${theme.tagIndicatorContent};` : '')}
```
sohkai, line 122: I'm wondering—maybe we should move this logic into the `LocalIdentityBadge`.

We already provide a prop via `connectedAccount` to control whether or not it's "you", and we can detect and build off of this in that component. We can leave a prop to enable this "special" connected mode, to default the label and style.

delfipolito: I didnt wanted to do it in `LocalIdentityBadge` cause I thought this change was only going to be used here because we dont have much space.

apps/token-manager/app/src/components/InfoBoxes.js
delfipolito: So, when we have a custom label (like in the first case of the picture) I have to show the custom label or 'YOU'? I think the two things are relevant
I am only replacing for 'YOU' in the badge when it doesnt have a custom label

sohkai: Gotcha. I think keeping the "you" makes sense then. Let's double check what happens if you use a long label; ideally we will keep the "you" full width and shrink the label's size.

apps/token-manager/app/src/components/InfoBoxes.js
sohkai: Yes... I was thinking it could be a special mode of the `LocalIdentityBadge` but this falls apart with needing the label to detect whether we should display the "you".

apps/token-manager/app/src/components/LocalIdentityBadge/LocalIdentityBadge.js
```diff
@@ -9,9 +9,10 @@ const LocalIdentityBadge = ({ entity, ...props }) => {
   const network = useNetwork()
   const [label, showLocalIdentityModal] = useIdentity(entity)
   const handleClick = () => showLocalIdentityModal(entity)
+  
   return (
     <IdentityBadge
-      label={label || ''}
+      label={label || props.customLabel }
```
sohkai, line 15: What do you think about renaming this prop to something else, e.g. `defaultLabel` or `label`?

If we pick a different prop name than `label`, we should also put it in the prop types!

PR #1126, opened 2020-04-10
Token Manager: Identify using token name rather than symbol

### LINKED ISSUE #1061: Token App Label should show token name rather than symbol
**Is your feature request related to a problem? Please describe.**
Token Symbols are typically all caps and limited to a few characters. This is makes sense as a ticker symbol, but isn't very friendly. Currently the client will use the Token Symbol rather than Token Name when multiple token managers are installed in an organization. It would be preferrable (imo) if we used Token Name instead as there seems to be plenty of space for most token names and it looks cleaner. 

**Describe the solution you'd like**
Use token name instead of token symbol when displaying instances of the token manager app. 

**Describe alternatives you've considered**
A better and more complete solution would be allow organizations to manage custom labels for apps displayed in the menu bar. This would be a much better solution and would make this issue irrelevant. 

**Additional context**
an example 
image
My current work around is to just make the token symbol the same as the token name, though this is not possible when using the onboarding template, only from the CLI, and it's a bit of a hack that might present problems down the line if third party apps require token symbol to be short.

fabriziovigevani:
closes #1061 

Do you think 16 characters is a reasonable max length for a token name?

### DIFF
--- a/apps/token-manager/app/src/script.js
+++ b/apps/token-manager/app/src/script.js
@@ -3,6 +3,7 @@ import tokenSettings, { hasLoadedTokenSettings } from './token-settings'
 import { addressesEqual } from './web3-utils'
 import tokenAbi from './abi/minimeToken.json'
 
+const IDENTIFIER_MAX_LENGTH = 16
 const app = new Aragon()
 
 /*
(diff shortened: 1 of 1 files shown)


bpierre:
Should we always use the full name, and use `text-overflow: ellipsis` in the menu (I think it’s already the case)?

fabriziovigevani:
Ah yes good point

sohkai:
@bpierre 

> Should we always use the full name, and use text-overflow: ellipsis in the menu (I think it’s already the case)?

I was worried in the case of some Tags, e.g. in Permissions, that render the Token's identifier overfilling.

bpierre:
> I was worried in the case of some Tags, e.g. in Permissions, that render the Token's identifier overfilling.

I think we might want to do this at the UI level? That way, we would be able to use a `title` with the full name for example.

sohkai:
> I think we might want to do this at the UI level? That way, we would be able to use a title with the full name for example.

The problem with trying to do this at the UI level is that app identifiers are treated as an abstract concept by the Client and cannot assume fallbacks or etc based on text length (at least, at the moment).

PR #1129, opened 2020-04-12
Finance: Use @aragon/ui formatters

### LINKED ISSUE #756: Finance: Use BN for amounts when exporting to CSV

fabriziovigevani:
closes #756


sohkai:
@fabriziovigevani Hmm, I think we may want to defer this for the formatting utilities aragonUI will expose.

bpierre:
It’s available now! :point_right:  https://github.com/aragon/aragon-ui/blob/bc35fec2f8165c370ff8c2760f7a0f1073b5db71/src/utils/format.js#L53-L67

fabriziovigevani:
Closing in favor of https://github.com/aragon/aragon-apps/pull/1121

PR #1130, opened 2020-04-13
Finance, Token Manager: Forms MAX button 

JuanCarlosFrutos:
Closes https://github.com/aragon/aragon/issues/1301

It adds 'MAX' button for selecting max amount available. The button is added on these sections: 

- [x] Finance: Deposit
- [x] Finance: Withdrawal 
- [x] Token Manager: Add/Remove token 

**Concerns**

1. I can't input for selecting an amount on the agent app (
According to the description of the issue, there should be).
2. I'm not sure about test strategy on this repo so PR hasn't tested.

### DIFF
--- a/apps/finance/app/src/components/AmountInput.js
+++ b/apps/finance/app/src/components/AmountInput.js
@@ -1,0 +1,50 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import { ButtonBase, TextInput, noop, GU, RADIUS } from '@aragon/ui'
+
+const MAX_BUTTON_WIDTH = 6 * GU
+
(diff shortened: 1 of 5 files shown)


john-light:
Just noticed this:

> Token Manager: Add/Remove token

For the "add tokens" feature I'm not sure it makes sense to have a "max" button because in our templates there is either a max of 1 or no maximum (and in the latter case I'm not sure what pressing the "Max" button would do).

Thoughts?

bpierre:
@john-light Nice catch! :pray: I opened a PR to only display it on remove: https://github.com/aragon/aragon-apps/pull/1150

PR #1131, opened 2020-04-14
Migrate to Buidler plugin

### LINKED ISSUE #1111: Migrate to buidler-aragon
- [x] Migrate each app (see original PR: https://github.com/aragon/aragon-apps/pull/1084)
  - [x] Agent (PR: https://github.com/aragon/aragon-apps/pull/1092)
  - [x] Finance (PR: https://github.com/aragon/aragon-apps/pull/1109)
  - [x] Token Manager  (PR: #1119)
  - [x] Vault (PR: https://github.com/aragon/aragon-apps/pull/1108)
  - [x] Voting (PR: #1125)
- [ ] Upgrade aragonOS across all apps
- [ ] Fix tests (see below)
- [ ] Fix linting
- [ ] Fix any other anomalies

-------

Tests:
* WIP Fix https://github.com/aragon/aragon-cli/issues/1509 (vault)
  * Failing test: set up the default vault correctly to recover ETH from the kernel - ALSO FAILS ON MASTER
* WIP Fix https://github.com/aragon/aragon-cli/issues/1484 (agent)
  * Also runs vault tests, and as stated above, one of them fails - ALSO FAILS ON MASTER
* WIP Fix https://github.com/aragon/aragon-cli/issues/1483 (finance)
  * Three tests failing:
      * "can recover ETH using AragonApp#transferToVault" - ALSO FAILS ON MASTER
      * "fails when too many period transitions are needed"
          * fails, unless gas is set to 3e6 (ten times as much!)
          * passes in master with 3e5 gas, even if ganache-cli.sh is modified to use istanbul
          * Buidler currently has an issue where the fork cannot be modified (reported, fix soon)
          * Also fails when running npx buidler test --network localhost with a running ganache-cli instance with any fork.

sohkai:
Fixes #1111 

This PR introduces a vanilla setup of buidler development tool. It includes several network configurations both for local development and interacts with real chains. It also has ganache configured to be use as an alternative local development chain (use  with `--network ganache`).

Note the gas reporter packages don't yet support handling the start and stop of the chain. We need to start it manually for now as explained in https://github.com/cgewecke/buidler-gas-reporter/tree/master#using-buidlerevm-warning.

TODO:
- [x] Support buidlerevm as default test chain. Depends on https://github.com/aragon/contract-helpers/pull/55.
- [x] Review skipped test. Seems it was affected by the Byzantium hard fork
- [x] Make coverage work, update to latest solidity-coverage
- [x] Take another look at the .gitignore / other configuration files
- [x] Update `truffle-extract` to support buidler. As discussed in https://github.com/aragon/aragon-apps/pull/1131#discussion_r468681133.

Move to address on a future PR:
- [ ] Migrate smart contract lint rules from .solium to .solhint (afterwards we should remove .soliumrc.json)


apps/vault/test/vault_shared.js
```diff
@@ -456,7 +456,7 @@ module.exports = (vaultName, { accounts, artifacts, web3 }) => {
         })
 
         // TODO: Review, this test is most likely afected by the byzantium upgrade
-        it('set up the default vault correctly to recover ETH from the kernel', async () => {
+        xit('set up the default vault correctly to recover ETH from the kernel', async () => {
```
sohkai, line 459: Seems like there might be a typo here

apps/vault/test/vault_shared.js
```diff
@@ -295,9 +295,9 @@ module.exports = (vaultName, { accounts, artifacts, web3 }) => {
 
             const vaultBalance = await token.balanceOf(vault.address)
             assert.equal(vaultBalance, 5, 'token accounting should be correct')
-            assert.deepEqual(
-              await vault.balance(token.address),
-              vaultBalance,
+            assert.equal(
```
sohkai, line 298: Ah, it's better to use the string comparison here—that way, the test failure will show both the expected and actual values in the output.

0xGabi: Oh nice, I didn't realize that. Thanks

apps/agent/package.json
```diff
@@ -2,24 +2,20 @@
...
   "scripts": {
-    "compile": "truffle compile",
-    "build": "cd app && npm run build",
-    "apm:prepublish": "npm run compile",
-    "apm:publish:major": "aragon apm publish major --files public/ --prepublish-script apm:prepublish",
-    "apm:publish:minor": "aragon apm publish minor --files public/ --prepublish-script apm:prepublish",
-    "apm:publish:patch": "aragon apm publish patch --files public/ --prepublish-script apm:prepublish",
+    "console": "buidler console",
+    "compile": "buidler compile --force",
+    "devchain": "buidler node",
```
sohkai, line 7: We probably don't need this :)

apps/agent/buidler.config.js
```diff
@@ -0,0 +1,27 @@
+const { usePlugin } = require('@nomiclabs/buidler/config')
+
+usePlugin("@nomiclabs/buidler-truffle5")
+usePlugin('buidler-gas-reporter')
+usePlugin('solidity-coverage')
+
+module.exports = {
+  defaultNetwork: 'localhost',
+  networks: {
```
sohkai, line 9: I'm wondering if it'd be worth adding some default networks like rinkeby and mainnet? I've still not quite figured out how to do the account management bit with buidler.

apps/agent/package.json
```diff
@@ -2,24 +2,20 @@
...
+    "build": "cd app && yarn build",
     "lint": "solium --dir ./contracts",
-    "test": "TRUFFLE_TEST=true npm run ganache-cli:test",
-    "test:gas": "GAS_REPORTER=true npm test",
-    "coverage": "npx truffle run coverage",
-    "ganache-cli:test": "./node_modules/@aragon/contract-helpers-test/scripts/ganache-cli.sh",
-    "abi:extract": "truffle-extract --output abi/ --keys abi",
```
sohkai, line 16: Right, we may want to replace this with something else. Not sure if there's a buidler plugin to do this, but it would be great to move the ABI out of `artifacts`.

apps/agent/scripts/ganache-cli.sh
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env bash
```
sohkai, line 1: I don't think there's much need for this, but perhaps I've missed something.

1. Perhaps it would be worth using the ganache plugin directly? I was trying this, and while it doesn't seem to fix the failing tests, it allows us to remove this script :).
2. I'm still wondering if we shouldn't just use buidlerevm by default; it appears to run much much faster on tests (and the only failures are due to revert issues that we should fix in `contract-helpers-test`)
3. The solidity coverage plugin appears to work well out of the box, and it controls starting and stopping ganache itself (e.g. running `npx buidler coverage --network coverage` seems to just work)

.gitignore
```diff
@@ -1,5 +1,5 @@
-# aragon-cli
-*/**/artifact.json
+# buidler
+ **/**/dist
```
sohkai, line 2: Does buidler product a `dist/` folder? In any case, it should be caught by the `*/*/dist` rule below?

apps/agent/package.json
0xGabi: Maybe we can update it to be a super simple plugin? Otherwise, if I update this line: https://github.com/sohkai/truffle-extract/blob/master/index.js#L36 with `npx buidler compile` the package keep working fine. I don't have a strong preference, do you?

apps/agent/buidler.config.js
0xGabi: Good point, I'll include a few network configurations.

apps/agent/scripts/ganache-cli.sh
0xGabi: No, you are right. Last time I checked (recently) it was not possible and was handle with a similar script than ours. Now the tools do the heavy lift themselves or use buidlerevm 🙌

apps/agent/package.json
0xGabi: Maybe a nice to have. The gas-reporter don't manage the node automatically if we would like to use buidlerevm as explained here: https://github.com/cgewecke/buidler-gas-reporter/tree/master#using-buidlerevm-warning

apps/agent/package.json
0xGabi: I decide to fork your package and publish a new version for buidler instead: https://github.com/0xGabi/buidler-extract

sohkai: Sounds good! Honestly not sure if this fits well as a small buidler plugin, but I can see it being useful!

I was thinking about this earlier, and in reality, we may actually want to have a tool that lets us configure what we publish. For most apps, there's a lot of extra fluff in the artifacts (e.g. Voting only needs `Voting.json`, not all the other artifacts).

apps/agent/package.json
```diff
@@ -7,10 +7,10 @@
     "devchain": "buidler node",
     "build": "cd app && yarn build",
     "lint": "solium --dir ./contracts",
-    "test": "yarn ganache-cli:test",
-    "test:gas": "GAS_REPORTER=true yarn ganache-cli:test",
-    "coverage": "SOLIDITY_COVERAGE=true npm run ganache-cli:test",
-    "ganache-cli:test": "./scripts/ganache-cli.sh"
+    "test": "buidler test",
+    "test:gas": "REPORT_GAS=true buidler test --network localhost",
```
sohkai, line 11: Not sure, but I don't think the `localhost` network exists anymore in the buidler config?

0xGabi: As we are using the gas reporter with the buidlerevm and they don't yet handle the start and stop of the chain. We need to manually ran `npx buidler node` and then connect to the chain with `--network localhost`. As explained on 
- https://github.com/cgewecke/buidler-gas-reporter/tree/master#using-buidlerevm-warning
- https://buidler.dev/buidler-evm/#connecting-to-buidler-evm-from-wallets-and-other-software

0xGabi: I include a comment to give more context about it: https://github.com/aragon/aragon-apps/pull/1131/commits/11146312c843878283e9168efafcb4c3f6b85e75#diff-31b24dc429d79c5898460ef28314ee00R38

apps/agent/buidler.config.js
```diff
@@ -17,6 +35,11 @@ module.exports = {
       runs: 10000,
     },
   },
+  // The gas reporter plugin do not properly handle the buidlerevm
+  // chain yet. In the mean time we should 'npx buidler node' and
+  // then attach to running process using '--network localhost' as
+  // explained in: https://buidler.dev/buidler-evm/#connecting-to-buidler-evm-from-wallets-and-other-software.
+  // You can also ran 'yarn devchain' and on a separate window 'yarn test:gas'
```
sohkai, line 47:
```suggestion
  // You can also run 'yarn devchain' and on a separate terminal run 'yarn test:gas'
```

apps/agent/buidler.config.js
```diff
@@ -0,0 +1,51 @@
...
+      accounts: [
+        process.env.ETH_KEY ||
+          '0xa8a54b2d8197bc0b19bb8a084031be71835580a01e70a45a13babd16c9bc1563',
+      ],
+    },
+    frame: {
+      httpHeaders: { origin: 'buidler' },
+      url: 'http://localhost:1248',
+    },
+    ganache: {
```
sohkai, line 31: I'd move this one to be above `coverage`

apps/agent/buidler.config.js
```diff
@@ -0,0 +1,51 @@
...
+usePlugin('@nomiclabs/buidler-truffle5')
+usePlugin('buidler-gas-reporter')
+usePlugin('solidity-coverage')
+
+module.exports = {
+  networks: {
+    coverage: {
+      url: 'http://localhost:8555',
+    },
+    mainnet: {
```
sohkai, line 13: Perhaps we can leave a few quick comments about the development networks (`ganache` and `coverage`) and the real ones?

apps/agent/package.json
sohkai: Gotcha, so the `--network local` flag would tell it to connect to the separately run `buidlerevm` instance?

0xGabi: Yes. The `buidlerevm` chain starts a JSON-RPC and WebSocket server and exposes it at `http://localhost:8545`. Then the buidler default configuration object has the following format:
```
{
    localhost: {
      url: "http://127.0.0.1:8545"
    },
    buidlerevm: {
      // See its defaults
    }
}
```

That's why you can use `--network localhost`. Maybe we should be explicit about `localhost` network?

0xGabi:
We have a green CI 🎉

@sohkai if you consider it's ok, we can merge.

apps/agent/test/agent_shared.js
```diff
@@ -734,11 +734,14 @@ module.exports = (agentName, { accounts, artifacts, web3 }) => {
         context('> Designated signer', () => {
           const ethSign = async (hash, signer) => {
             const packedSig = await web3.eth.sign(hash, signer)
+            const chainId = await web3.eth.net.getId()
+            // If we detect buidlerevm we don't use ganache encoding
+            const offset = chainId === 31337 ? 0 : 27
```
sohkai, line 739: Ohhh interesting, are there docs for this? I'd like to leave a few more comments :).

We may also want to push something like this into the contract-helpers (we can do it later though)

apps/voting/.solcover.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   skipFiles: [
     'test',
     '@aragon/os',
-    '@aragon/apps-shared-minime',
+    '@aragon/minime',
```
sohkai, line 5: Just to double check, did we update the package.jsons to use `@aragon/minime` too?

0xGabi: Yes, this was updated on `package.json` but we missed to update `.solcover.js`.

apps/agent/test/agent_shared.js
0xGabi: This was pointing out by Pato from buidler. He told me that ganache encodes the v value of the signature as 0 or 1, while it should be 27 or 28. But buidler uses the same values as geth.

Related discussions about the topic:
- https://github.com/ethereumjs/ethereumjs-util/pull/67#issuecomment-274926471
- https://github.com/trufflesuite/ganache-core/issues/555

sohkai: Interesting, let's leave these in the inline code!

Issue #1132, opened 2020-04-15
Agreements: Implement app


facuspagnuolo:
The implementation has already been started in https://github.com/aragon/aragon-apps/pull/1133

To discuss:
- [x] Support `IForwarder` interface?
- [x] Support Staking ERC900?
- [x] Support cool-down period? (post-execution disputable period)

To implement:
- [x] Add inline documentation
- [x] Add readme and update main readme
- [x] Finalize unit tests
- [x] Implement gas cost tests
- [x] Implement integration tests (agreements changes, settings changes, failing actions, etc)
- [x] Organize error messages
- [x] Optimize setting variables historic values: collateral token, period durations
- [x] Model staking permissions to support token balance oracles
- [x] Review events logged information
- [x] Review functions/variables naming
- [x] Optimize SLOADs
- [x] Adjust compiler optimizer config

PR #1133, opened 2020-04-15
Agreement: Implement protocol for disputable apps

facuspagnuolo:
Fixes https://github.com/aragon/aragon-apps/issues/1132

image

### Related PRs
- Disputable base app (https://github.com/aragon/aragonOS/pull/581)
- Disputable voting app (https://github.com/aragon/aragon-apps/pull/1149)
- Aragon Court subgraph - dispute metadata parsing (https://github.com/aragon/court-subgraph/pull/67)

### Current version - Rinkeby
- Testing DAO [agreement](https://rinkeby.aragon.org/#/agreement)
- Registered under [`agreement.open.aragonpm.eth`](https://rinkeby.etherscan.io/tx/0xf16b22950d1a7c18697d14dd6c80c1673586fea17782fbae496a2c7ee491840e)
  - App ID `0x34c62f3aec3073826f39c2c35e9a1297d9dbf3cc77472283106f09eee9cf47bf`
  - Contract [`0x1C0032515055387f49053d2A20b5A5a5ef365b94`](https://rinkeby.etherscan.io/address/0x1C0032515055387f49053d2A20b5A5a5ef365b94)
  - Content [`QmfRrtVfDS2udzeGDhHpfkBSm5AWRr6ANi7EzkkVVEcbEc`](https://ipfs.io/ipfs/QmfRrtVfDS2udzeGDhHpfkBSm5AWRr6ANi7EzkkVVEcbEc)

### Old version - Rinkeby
- Testing DAO [agreementstest](https://rinkeby.aragon.org/#/agreementstest)
- Agreements [subgraph](https://thegraph.com/explorer/subgraph/facuspagnuolo/agreements-app-rinkeby)
- Registered under [`agreements.open.aragonpm.eth`](https://rinkeby.etherscan.io/tx/0x7faf403ec6f4f96d63320fd08e0747857e3f5d813587d4d721096a67ceb406bf)
  - App ID `0x980c281816072b3147b96fa284b7b1e78d51f7df83e33073276b4d7e48b44e8f`
  - Contract [`0x108D4b8C7FBdFd56065a52cFA698fE26De27c183`](https://rinkeby.etherscan.io/address/0x108D4b8C7FBdFd56065a52cFA698fE26De27c183)
  - Content [`QmdoRD25mxmDkrpgEjH9jcgaM1RnwpVxzHVzhfUjrKiHwv`](https://ipfs.io/ipfs/QmdoRD25mxmDkrpgEjH9jcgaM1RnwpVxzHVzhfUjrKiHwv)


apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,872 @@
...
+        uint256 amount = _getChallengeStake(_setting);
+        if (amount > 0) {
+            require(collateralToken.transfer(_to, amount), ERROR_COLLATERAL_TOKEN_TRANSFER_FAILED);
+        }
+    }
+
+    function _returnArbitratorFees(Challenge storage _challenge) internal {
+        uint256 amount = _challenge.arbitratorFeeAmount;
+        if (amount > 0) {
+            require(_challenge.arbitratorFeeToken.transfer(_challenge.challenger, amount), ERROR_COLLATERAL_TOKEN_TRANSFER_FAILED);
```
izqui, line 682:
```suggestion
            require(_challenge.arbitratorFeeToken.transfer(_challenge.challenger, amount), ERROR_ARBITRATOR_FEE_RETURN_FAILED);
```

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,872 @@
...
+
+        Challenge storage challenge = action.challenge;
+        uint256 disputeId = _createDispute(action, setting);
+        challenge.state = ChallengeState.Disputed;
+        challenge.disputeId = disputeId;
+        disputes[disputeId].actionId = _actionId;
+        emit ActionDisputed(_actionId);
+    }
+
+    function submitEvidence(uint256 _disputeId, bytes _evidence, bool _finished) external {
```
izqui, line 293: Should we use `_actionId` as a param to this function for consistency with all the other functions that Agreement users call?

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,872 @@
...
+        (Action storage action, Setting storage setting) = _getActionAndSetting(_actionId);
+        require(_canDispute(action, setting), ERROR_CANNOT_DISPUTE_ACTION);
+        require(msg.sender == action.submitter, ERROR_SENDER_NOT_ALLOWED);
+
+        Challenge storage challenge = action.challenge;
+        uint256 disputeId = _createDispute(action, setting);
+        challenge.state = ChallengeState.Disputed;
+        challenge.disputeId = disputeId;
+        disputes[disputeId].actionId = _actionId;
+        emit ActionDisputed(_actionId);
```
izqui, line 290: Let's log the `disputeId` and `arbitrator` here as well to allow having a link to the dispute just by looking at logs

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,872 @@
...
+    event ActionScheduled(uint256 indexed actionId);
+    event ActionChallenged(uint256 indexed actionId);
+    event ActionSettled(uint256 indexed actionId);
+    event ActionDisputed(uint256 indexed actionId);
+    event ActionAccepted(uint256 indexed actionId);
+    event ActionVoided(uint256 indexed actionId);
+    event ActionRejected(uint256 indexed actionId);
+    event ActionCancelled(uint256 indexed actionId);
+    event ActionExecuted(uint256 indexed actionId);
+    event SettingChanged(uint256 indexed settingId);
```
izqui, line 83: I imagine that these are provisional, but we should think about what other data we want to log to make state reducing easier

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,872 @@
...
+        // Create dispute
+        address submitter = _action.submitter;
+        require(feeToken.safeTransferFrom(submitter, address(this), missingFees), ERROR_ARBITRATOR_FEE_DEPOSIT_FAILED);
+        require(feeToken.safeApprove(recipient, totalFees), ERROR_ARBITRATOR_FEE_APPROVAL_FAILED);
+        uint256 disputeId = _setting.arbitrator.createDispute(DISPUTES_POSSIBLE_OUTCOMES, _setting.content);
+
+        // Update action and submit evidences
+        address challenger = challenge.challenger;
+        _submitEvidence(disputeId, submitter, _action.context, false);
+        _submitEvidence(disputeId, challenger, challenge.context, false);
```
izqui, line 537: Hard problem, but it would be great to have a way to not put these in storage if the only thing that we doing is logging them later

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,872 @@
...
+        emit Ruled(arbitrator, _disputeId, _ruling);
+
+        if (_ruling == DISPUTES_RULING_SUBMITTER) {
+            _rejectChallenge(action, setting);
+            emit ActionAccepted(dispute.actionId);
+        } else if (_ruling == DISPUTES_RULING_CHALLENGER) {
+            _acceptChallenge(action, setting);
+            emit ActionRejected(dispute.actionId);
+        } else {
+            _voidChallenge(action, setting);
```
izqui, line 331: In this case, is it possible that Aragon Court could have refunded fees (in case no jurors voted for example) and do we want to handle it?

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,872 @@
...
+        (address recipient, ERC20 feeToken, uint256 missingFees, uint256 totalFees) = _getMissingArbitratorFees(
+            _setting,
+            challengerFeeToken,
+            challengerFeeAmount
+        );
+
+        // Create dispute
+        address submitter = _action.submitter;
+        require(feeToken.safeTransferFrom(submitter, address(this), missingFees), ERROR_ARBITRATOR_FEE_DEPOSIT_FAILED);
+        require(feeToken.safeApprove(recipient, totalFees), ERROR_ARBITRATOR_FEE_APPROVAL_FAILED);
```
izqui, line 531: Should we always set the approval to 0 first to prevent approvals from failing in case that the arbitrator didn't transfer all the tokens that it could have? I don't think this should ever happen with Aragon Court, but if it were to happen with an arbitrable it could block all new disputes

apps/agreement/contracts/Agreement.sol
facuspagnuolo: This comes from the Arbitrable interface

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,872 @@
...
+        require(_amount > 0, ERROR_INVALID_UNSTAKE_AMOUNT);
+        _unstakeBalance(msg.sender, _amount);
+    }
+
+    function receiveApproval(address _from, uint256 _amount, address _token, bytes /* _data */) external authAddr(_from, STAKE_ROLE) {
+        require(msg.sender == _token && _token == address(collateralToken), ERROR_SENDER_NOT_ALLOWED);
+        _stakeBalance(_from, _from, _amount);
+    }
+
+    function schedule(bytes _context, bytes _script) external {
```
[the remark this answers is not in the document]
facuspagnuolo: Yup :)

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,872 @@
...
+    Setting[] private settings;
+    mapping (address => Stake) private stakeBalances;
+    mapping (uint256 => Dispute) private disputes;
+
+    function initialize(
+        string _title,
+        bytes _content,
+        ERC20 _collateralToken,
+        uint256 _collateralAmount,
+        uint256 _challengeLeverage,
```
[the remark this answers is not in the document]
facuspagnuolo: Yea, let's do that!

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,872 @@
...
+        _;
+    }
+
+    string public title;
+    ERC20 public collateralToken;
+
+    Action[] private actions;
+    Setting[] private settings;
+    mapping (address => Stake) private stakeBalances;
+    mapping (uint256 => Dispute) private disputes;
```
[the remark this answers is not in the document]
facuspagnuolo: Nice catch, let's have these per arbitrator address

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,872 @@
...
+        require(isContract(address(_arbitrator)), ERROR_ARBITRATOR_NOT_CONTRACT);
+
+        uint256 id = settings.length++;
+        settings[id] = Setting({
+            content: _content,
+            collateralAmount: _collateralAmount,
+            challengeLeverage: _challengeLeverage,
+            arbitrator: _arbitrator,
+            delayPeriod: _delayPeriod,
+            settlementPeriod: _settlementPeriod
```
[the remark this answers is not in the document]
facuspagnuolo: I see the reason behind it, but I'm hesitant about adding a strong validation about these in the smart contract. I was thinking about encouraging users from the UI to use realistic values for these instead. WDYT?

apps/agreement/contracts/Agreement.sol
facuspagnuolo: We shouldn't handle that here cause it will occur when penalties are settled in court. But it's an edge case we should consider somehow.

apps/agreement/contracts/Agreement.sol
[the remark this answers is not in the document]
facuspagnuolo: Just for the record, we finally decided not to allow changing the arbitrator for now

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,872 @@
...
+        bytes _content,
+        uint256 _collateralAmount,
+        uint256 _challengeLeverage,
+        IArbitrator _arbitrator,
+        uint64 _delayPeriod,
+        uint64 _settlementPeriod
+    )
+        internal
+    {
+        require(isContract(address(_arbitrator)), ERROR_ARBITRATOR_NOT_CONTRACT);
```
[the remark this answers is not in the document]
facuspagnuolo: Yea it's a shame we didn't consider ERC165 for the arbitrator, although I'd implement thorough  validations from the UI to be honest

apps/agreement/contracts/Agreement.sol
facuspagnuolo: Interesting, let's see if we find a way to have that

apps/agreement/contracts/Agreement.sol
facuspagnuolo: Yup, it was provisional. I added more information now

facuspagnuolo:
@izqui thx for the review and the thorough feedback, I addressed almost all your comments. There are a few pending to be discussed

facuspagnuolo:
For the record, the Agreement app contract is super close to size limit, currently it is 23,928

apps/agreement/contracts/Agreement.sol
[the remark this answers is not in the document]
izqui: Sounds good

apps/agreement/test/agreement_permissions.js
```diff
@@ -0,0 +1,157 @@
...
+        beforeEach('mint tokens', async () => {
+          const currentBalance = await permissionToken.balanceOf(signer)
+          if (currentBalance.eq(balance)) return
+
+          if (currentBalance.gt(balance)) {
+            const balanceDiff = currentBalance.sub(balance)
+            await permissionToken.destroyTokens(signer, balanceDiff)
+          } else {
+            const balanceDiff = balance.sub(currentBalance)
+            await permissionToken.generateTokens(signer, balanceDiff)
```
izqui, line 31:
```suggestion
          const currentBalance = await permissionToken.balanceOf(holder)
          if (currentBalance.eq(balance)) return

          if (currentBalance.gt(balance)) {
            const balanceDiff = currentBalance.sub(balance)
            await permissionToken.destroyTokens(holder, balanceDiff)
          } else {
            const balanceDiff = balance.sub(currentBalance)
            await permissionToken.generateTokens(holder, balanceDiff)
```

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1358 @@
...
+        emit ActionCancelled(_actionId);
+    }
+
+    /**
+    * @notice Challenge an action #`_actionId` with a settlement offer of `@tokenAmount(self.collateralToken(): address, _settlementOffer)`
+    * @param _actionId Identification number of the action being challenged
+    * @param _settlementOffer Amount of collateral tokens the challenger would accept for resolving the dispute without involving the arbitrator
+    * @param _context Link to a human-readable text giving context for the challenge
+    */
+    function challengeAction(uint256 _actionId, uint256 _settlementOffer, bytes _context) external authP(CHALLENGE_ROLE, arr(_actionId)) {
```
izqui, line 314: Allowing only holders of a certain token to challenge actions could be a pretty common configuration for Agreements.

Since we require `msg.sender` to be the actual originator of the intent (i.e. cannot use a TokenManager forwarder as we need to `transferFrom` the account that challenges), in order for us to limit who can challenge based on a token balance we'd need to use a parametrized role with an ACL oracle or use the same approach as in `canSign`.

If we take the approach of having a custom permission here as well, we will need to decide whether it would be completely different from the signing permission, allowing the token and the amount that you need to hold to be different

apps/agreement/contracts/Agreement.sol
```diff
@@ -75,10 +75,10 @@ contract Agreement is IArbitrable, IForwarder, AragonApp {
...
     bytes32 public constant CHANGE_TOKEN_BALANCE_PERMISSION_ROLE = 0x4413cad936c22452a3bdddec48f42af1848858d1e8a8b62b7c0ba489d6d77286;
 
-    event ActionScheduled(uint256 indexed actionId);
-    event ActionChallenged(uint256 indexed actionId);
-    event ActionSettled(uint256 indexed actionId);
-    event ActionDisputed(uint256 indexed actionId);
+    event ActionScheduled(uint256 indexed actionId, address indexed submitter);
+    event ActionChallenged(uint256 indexed actionId, address indexed challenger);
+    event ActionSettled(uint256 indexed actionId, uint256 offer);
+    event ActionDisputed(uint256 indexed actionId, IArbitrator indexed arbtirator, uint256 disputeId);
```
izqui, line 81:
```suggestion
    event ActionDisputed(uint256 indexed actionId, IArbitrator indexed arbitrator, uint256 disputeId);
```

apps/agreement/contracts/Agreement.sol
```diff
@@ -559,7 +559,8 @@ contract Agreement is IArbitrable, IForwarder, AragonApp {
         // Create dispute
         address submitter = _action.submitter;
         require(feeToken.safeTransferFrom(submitter, address(this), missingFees), ERROR_ARBITRATOR_FEE_DEPOSIT_FAILED);
-        require(feeToken.safeApprove(recipient, totalFees), ERROR_ARBITRATOR_FEE_APPROVAL_FAILED);
+        _approveArbitratorFeeTokens(feeToken, recipient, 0);
```
izqui, line 809: Should we just do the two approval calls inside the `_approveArbitratorFeeTokens` function? Also let's add a comment on why we are doing this

facuspagnuolo: Let's add a comment definitely, but note that this way we are reusing the call encode, just saying because of the size limit haha

apps/agreement/contracts/Agreement.sol
facuspagnuolo: Interesting, let's try to have a separate balance condition here then :)

apps/agreement/test/agreement_integration.js
```diff
@@ -0,0 +1,201 @@
...
+    await permissionToken.generateTokens(holder40, permissionBalance.mul(4))
+    await permissionToken.generateTokens(holder50, permissionBalance.mul(5))
+  })
+
+  before('deploy agreement instance', async () => {
+    agreement = await deployer.deployAndInitializeWrapper({ collateralAmount, challengeCollateral, signers: [holder10, holder20, holder30, holder40, holder50]})
+  })
+
+  describe('integration', () => {
+    it('only holders with more than 10 permission tokens can sign', async () => {
```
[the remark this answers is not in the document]
facuspagnuolo: Yes we can :)

facuspagnuolo:
@izqui thx for the review again, I change the contract to allow having an optional token balance permission for challenges similar as it was implemented for the signing process

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
+/*
```
sohkai, line 1: Update to here.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+    * @dev Unlock and slash a number of tokens in the staking pool from a user in favour of a challenger
+    * @param _staking Staking pool for the ERC20 token to be unlocked and slashed
+    * @param _user Address of the user to be unlocked and slashed
+    * @param _unlockAmount Number of collateral tokens to be unlocked
+    * @param _challenger Address receiving the slashed tokens
+    * @param _slashAmount Number of collateral tokens to be slashed
+    */
+    function _unlockAndSlashBalance(Staking _staking, address _user, uint256 _unlockAmount, address _challenger, uint256 _slashAmount) internal {
+        if (_unlockAmount != 0 && _slashAmount != 0) {
+            _staking.unlockAndSlash(_user, _unlockAmount, _challenger, _slashAmount);
```
sohkai, line 940: ⚠️ Maybe this is addressed in the staking PR, but I would assume the staking module to handle 0 for either of the amounts, so we should just that check that one of the amounts is non-zero.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+        (, Challenge storage challenge, ) = _getChallengedAction(_actionId);
+        return _canDispute(_actionId, challenge);
+    }
+
+    /**
+    * @dev Tell whether an action's current challenge settlement can be claimed
+    * @param _actionId Identification number of the action to be queried
+    * @return True if the action settlement can be claimed, false otherwise
+    */
+    function canClaimSettlement(uint256 _actionId) external view returns (bool) {
```
sohkai, line 672: ⚠️ Not sure, but wouldn't this be a problem for multiple challenges? It would be impossible to recover funds from earlier challenges once new challenges got created?

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+
+    /**
+    * @dev Reject a challenge proposed against an action ("accept action")
+    * @param _actionId Identification number of the action to be accepted
+    * @param _action Action instance to be accepted
+    * @param _challengeId Current challenge identification number for the action
+    * @param _challenge Current challenge instance for the action
+    */
+    function _rejectChallenge(uint256 _actionId, Action storage _action, uint256 _challengeId, Challenge storage _challenge) internal {
+        _challenge.state = ChallengeState.Rejected;
```
sohkai, line 862: ⚠️ Should we also reset the action's current challenge ID to 0 when the challenge is rejected / voided?

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+        // To be safe, We first set the allowance to zero in case there is a remaining approval for the arbitrator.
+        // This is not strictly necessary for ERC20s, but some tokens, e.g. MiniMe (ANT and ANJ),
+        // revert on an approval if an outstanding allowance exists
+        _approveArbitratorFeeTokens(feeToken, disputeFeeRecipient, 0);
+        _approveArbitratorFeeTokens(feeToken, disputeFeeRecipient, totalFees);
+        uint256 disputeId = _arbitrator.createDispute(DISPUTES_POSSIBLE_OUTCOMES, _metadata);
+
+        // Return any remaining portion of the pre-paid arbitrator fees to the challenger, if necessary
+        if (challengerFeeToken != feeToken) {
+            _transfer(challengerFeeToken, _challenge.challenger, challengerFeeAmount);
```
sohkai, line 780: ⚠️ I think there may be a problem if the arbitrator lowers its fees to below what the challenger pre-paid.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+        uint256 unlockedAmount = actionCollateral - slashedAmount;
+
+        address challenger = challenge.challenger;
+        _unlockAndSlashBalance(requirement.staking, submitter, unlockedAmount, challenger, slashedAmount);
+        _transfer(requirement.token, challenger, requirement.challengeAmount);
+        _transfer(challenge.arbitratorFeeToken, challenger, challenge.arbitratorFeeAmount);
+
+        challenge.state = ChallengeState.Settled;
+        disputable.onDisputableActionRejected(action.disputableActionId);
+        emit ActionSettled(_actionId, challengeId);
```
sohkai, line 353: ⚠️ Want to be careful here:

- Is `onDisputableActionRejected` the correct event? It's a bit confusing.
- Should we also emit an event that the action was rejected?
- Should we reset the currentChallengeId back to 0?

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+
+    /**
+    * @dev Helper to build an agreement dispute metadata as "agreements:[ACTION_ID]"
+    * @param _actionId Identification number of the action to create a dispute for
+    * @return dispute metadata for the requested action
+    */
+    function _buildDisputeMetadata(uint256 _actionId) internal view returns (bytes memory) {
+        // Header "agreement:"
+        bytes memory metadata = new bytes(11);
+        assembly { mstore(add(metadata, 32), 0x61677265656d656e74733A000000000000000000000000000000000000000000) }
```
sohkai, line 1347: ⚠️ I think this is being decoded as agreements at the moment (despite the comment).

But I also wonder if it would be better to use the Agreement's appID in string or namehash form instead? Maybe the namehash is best, as it's a short bytes32 value.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+        challenge.ruling = _ruling;
+        emit Ruled(arbitrator, _disputeId, _ruling);
+
+        // TODO: implement try catch
+        if (_ruling == DISPUTES_RULING_SUBMITTER) {
+            _rejectChallenge(actionId, action, challengeId, challenge);
+        } else if (_ruling == DISPUTES_RULING_CHALLENGER) {
+            _acceptChallenge(actionId, action, challengeId, challenge);
+        } else {
+            _voidChallenge(actionId, action, challengeId, challenge);
```
sohkai, line 423: Would it be better to revert on all other `ruling`s that don't match our expected results?

Or is it better to just leave them all as `void`?

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+
+    /**
+    * @dev Tell whether a challenge exists for an action
+    * @param _actionId Identification number of the action being queried
+    * @param _challengeId Identification number of the challenge being queried
+    * @param _challenge Challenge instance associated with the challenge identification number
+    * @return True if the requested challenge exists, false otherwise
+    */
+    function _existChallenge(uint256 _actionId, uint256 _challengeId, Challenge storage _challenge) internal view returns (bool) {
+        return _existChallenge(_challengeId) && _actionId == _challenge.actionId;
```
sohkai, line 1286: Not sure if `_actionId == _challenge.actionId` is that useful.

Wouldn't it be better if we checked `action.currentChallengeId === _challengeId`?

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+    }
+
+    /**
+    * @dev Tell whether an action is challenged and if it's waiting to be answered
+    * @param _actionId Identification number of the action to be queried
+    * @param _challenge Current challenge instance for the action
+    * @return True if the action is challenged and it's waiting to be answered, false otherwise
+    */
+    function _isWaitingChallengeAnswer(uint256 _actionId, Challenge storage _challenge) internal view returns (bool) {
+        return _actionId == _challenge.actionId && _challenge.state == ChallengeState.Waiting;
```
sohkai, line 1155: Similar to other comment above, I feel like the `_actionId == _challenge.actionId` is not that useful.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+    }
+
+    /**
+    * @dev Tell whether an action can be settled
+    * @param _actionId Identification number of the action to be queried
+    * @param _challenge Current challenge instance for the action
+    * @return True if the action can be settled, false otherwise
+    */
+    function _canSettle(uint256 _actionId, Challenge storage _challenge) internal view returns (bool) {
+        return _isWaitingChallengeAnswer(_actionId, _challenge);
```
sohkai, line 1107: Just to double check—this doesn't check the time, so the submitter can also forcibly settle a challenge after the expiry?

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+    }
+
+    /**
+    * @dev Tell whether an address has permission to challenge actions on a specific disputable app
+    * @param _disputable Disputable app being queried
+    * @param _challenger Address of the challenger
+    * @return True if the challenger can be challenge actions on the disputable app, false otherwise
+    */
+    function _canPerformChallenge(IDisputable _disputable, address _challenger) internal view returns (bool) {
+        if (!hasInitialized()) {
```
sohkai, line 1036: Note that we theoretically don't need this check—everywhere we use it, it should already be checked by the action related to the disputable existing.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+        _newSetting(_title, _content, _arbitrator);
+    }
+
+    /**
+    * @notice Sign the agreement
+    */
+    function sign() external {
+        uint256 currentSettingId = _getCurrentSettingId();
+        uint256 lastSettingIdSigned = lastSettingSignedBy[msg.sender];
+        require(lastSettingIdSigned < currentSettingId, ERROR_SIGNER_ALREADY_SIGNED);
```
sohkai, line 243: So I think this currently protects this function from being used before being initialized, but we may either want to note that down or add a more explicit check.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+    * @return arbitrator Address of the IArbitrator that will be used to resolve disputes
+    */
+    function _getSettingFor(Action storage _action) internal view returns (string title, bytes content, IArbitrator arbitrator) {
+        uint256 settingId = _action.settingId;
+        require(settingId < settingsLength, ERROR_MISSING_AGREEMENT_SETTING);
+
+        Setting storage setting = settings[settingId];
+        title = setting.title;
+        content = setting.content;
+        arbitrator = setting.arbitrator;
```
sohkai, line 1256: Maybe we could just return the storage pointer for `Setting`? It looks like all the usages of this function don't use all these variables.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+
+    struct DisputableInfo {
+        bool registered;                                                    // Whether a Disputable app is registered
+        uint256 collateralRequirementsLength;                               // Identification number of the next collateral requirement instance
+        mapping (uint256 => CollateralRequirement) collateralRequirements;  // List of collateral requirements indexed by id
+    }
+
+    StakingFactory public stakingFactory;                           // Staking factory to be used for the collateral staking pools
+
+    uint256 private settingsLength;
```
sohkai, line 127: Maybe better to name these as `nextSettingsId`/`currentSettingsId` instead, since we are using them as IDs elsewhere.

I think it might also make the code a tad easier to work with if we go with `current`, as we'll avoid having to do a `-1` every time we need it.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+        ERC20 token;                        // ERC20 token to be used for collateral
+        uint256 actionAmount;               // Amount of collateral token that will be locked from the submitter's staking pool every time an action is created
+        uint256 challengeAmount;            // Amount of collateral token that will be locked from the challenger's own balance every time an action is challenged
+        uint64 challengeDuration;           // Challenge duration in seconds, during which the submitter can raise a dispute
+        Staking staking;                    // Staking pool cache for the collateral token
+    }
+
+    struct DisputableInfo {
+        bool registered;                                                    // Whether a Disputable app is registered
+        uint256 collateralRequirementsLength;                               // Identification number of the next collateral requirement instance
```
sohkai, line 121: Similar to above, perhaps we should name this variable `current/nextCollateralRequirementsId` instead?

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+    * @dev Change Agreement settings
+    * @param _title String indicating a short description
+    * @param _content Link to a human-readable text that describes the initial rules for the Agreements instance
+    * @param _arbitrator Address of the IArbitrator that will be used to resolve disputes
+    */
+    function _newSetting(string _title, bytes _content, IArbitrator _arbitrator) internal {
+        require(isContract(address(_arbitrator)), ERROR_ARBITRATOR_NOT_CONTRACT);
+
+        uint256 id = settingsLength++;
+        settings[id] = Setting({ title: _title, content: _content, arbitrator: _arbitrator });
```
sohkai, line 991: Might be worth doing a quick test; I think setting them separately:

```solidity
Setting storage setting = settings[id];
setting.title = ...
```

Is a bit cheaper (avoids a memory alloc when doing it like this).

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+            return false;
+        }
+
+        IKernel linkedKernel = kernel();
+        if (address(linkedKernel) == address(0)) {
+            return false;
+        }
+
+        bytes memory params = ConversionHelpers.dangerouslyCastUintArrayToBytes(arr(_challenger));
+        return linkedKernel.hasPermission(_challenger, address(_disputable), CHALLENGE_ROLE, params);
```
sohkai, line 1046: I think we should be able to remove these `params` now, with the ACLOracle change.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+
+        DisputableInfo storage disputableInfo = disputableInfos[msg.sender];
+        _ensureRegisteredDisputable(disputableInfo);
+
+        uint256 currentCollateralRequirementId = disputableInfo.collateralRequirementsLength.sub(1);
+        CollateralRequirement storage requirement = disputableInfo.collateralRequirements[currentCollateralRequirementId];
+        _lockBalance(requirement.staking, _submitter, requirement.actionAmount);
+
+        uint256 id = actionsLength++;
+        Action storage action = actions[id];
```
sohkai, line 270: Can do last if we have room, but we have typically errored if the struct did not exist.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+    * @param _actionId Identification number of the action to be queried
+    * @param _challenge Current challenge instance for the action
+    * @return True if the action can be disputed, false otherwise
+    */
+    function _canDispute(uint256 _actionId, Challenge storage _challenge) internal view returns (bool) {
+        if (!_isWaitingChallengeAnswer(_actionId, _challenge)) {
+            return false;
+        }
+
+        return _challenge.endDate > getTimestamp64();
```
sohkai, line 1121: In cases where we're comparing timestamps, we should cast up to uint256 as it's a bit cheaper.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+
+    StakingFactory public stakingFactory;                           // Staking factory to be used for the collateral staking pools
+
+    uint256 private settingsLength;
+    mapping (uint256 => Setting) private settings;                  // List of historic settings indexed by ID
+    mapping (address => uint256) private lastSettingSignedBy;       // List of last setting signed by user
+
+    uint256 private actionsLength;
+    mapping (uint256 => Action) private actions;                    // List of actions indexed by ID
+    mapping (address => DisputableInfo) private disputableInfos;    // Mapping of disputable address => disputable infos
```
sohkai, line 133: Nitpick, but perhaps we should also move `disputableInfos` to be right after the settings so we consistently order our concepts as:

1. Signers
1. Settings
1. Disputables
1. Actions

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+import "./staking/StakingFactory.sol";
+
+
+contract Agreement is IAgreement, AragonApp {
+    using SafeMath for uint256;
+    using SafeMath64 for uint64;
+    using SafeERC20 for ERC20;
+    using BytesHelper for bytes;
+
+    /* Arbitrator outcomes constants */
```
sohkai, line 28: Just wondering, we don't have an enum type defined in `IArbitrator` for this, right?

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+    }
+
+    /**
+    * @dev Tell whether a challenge exists for an action
+    * @param _actionId Identification number of the action being queried
+    * @param _challengeId Identification number of the challenge being queried
+    * @param _challenge Challenge instance associated with the challenge identification number
+    * @return True if the requested challenge exists, false otherwise
+    */
+    function _existChallenge(uint256 _actionId, uint256 _challengeId, Challenge storage _challenge) internal view returns (bool) {
```
sohkai, line 1285: Maybe better to rename as `_existChallengeForAction()` to avoid an override that doesn't quite do the same thing.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+
+    /**
+    * @dev Dispute an action
+    * @param _action Action instance to be disputed
+    * @param _challenge Current challenge instance for the action
+    * @return _arbitrator Address of the IArbitrator applicable for the action
+    * @return _metadata Metadata content to be used for the dispute
+    * @return Identification number of the dispute created in the arbitrator
+    */
+    function _createDispute(Action storage _action, Challenge storage _challenge, IArbitrator _arbitrator, bytes memory _metadata)
```
sohkai, line 753: Since this is doing a lot more than just creating the dispute (token transfers/etc), maybe we can inline this function into `disputeAction()`?

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+        }
+    }
+
+    /**
+    * @dev Transfer tokens to an address
+    * @param _token ERC20 token to be transferred
+    * @param _to Address receiving the tokens
+    * @param _amount Number of tokens to be transferred
+    */
+    function _transfer(ERC20 _token, address _to, uint256 _amount) internal {
```
sohkai, line 953: Perhaps we can rename these token transfer functions as:

- `_transferTo()` (currently `_transfer()`)
- `_pullFrom()` (currently `_transferFrom()`)
- `_approveFor()` (current `_approveArbitratorFeeTokens()`)

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+        return getTimestamp64() >= _challenge.endDate;
+    }
+
+    /**
+    * @dev Tell whether an action's dispute can be ruled
+    * @param _actionId Identification number of the action to be queried
+    * @param _challenge Current challenge instance for the action
+    * @return True if the action's dispute can be ruled, false otherwise
+    */
+    function _canRuleDispute(uint256 _actionId, Challenge storage _challenge) internal view returns (bool) {
```
sohkai, line 1144: Not sure we really need this indirection, maybe we can just use `_isDisputed()` directly?

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+    }
+
+    /**
+    * @dev Accept a challenge proposed against an action ("reject action")
+    * @param _actionId Identification number of the action to be rejected
+    * @param _action Action instance to be rejected
+    * @param _challengeId Current challenge identification number for the action
+    * @param _challenge Current challenge instance for the action
+    */
+    function _acceptChallenge(uint256 _actionId, Action storage _action, uint256 _challengeId, Challenge storage _challenge) internal {
```
sohkai, line 841: Wondering if we can find a better name for these three functions.

They make sense, but I find them to be confusing at a glance because we use both "challenge accepted" and "action rejected" as concepts that are synonymous with each other.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1350 @@
...
+        stakingFactory = _stakingFactory;
+
+        settingsLength++; // Setting ID zero is considered the null setting for further validations
+        _newSetting(_title, _content, _arbitrator);
+    }
+
+    /**
+    * @notice Register `_disputable`, setting its collateral requirements to:
+    * @notice - `@tokenAmount(_collateralToken: address, _actionAmount)` for submissions
+    * @notice - `@tokenAmount(_collateralToken: address, _challengeAmount)` for challenges
```
sohkai, line 159: It's unlikely we'll be able to show this via radspec at the moment, so I wonder if we shouldn't just remove them.

apps/agreement/README.md
```diff
@@ -0,0 +1,5 @@
+# Agreement
```
sohkai, line 1: I think it would also be nice to start introducing a more standard format for these readmes with information on their status, like the [voting-connectors do](https://github.com/aragonone/voting-connectors/blob/master/apps/voting-aggregator/readme.md) (for now, we can just say it's in heavy development and pending audit).

apps/agreement/.solcover.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+    norpc: true,
+    copyPackages: ['@aragon/os', '@aragon/apps-vault'],
+    skipFiles: [
+        'test',
+        '@aragon/os',
+        '@aragon/apps-vault',
```
sohkai, line 7: Feels a bit like deja-vu, but were we able to remove the `vault` dependency?

apps/agreement/arapp.json
```diff
@@ -0,0 +1,56 @@
...
+  "roles": [
+    {
+      "name": "Sign Agreement",
+      "id": "SIGN_ROLE",
+      "params": [
+        "Signer address"
+      ]
+    },
+    {
+      "name": "Challenge Agreement actions",
```
sohkai, line 38: This may be due to tooling, but we should eventually remove this from this app's `arapp.json` since it's not protecting anything here.

apps/agreement/contracts/example/RegistryApp.sol
```diff
@@ -0,0 +1,289 @@
+/*
```
sohkai, line 1: A bit of clean up, but perhaps we should make this example a part of aragonOS instead?

I think we may also have to update this with the new interfaces too (so a good thing to do as a second PR).

apps/agreement/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "@aragon/os": "aragon/aragonOS#add_disputable_base_app"
+  },
+  "devDependencies": {
+    "@aragon/apps-shared-migrations": "1.0.0",
+    "@aragon/cli": "^7.1.3",
+    "@aragon/minime": "^1.0.0",
+    "@aragon/contract-test-helpers": "^0.0.1",
+    "@aragon/truffle-config-v5": "^1.0.0",
+    "eth-gas-reporter": "^0.2.0",
+    "ethereumjs-testrpc-sc": "^6.5.1-sc.0",
```
sohkai, line 36: Quick note, I think we can remove this now, since `solidity-coverage@^0.7.0` doesn't need it anymore!

apps/agreement/contracts/test/mocks/helpers/TimeHelpersMock.sol
```diff
@@ -0,0 +1,54 @@
+pragma solidity ^0.4.24;
```
sohkai, line 1: Quick thing; does the version exported by `contract-test-helpers` not work?

Should we update that repo with this version?

apps/agreement/test/helpers/assert/assertBn.js
```diff
@@ -0,0 +1,7 @@
+const assertBn = (actual, expected, errorMsg) => {
```
sohkai, line 1: Could all of these helpers be used from or moved to the contract-test-helpers too?

apps/agreement/contracts/Agreement.sol
sohkai: Actually, thinking a bit more about this, maybe we **don't** want to reset the value.

I thought resetting it would make it easier for us to validate if an action had an ongoing challenge, but:

1. We are currently handling it fine on-chain without having to reset the value
2. Keeping the value is helpful for a frontend to easily know how many challenges an action had (its entire challenge history), without resorting to event fetching

apps/agreement/contracts/Agreement.sol
facuspagnuolo: I think it's safe to understand it as an invalid ruling (refuse - void)

facuspagnuolo: I'll add it to the list of assumptions

apps/agreement/contracts/Agreement.sol
facuspagnuolo: Given that one of the main assumptions of the Agreement app is that there can only be one challenge at a time, all the exposed functions work with the action ID and we use the `currentChallengeId` to get the challenge instance we need to work with. Thus, all the `challengeId` or `challenge` instances loaded from storage are got by doing `action.currentChallengeId`. 

However, this internal function is used only in one place.. I'll inline it

apps/agreement/contracts/Agreement.sol
facuspagnuolo: For the record, we can probably continue the discussion [here](https://github.com/aragon/aragon-apps/pull/1133#discussion_r439623006)

apps/agreement/contracts/Agreement.sol
facuspagnuolo: Yes, not answering a settlement is the same as settling, the only thing the submitter is allowed to do during the settlement window is to reject the settlement raising the dispute to court.

apps/agreement/contracts/Agreement.sol
facuspagnuolo: I tried using current before, but it doesn't fit properly for the first case where it is initialized as 0 and there is actually no instance stored yet. Let's go with `next*Id` for now :)

apps/agreement/contracts/Agreement.sol
facuspagnuolo: We don't :/

apps/agreement/contracts/Agreement.sol
facuspagnuolo: This function was inlined

apps/agreement/contracts/Agreement.sol
facuspagnuolo: Stack too deep issues my friend haha. I like the abstraction to be honest, would you prefer using a more white-box name?

apps/agreement/contracts/Agreement.sol
facuspagnuolo: What do you think about `_depositFrom` instead of `_pullFrom`?

apps/agreement/contracts/test/mocks/helpers/TimeHelpersMock.sol
facuspagnuolo: It does work, the problem is that when working with multiple contracts, in this case the agreement and disputable apps, you need to mock each of them individually. Here I modeled a clock that is shared among all these, so you only need to update one thing, and you can make sure all these instances have the same timestamp :)

So yes, I'd love to update the `contract-test-helpers` version :)

apps/agreement/test/helpers/assert/assertBn.js
facuspagnuolo: Yes, definitely, many of these should be moved to the helpers repo

apps/agreement/contracts/Agreement.sol
facuspagnuolo: Yea, I agree. For the moment I changed them to be in terms of the action instead, as that is what we are doing for the events and callbacks

apps/agreement/contracts/Agreement.sol
facuspagnuolo: Not sure I follow

apps/agreement/contracts/Agreement.sol
facuspagnuolo: One of the main assumptions of the Agreement app is that there can only be one challenge at a time, therefore you will need to close a previous challenge to challenge an action again, i.e. claiming the settlement in this case

apps/agreement/contracts/Agreement.sol
facuspagnuolo: > Should we also reset the action's current challenge ID to 0 when the challenge is rejected / voided?

I actually tried to keep it consistent so `currentChallengeID` always references to the current/last without having to reset it. I think we are safe to keep it given the exhaustive checks we are already doing on every external function that could affect the action or challenge state. Otherwise, some of those checks are wrong and we should fix them.

I think you arrived the same conclusion reading your second comment :)

apps/agreement/contracts/Agreement.sol
facuspagnuolo: Good catch! I'll add a test for this :)

apps/agreement/contracts/Agreement.sol
facuspagnuolo: Yup, let's do that

apps/agreement/contracts/Agreement.sol
facuspagnuolo: > Is `onDisputableActionRejected` the correct event? It's a bit confusing.

I lifted this as another assumption, I mean treating settled actions as rejected ones. 

> Should we also emit an event that the action was rejected?

I think it will make it ambiguous if we do that. I know we are deciding to sum it up for the disputable app treating this case as rejected, but it's simply because we will be forcing disputables to handle another scenario that in 99.9% of the cases will be handled in the same way as rejected actions.

> Should we reset the `currentChallengeId` back to 0?

Already discussed [here](https://github.com/aragon/aragon-apps/pull/1133#discussion_r439650486)

facuspagnuolo:
@sohkai thanks once again! I addressed most of your comments and answered others, let me know your thoughts!

apps/agreement/contracts/Agreement.sol
bingen: Actually, the settlement claim can only happen once, right? Because if a challenge is settled the action is closed, and therefore cannot be challenged again.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1338 @@
...
+    function _canClose(uint256 _actionId, Action storage _action) internal view returns (bool) {
+        return _canProceed(_actionId, _action) && _action.disputable.canClose(_action.disputableActionId);
+    }
+
+    /**
+    * @dev Tell whether an action can proceed.
+    * @dev An action can proceed if it is not:
+    * @dev  - Closed
+    * @dev  - Currently challenged
+    * @dev  - Ruled as voided
```
bingen, line 1085: I think these 2 lines are wrong

apps/agreement/contracts/test/mocks/helpers/TimeHelpersMock.sol
sohkai: Right, I was just thinking about how to do this (e.g. if you were building an app and wanted to mock time in other contracts as well).

Let's move it over!

apps/agreement/contracts/Agreement.sol
facuspagnuolo: I think these are fine, actions can proceed if they were ruled in favor of the submitter or voided

apps/agreement/contracts/Agreement.sol
facuspagnuolo: Yes, you are right, if there are multiple challenges, it means none of the previous ones were settled or ruled in favor of the challenger (i.e. the submitter won)

apps/agreement/contracts/Agreement.sol
```diff
@@ -761,21 +775,25 @@ contract Agreement is IAgreement, AragonApp {
...
+        _approveFor(feeToken, disputeFeeRecipient, 0);
+        _approveFor(feeToken, disputeFeeRecipient, totalFees);
         uint256 disputeId = _arbitrator.createDispute(DISPUTES_POSSIBLE_OUTCOMES, _metadata);
 
-        // Return any remaining portion of the pre-paid arbitrator fees to the challenger, if necessary
         if (challengerFeeToken != feeToken) {
-            _transfer(challengerFeeToken, _challenge.challenger, challengerFeeAmount);
+            // Return any remaining portion of the pre-paid arbitrator fees to the challenger, if necessary
+            _transferTo(challengerFeeToken, _challenge.challenger, challengerFeeAmount);
+        } else if (missingFees == 0) {
```
sohkai, line 793: I wonder if we shouldn't use `challengerFeeAmount < totalFees` as the check instead, since that's what we're subtracting.

But another thought I had was to directly return a `challengerRefund` value from `_getMissingArbitratorFees()` instead.

apps/agreement/contracts/Agreement.sol
sohkai: We use "pull" a lot when talking about token movements, but `_depositFrom()` sounds good as well!

apps/agreement/contracts/Agreement.sol
sohkai: Nope, on a second look this is good!

apps/agreement/contracts/Agreement.sol
sohkai: > I lifted this as another assumption, I mean treating settled actions as rejected ones.

Sounds good; I think this is a reasonable assumption.

My confusion from before was actually between the "challenge accepted" vs. "action accepted" terminology—I think this has been cleared up now!

> Should we also emit an event that the action was rejected?

Agreed that we can leave it out. We can treat the `ActionAccepted`, `ActionVoided`, and `ActionRejected` events as dependent on a dispute being created.

apps/agreement/contracts/Agreement.sol
```diff
@@ -0,0 +1,1338 @@
...
+            uint256 disputableActionId,
+            uint256 collateralRequirementId,
+            uint256 settingId,
+            address submitter,
+            bool closed,
+            bytes context,
+            uint256 currentChallengeId
+        )
+    {
+        Action storage action = actions[_actionId];
```
sohkai, line 535: What I meant in a previous comment about erroring if something doesn't exist was meant for the getters: https://github.com/aragon/aragon-apps/pull/1133/files/f698674b346af5052bcc488df8eb51ddbb8f3f5f#diff-9ebddb6de87cd26830aa77a8db008ced.

In the past we have added checks in getters to revert if the id doesn't exist yet. However, I'm starting to question that decision since it might lead to a large bytecode increase (at least with modifiers).

apps/agreement/contracts/Agreement.sol
sohkai: Yes, both are true!

facuspagnuolo:
@sohkai addressed some last comments and conversations we had off-line, let me know your thoughts!

apps/agreement/test/agreement/agreement_gas_cost.js
```diff
@@ -31,7 +31,7 @@ contract('Agreement', ([_, user]) => {
     })
 
     context('newAction', () => {
```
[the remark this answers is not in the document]
facuspagnuolo: Honestly don't know, I would need to check what change introduce this increase :/

PR #1137, opened 2020-04-29
Tokens: Vesting UI

delfipolito:


apps/token-manager/app/src/components/VestingInfoBoxes.js
```diff
@@ -0,0 +1,160 @@
...
+    <div
+      css={`
+        display: flex;
+        justify-content: flex-start;
+      `}
+    >
+      {icon}
+      <div
+        css={`
+          display: inline-grid;
```
sohkai, line 139: Would `inline-flex` work here? It seems a bit weird specifying `grid` but then not setting any other grid values.

apps/token-manager/app/src/components/VestingInfoBoxes.js
```diff
@@ -0,0 +1,160 @@
...
+          css={`
+            padding: ${3 * GU}px;
+          `}
+        >
+          <InfoBoxHeader
+            icon={<TokenIcon />}
+            label={'Total Tokens'}
+            content={formatTokenAmount(
+              selectedHolder.holderBalance.balance,
+              18,
```
sohkai, line 43: Note that here and below, we shouldn't assume `18` and should be using the token's decimals instead.

apps/token-manager/app/src/components/VestingInfoBoxes.js
```diff
@@ -0,0 +1,160 @@
...
+
+                > span:nth-child(1) {
+                  color: ${theme.surfaceContentSecondary};
+                }
+                > span:nth-child(2) {
+                  // “:” is here for accessibility reasons, we can hide it
+                  opacity: 0;
+                  width: 10px;
+                }
+                > span:nth-child(3) {
```
sohkai, line 98: It doesn't look like we have a third `span`; should we wrap `content` around with one?

apps/token-manager/app/src/screens/Details.js
```diff
@@ -0,0 +1,193 @@
...
+  useLayout,
+  useTheme,
+} from '@aragon/ui'
+import SummaryBar from '../components/SummaryBar'
+import SummaryRows from '../components/SummaryRows'
+import VestingInfoBoxes from '../components/VestingInfoBoxes'
+import { useAppLogic, toISODate, useVestedTokensInfo } from '../app-logic'
+import { format, formatDistanceStrict, parseISO } from 'date-fns'
+
+const formatDate = date => `${format(date, 'do MMM yyyy, HH:mm O')}`
```
sohkai, line 21: We should move this to a `date-utils` or `utils` file :).

Also, I think we should discuss the date format. Elsewhere, we've settled on a different one (https://github.com/aragon/aragon-apps/pull/1103#discussion_r406679045).

apps/token-manager/app/src/screens/Details.js
```diff
@@ -0,0 +1,193 @@
...
+
+function Details({ tokenSymbol, selectHolder, vestings }) {
+  const theme = useTheme()
+  const { selectedHolder } = useAppLogic()
```
sohkai, line 25: If we're already using this, perhaps we should just get `selectHolder` here instead of having to pass it down :).

apps/token-manager/app/src/screens/Details.js
```diff
@@ -0,0 +1,193 @@
...
+  const handleBack = useCallback(() => selectHolder(-1), [selectHolder])
```
sohkai, line 26: Note that here we're doing an implicit cast from `-1` to `'-1'`, and we may want to use a different mechanism for this.

Regardless of whether we change the unselected state to be `null` or a falsey value, we may want to expose a `unselectHolder` function from `useAppLogic()`.

apps/token-manager/app/src/screens/Details.js
```diff
@@ -0,0 +1,193 @@
...
+  return (
+    <React.Fragment>
+      <Bar>
+        <BackButton onClick={handleBack} />
+      </Bar>
+      <Split
+        primary={
+          <Box>
```
sohkai, line 34: From the designs, it looks like we should also have an empty state here?

apps/token-manager/app/src/screens/Details.js
```diff
@@ -0,0 +1,193 @@
...
+      </div>
+      <div
+        css={`
+          flex-grow: 1;
+        `}
+      >
+        <SummaryBar
+          vestingInfo={vestingInfo}
+          css={`
+            margin-top: 0;
```
sohkai, line 92: Wondering if we can remove this

apps/token-manager/app/src/screens/Details.js
```diff
@@ -0,0 +1,193 @@
...
+      `}
+    >
+      <div>
+        <ContentBox>
+          <label
+            css={`
+              color: ${theme.surfaceContentSecondary};
+            `}
+          >
+            START DAY
```
sohkai, line 121: Let's use normal case and use `text-transform: uppercase` to style :).

apps/token-manager/app/src/screens/Details.js
```diff
@@ -0,0 +1,193 @@
...
+              symbol: tokenSymbol,
+            })}
+          </p>
+        </ContentBox>
+      </div>
+    </div>
+  )
+}
+
+function ContentBox({ props, children }) {
```
sohkai, line 170: Let's inline this in the other component where it's used.

It's especially awkward because this component styles children elements _implicitly_ via its css selectors.

apps/token-manager/app/src/screens/Details.js
```diff
@@ -0,0 +1,193 @@
...
+      css={`
+        padding: ${1 * GU}px 0;
+        label {
+          color: ${theme.surfaceContentSecondary};
+          ${textStyle('label2')};
+        }
+        p {
+          ${textStyle('body4')};
+        }
+        button {
```
sohkai, line 183: As far as I can tell, there is no button being styled?

apps/token-manager/app/src/app-logic.js
```diff
@@ -0,0 +1,188 @@
...
+    // Make sure we have loaded information about the current app and other installed apps before showing votes
+    if (!(vestings && currentApp && installedApps)) {
+      return [[], []]
+    }
+
+    return vestings
+  }, [vestings])
+}
+
+export function toISODate(seconds) {
```
sohkai, line 81: To be consistent with the other apps, we should be transforming the times in the script.js to their unix time equivalents in ms.

You can find examples in Voting's script, where we use a `marshall()` function to do the transformations.

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -69,18 +81,24 @@ function Holders({
...
-              values.push(formatBalance(balance, tokenDecimalsBase))
+              values.push(formatTokenAmount(balance, tokenDecimals))
             }
 
-            return values
+            const amount = vestings.reduce((total, vesting) => {
+              return total.add(new BN(vesting.amount))
+            }, new BN(0))
+
+            return [...values, formatTokenAmount(amount, tokenDecimals)]
```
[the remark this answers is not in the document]
delfipolito: You are right! I don't know where I got this from. Will change it (Fortunately is an easy change) :)

apps/token-manager/app/src/components/SummaryRows.js
```diff
@@ -0,0 +1,85 @@
+import React from 'react'
+import styled from 'styled-components'
+import { GU, formatTokenAmount, textStyle, useTheme } from '@aragon/ui'
+
+function SummaryRows({ vestingInfo, tokenSymbol }) {
```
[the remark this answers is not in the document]
delfipolito: Yes I think this suggestion is much cleaner. I did the changes, but take a look just in case to see if that was what you had in mind

apps/token-manager/app/src/components/SummaryBar.js
```diff
@@ -0,0 +1,110 @@
+import React from 'react'
```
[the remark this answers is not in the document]
delfipolito: Yes, I am planning on adding it to aragonUI cause it is used in several places, but for the moment I just steal it from voting so is not a blocker for merging this PR

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -6,15 +6,16 @@ import {
...
+  IconInfo,
   IconLabel,
   IconRemove,
   Split,
-  GU,
+  formatTokenAmount,
   useLayout,
   useTheme,
 } from '@aragon/ui'
-import { formatBalance } from '../utils'
```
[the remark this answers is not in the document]
delfipolito: Sure! I will update this in a separate PR :)

apps/token-manager/app/src/components/VestingInfoBoxes.js
```diff
@@ -0,0 +1,160 @@
...
+  return (
+    <React.Fragment>
+      <Box padding={0}>
+        <div
+          css={`
+            padding: ${4 * GU}px;
+            border-bottom: 1px solid ${theme.border};
+          `}
+        >
+          <LocalIdentityBadge
```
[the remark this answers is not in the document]
delfipolito: We talked with Paty and Fiore about this and resolved to also keep it when is the currrently connected account details screen

sohkai:
Last thing blocking this from being merged is https://github.com/aragon/aragon-ui/pull/777.

rashiq-ddc:
Is this feature available in rinkeby version ?

sohkai:
@rashiq-ddc It is available on both mainnet and rinkeby. However, there is no UI for creating vestings so you'll have to interact programmatically with the contract to do so (or use templates like Fundraising that create vestings).

PR #1139, opened 2020-05-04
Agreements: Implement explicit signing

facuspagnuolo:
This PR builds on top of #1133 

The idea is to have an explicit way of signing the agreement and a mandatory flow to resign it every time it changes.


facuspagnuolo:
Closing in favor of https://github.com/aragon/aragon-apps/pull/1140

PR #1143, opened 2020-05-11
Adding the Aave Tokens into the token list

marczeller:
Hello Aragon Team,

I've added the Aave aTokens to this list, that what it's easier for AragonDAO users to interact with their DAO finance app with aTokens.


sohkai:
Hmm, honestly not super sure what to do at this point as MyCrypto (upstream from us) seems to have removed their verified tokens list (see https://github.com/MyCryptoHQ/MyCrypto/pull/3230).

I've asked if they've opened sourced it, but otherwise, we may have to switch to a different resource for token addresses. Would it also make sense to upload these tokens to https://github.com/trustwallet/assets?

apps/finance/app/src/lib/verified-tokens.js
```diff
@@ -21,13 +21,16 @@ const MYCRYPTO_VERIFIED_TOKENS = [
   ['599GTO1', '0x5d9776472483ee2c2b204775547bff6db5a30fed'],
```
sohkai, line 21: @marczeller Could you also copy this file into the Agent app's list: https://github.com/aragon/aragon-apps/blob/master/apps/agent/app/src/lib/verified-tokens.js?

marczeller: of course, [PR done](https://github.com/aragon/aragon-apps/pull/1144). 

May I suggest Kleros Token TCR?, given both project history, they use Aragon for their governance, Aragon uses the TCR for curation, everybody wins.

but I guess this discussion is outside the immediate scope of this PR :joy:

sohkai: Haha yes, I was thinking that too, but do you have the Aave tokens listed on the Token TCR? Anyhow, that was definitely a resolver we'd like to build into https://github.com/aragon/use-token for verification purposes.

marczeller: CLA signed, yeah, I've added all Aave related tokens there. use-token() seems cool!

sohkai:
@marczeller It looks like MyCrypto pulls from [`ethereum-lists/tokens`](https://github.com/ethereum-lists/tokens/tree/master/tokens): https://github.com/MyCryptoHQ/MyCrypto/pull/3230#issuecomment-627470812

If you haven't already, it would also be great to add the Aave tokens to that list—we will most likely periodically update this list from theirs!

apps/finance/app/src/lib/verified-tokens.js
marczeller: If you want to query the tokens of the T2CR, it's pretty easy, you can look at this example: https://codesandbox.io/s/t2cr-and-badges-nyqi6?file=/src/index.js

PR #1148, opened 2020-05-18
Agent: formatTokenAmount() and Rounding Tooltip

ECWireless:
This first commit simply re-adds the reverts @bpierre made in #1121 

Fixes aragon/aragon#506, #756, #1065.

This PR is still a work in progress. @bpierre are there any other issues I could tack onto this PR?

To do:
- [ ] Add rounding tooltip
- [ ] Add formatTokenAmount from UI

### DIFF
--- a/apps/agent/app/src/components/BalanceToken.js
+++ b/apps/agent/app/src/components/BalanceToken.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { textStyle, useTheme } from '@aragon/ui'
+import { GU, Help, textStyle, useTheme } from '@aragon/ui'
 import { useNetwork } from '@aragon/api-react'
 import { tokenIconUrl } from '../lib/icon-utils'
 import { formatTokenAmount } from '../lib/utils'
@@ -13,6 +13,23 @@ function BalanceToken({
 }) {
   const theme = useTheme()
   const network = useNetwork()
+
+  const [integerAmount, fractionalAmount] = formatTokenAmount(
+    amount.toFixed(3)
+  ).split('.')
+
+  const [_, notRoundedFractionalAmount] = formatTokenAmount(
+    amount,
+    false,
+    0,
+    false,
+    {
+      rounding: 18,
+    }
+  ).split('.')
+
+  const amountWasRounded = fractionalAmount !== notRoundedFractionalAmount
+  
   return (
     <React.Fragment>
       <div
@@ -42,9 +59,23 @@ function BalanceToken({
         <div
           css={`
             ${textStyle('title2')}
+            display:flex;
           `}
         >
-          <SplitAmount amount={amount.toFixed(3)} />
+          <SplitAmount integer={integerAmount} fractional={fractionalAmount} />
+          {amountWasRounded && (
+            <div
+              css={`
+                display: flex;
+                align-items: center;
+                margin-left: ${GU}px;
+              `}
+            >
+              <Help hint={'This is an approximation, see the complete amount'}>
+                {amount}
+              </Help>
+            </div>
+          )}
         </div>
         <div
           css={`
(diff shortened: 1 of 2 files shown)


bpierre:
@ECWireless Thank you for this!

> @bpierre are there any other issues I could tack onto this PR?

I think we might want to reproduce all the changes made in #1121, because the codebase of the two apps is really similar:

- Use BN.js values everywhere.
- Move to aragonUI’s `formatTokenAmount()`.
- Add the tooltip when an amount is rounded (what you did already).
- Align the `Balances.js` and `Transfers.js` components between the two apps.

Let me know if you need anything!

ECWireless:
@bpierre @Evalir sorry for the long delay in adding more to this; just finished moving from one side of the US to the other... But I've made a few more additions to the Agent App! This primarily includes just adding formatTokenAmount() from aragon/ui, as well as BN.js, into any appropriate files.

I believe the only other file that needs updated is useDownloadData.js, although I'm having trouble testing out changes since Chrome is blocking all downloads from iFrames. Not sure if there is an easy way around this...

But let me know what you think! I'll keep messing around with useDownloadData.js in the meantime.

sohkai:
> I believe the only other file that needs updated is useDownloadData.js, although I'm having trouble testing out changes since Chrome is blocking all downloads from iFrames. Not sure if there is an easy way around this...

Ah, this is a good catch. We may need to update our `allow` directive to include `allow-downloads` on the iframe.

sohkai:
On the above, this is fixed with: https://github.com/aragon/aragon/pull/1436.

ECWireless:
@bpierre and @Evalir Finishing touches have just been added, including updates to useDownloadData.js and the removal of formatTokenAmount() (and associated functions) from utils. Let me know about any changes that need made!

sohkai:
@ECWireless I'm wondering if we could double check and align the `BalanceToken.js`, `Balances.js`, and `Transfers.js` files from #1121, as we've updated them there!

I know for `BalanceToken` and `Balances`, we should probably be able to just copy the files over!

ECWireless:
@sohkai Ah, got it! I just copied over BalanceToken.js and Balances.js and everything works fine. It looks like Finance's Transfers.js and Agent's Transactions.js can't simply be copied between each other though, so I'll go through it again and try to align it a bit more.

ECWireless:
@sohkai I believe Agent's Transactions.js and Finance's Transfers.js are pretty well aligned now. I double-checked it, and all I found was a repeated 'symbol' (because of adding formatTokenAmount() from the UI).

apps/agent/app/src/components/Transactions.js
```diff
@@ -257,11 +257,9 @@ const Transactions = React.memo(function Transactions({
 
           const { symbol, decimals } = tokenDetails[toChecksumAddress(token)]
           const formattedAmount = formatTokenAmount(
-            amount,
-            isIncoming,
+            isIncoming ? amount : -Math.abs(amount),
```
sohkai, line 260: Oh interesting, do we only have a BN object for `amount` at this moment? Otherwise, we should use `amount.neg()`.

apps/agent/app/src/components/Transactions.js
```diff
@@ -302,11 +300,9 @@ const Transactions = React.memo(function Transactions({
           const isIncoming = Boolean(from)
           const { symbol, decimals } = tokenDetails[toChecksumAddress(token)]
           const formattedAmount = formatTokenAmount(
-            amount,
-            isIncoming,
+            isIncoming ? amount : -Math.abs(amount),
```
sohkai, line 303: Same as above :).

apps/agent/app/src/components/useDownloadData.js
```diff
@@ -26,12 +26,11 @@ async function getDownloadData({ transactions, tokenDetails, resolveAddress }) {
         tokenTransfers.map(async ({ amount, from, to, token }) => {
```
sohkai, line 26: To be safe, let's stringify each of the values we're returning in this `map()` (similar to https://github.com/aragon/aragon-apps/pull/1121/files#diff-c6b357d815fc23c7aea312476aabe3beR49).

PR #1149, opened 2020-05-23
Disputable Voting: Implement smart contracts

facuspagnuolo:
This PR builds on top of #1140 

### Current version - Rinkeby
- Testing DAO [agreement](https://rinkeby.aragon.org/#/agreement)
- Registered under [`disputable-voting.open.aragonpm.eth`](https://rinkeby.etherscan.io/tx/0xf16b22950d1a7c18697d14dd6c80c1673586fea17782fbae496a2c7ee491840e)
  - App ID `0x705b5084c67966bb8e4640b28bab7a1e51e03d209d84e3a04d2a4f7415f93b34`
  - Contract [`0xb7c1938a827a3d1a984ebb7d66b4a8acf10d2eed`](https://rinkeby.etherscan.io/address/0x1C0032515055387f49053d2A20b5A5a5ef365b94)
  - Content [`Qmc8TpeZXmTmsQ8zp6XavorzB3nMfEUuUxGpk2bYDtTFLC`](https://ipfs.io/ipfs/Qmc8TpeZXmTmsQ8zp6XavorzB3nMfEUuUxGpk2bYDtTFLC)


apps/voting/contracts/Voting.sol
```diff
@@ -515,12 +547,69 @@ contract Voting is IForwarder, AragonApp {
...
+        vote_.pauseDuration = getTimestamp64().sub(vote_.pausedAt);
+        emit CancelVote(_voteId);
+    }
+
+    /**
+    * @dev Void an entry
+    * @param _voteId Identification number of the entry to be voided
+    */
+    function _onDisputableVoided(uint256 _voteId) internal {
+        _onDisputableRejected(_voteId);
```
bingen, line 610: Are we sure this is what we want? I wonder if this should be an init param (to decide the behavior, I mean)

apps/voting/contracts/test/mocks/DisputableVotingMock.sol
```diff
@@ -0,0 +1,15 @@
+pragma solidity 0.4.24;
+
+import "../../Voting.sol";
+import "@aragon/test-helpers/contracts/TimeHelpersMock.sol";
+
+
+contract DisputableVotingMock is Voting, TimeHelpersMock {
```
bingen, line 7: Why are we splitting this?

apps/voting/test/contracts/voting_delegation.js
```diff
@@ -439,8 +438,8 @@ contract('Voting delegation', ([_, root, voter, anotherVoter, thirdVoter, repres
       })
 
       context('when the input is not valid', () => {
-        const repeat = (x, y) => [...Array(x).map(() => y)]
-        const voters = repeat(101, voter)
+        const repeat = (x, y) => [...Array(x)].map(() => y)
+        const voters = repeat(71, voter)
```
bingen, line 442: We could define this as a constant somewhere in these tests, or even better, add a getter to the mock contract that returns the constant and fetch it from there.

apps/voting/truffle.js
```diff
@@ -1 +1,6 @@
-module.exports = require('@aragon/truffle-config-v4')
+const TruffleConfig = require('@aragon/truffle-config-v5/truffle-config')
+
+TruffleConfig.compilers.solc.version = '0.4.24'
+TruffleConfig.compilers.solc.settings.optimizer.runs = 1000
```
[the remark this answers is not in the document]
facuspagnuolo: Yup, will see how we can optimize it

apps/voting/contracts/Voting.sol
```diff
@@ -40,9 +41,19 @@ contract Voting is IForwarder, AragonApp {
     string private constant ERROR_INVALID_OVERRULE_WINDOW = "VOTING_INVALID_OVERRULE_WINDOW";
     string private constant ERROR_DELEGATES_EXCEEDS_MAX_LEN = "VOTING_DELEGATES_EXCEEDS_MAX_LEN";
     string private constant ERROR_INVALID_DELEGATES_INPUT_LEN = "VOTING_INVALID_DELEGATES_INPUT_LEN";
+    string private constant ERROR_CANNOT_PAUSE_VOTE = "VOTING_CANNOT_PAUSE_VOTE";
+    string private constant ERROR_VOTE_NOT_PAUSED = "VOTING_VOTE_NOT_PAUSED";
 
     enum VoterState { Absent, Yea, Nay }
 
+    enum DisputableStatus {
+        Invalid,                        // A vote that has not been reported to the Agreement
```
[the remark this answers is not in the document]
facuspagnuolo: This would be the case for an existing vote in case you upgrade a current voting app to this one

apps/voting/contracts/Voting.sol
```diff
@@ -291,7 +300,9 @@ contract Voting is IForwarder, AragonApp {
     * @return True if the given vote can be executed, false otherwise
     */
     function canExecute(uint256 _voteId) public view voteExists(_voteId) returns (bool) {
-        return _canExecute(votes[_voteId]);
+        Vote storage vote_ = votes[_voteId];
+        bool can = _canExecute(vote_);
+        return vote_.disputableStatus == DisputableStatus.Invalid ? can : can && _canProceed(vote_.actionId);
```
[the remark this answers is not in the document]
facuspagnuolo: I decided not to have it in the internal method but to have it in the external one. Note that  `closeAction` is called when a vote is executed. It is the Agreement responsibility to not allow closing actions that are being disputed.

apps/voting/contracts/Voting.sol
```diff
@@ -301,7 +312,9 @@ contract Voting is IForwarder, AragonApp {
     * @return True if the given voter can participate a certain vote, false otherwise
     */
     function canVote(uint256 _voteId, address _voter) public view voteExists(_voteId) returns (bool) {
-        return _canVote(votes[_voteId], _voter);
+        Vote storage vote_ = votes[_voteId];
+        bool can = _canVote(vote_, _voter);
+        return vote_.disputableStatus == DisputableStatus.Invalid ? can : can && _canProceed(vote_.actionId);
```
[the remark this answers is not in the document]
facuspagnuolo: Read the answer posted above. However, for this case, not that `_isVoteOpen` will return false if the vote is paused or cancelled

apps/voting/contracts/Voting.sol
facuspagnuolo: Based on a conversation with @lkngtn that's the case. However, we could parameterize this in the future

apps/agreement/contracts/disputable/IDisputable.sol
```diff
@@ -0,0 +1,38 @@
+/*
+ * SPDX-License-Identitifer:    GPL-3.0-or-later
+ */
+
+pragma solidity 0.4.24;
```
sohkai, line 5: A bit of a nitpick, but usually for these inheritable contracts, we use license as MIT and allow a loose pragma.

apps/voting/contracts/Voting.sol
```diff
@@ -174,16 +192,6 @@ contract Voting is IForwarder, AragonApp {
         emit ChangeRepresentative(msg.sender, _representative);
     }
 
-    /**
-    * @notice Create a new vote about "`_metadata`"
-    * @param _executionScript EVM script to be executed on approval
-    * @param _metadata Vote metadata
-    * @return voteId Id for newly created vote
-    */
-    function newVote(bytes _executionScript, string _metadata) external auth(CREATE_VOTES_ROLE) returns (uint256 voteId) {
```
sohkai, line 183: It looks like we may have removed this due to space requirements, but we will have to be careful about being backwards compatible on frontends/etc.

sohkai, line 183: Also wondering, are we able to repurpose the `string` metadata argument into the Agreement context?

apps/voting/contracts/Voting.sol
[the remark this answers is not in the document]
bingen: Ok

bingen:
Another generic comment I have: we should remove completely the option of “early execution”. With it somebody can create a vote, cast and execute in one transaction, bypassing the protection of the Agreements (for instance for 51% attack).

apps/voting/contracts/Voting.sol
```diff
@@ -291,7 +277,9 @@ contract Voting is IForwarder, AragonApp {
     * @return True if the given vote can be executed, false otherwise
     */
     function canExecute(uint256 _voteId) public view voteExists(_voteId) returns (bool) {
-        return _canExecute(votes[_voteId]);
+        Vote storage vote_ = votes[_voteId];
+        bool can = _canExecute(vote_);
+        return vote_.disputableStatus == DisputableStatus.Invalid ? can : can && _canProceed(vote_.actionId);
```
willjgriff, line 282: I don't fully understand the Agreements app, but what's the reason for checking `_canProceed` here? Does the `_isPausedOrCancelled` function called by `_canExecute` not cover this?

bingen: Yes, that sounds right. Actually see this comment: https://github.com/aragon/aragon-apps/pull/1149#discussion_r430596773

apps/voting/contracts/Voting.sol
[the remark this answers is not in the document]
bingen: So then, as `_canVote` calls `_isVoteOpen`, we don’t even need to call `_canProceed` here either, right?
See this comment: https://github.com/aragon/aragon-apps/pull/1149#discussion_r432418812

apps/voting/contracts/Voting.sol
facuspagnuolo: I think we are, that is indeed how the proposed implementation works

apps/voting/contracts/Voting.sol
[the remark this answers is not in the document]
facuspagnuolo: I'll resolve this conversation and continue it in Will's thread bellow to avoid confusion

apps/voting/contracts/Voting.sol
facuspagnuolo: > what's the reason for checking `_canProceed` here?

The rationale is that some of the things happening in the Agreement app are being cached in the side of the Voting app. This is the case of a vote being challenged (paused) for example. 

Note that if we don't cache this status there is no security issue tho, I think it's better to do it to ensure a better UX. For example, if you try to execute a vote being challenged, therefore you will try to close the Agreement action and it will fail. Having these caches allows us to improve the querying methods, otherwise `canVote` or `canExecute` could return a false positive. 

> Does the `_isPausedOrCancelled` function called by `_canExecute` not cover this?

Potentially, we cannot be 100% sure our cache works totally perfect. We can now, but the Agreement app could be upgraded or whatever. Thus, I think it's a good idea to properly check against the Agreement app if an action can proceed in these query methods.

What do you guys think?

willjgriff: Ah yeah, I understand now. It seems reasonable. However, would like to point out, in the case an agreement action is not ready `vote()` is only restricted by agreements if it can also execute the vote. Where as `canVote()` is restricted by agreements whether it can execute it or not. So I think there could be situations where `canVote()` fails but `vote()` is still executable.

bingen: It sounds good to me, although a little bit inconsistent, as for casting a vote we rely on the local cache only, instead of also going to Agreements to check, right?

bingen: > Ah yeah, I understand now. It seems reasonable. However, would like to point out, in the case an agreement action is not ready `vote()` is only restricted by agreements if it can also execute the vote. Where as `canVote()` is restricted by agreements whether it can execute it or not. So I think there could be situations where `canVote()` fails but `vote()` is still executable.

Oh, I didn’t have Github page refreshed while I was trying to analyze this, so I didn’t see your previous response, but yes, I mean essentially the same (you explained it better actually :wink: )

facuspagnuolo: That totally makes sense, added the `_canProceed` check to the internal methods for now then

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+        // If non of the above conditions are met, it can be executed
+        return true;
+    }
+
+    /**
+    * @dev Internal function to check if a voter can participate on a vote. It assumes the queried vote exists.
+    * @return True if the given voter can participate a certain vote, false otherwise
+    */
+    function _canVote(Vote storage vote_, address _voter) internal view returns (bool) {
+        return _isVoteOpen(vote_) && _canProceedAgreementAction(vote_.actionId) && token.balanceOfAt(_voter, vote_.snapshotBlock) > 0;
```
willjgriff, line 598: On second thoughts this might not work actually. I've looked at Agreements more and afaik `_canProceedAgreementAction(vote_.actionId)` will return false if an Action is voided. But in the case of this app we still want to be able to vote if an Action is voided.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+    function executeVote(uint256 _voteId) external voteExists(_voteId) {
+        _executeVote(_voteId);
+    }
+
+    // Forwarding fns
+
+    /**
+    * @notice Creates a vote to execute the desired action, and casts a support vote if possible
+    * @dev IForwarder interface conformance
+    *      Disputable apps require to be the initial step in the forwarding chain
```
sohkai, line 228:
```suggestion
    *      Disputable apps are required to be the initial step in the forwarding chain
```

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+    * @param _evmScript Start vote with script
```
sohkai, line 229:
```suggestion
    * @param _evmScript EVM script to be executed on approval
```

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+    * @param _voteId Vote identifier
+    * @return Vote agreement action ID
+    * @return Vote paused date
+    * @return Vote paused duration
+    * @return Vote disputable status
+    */
+    function getDisputableInfo(uint256 _voteId)
+        external
+        view
+        returns (
```
sohkai, line 341: This should have the `voteExists` modifier

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+        return _isRepresentativeOf(_voter, _representative);
+    }
+
+    /**
+    * @notice Tells whether vote #`_voteId` is within the overrule period for delegated votes or not
+    * @dev Initialization check is implicitly provided by `voteExists()` as new votes can only be
+    *      created via `newVote(),` which requires initialization
+    * @return True if the requested vote is within the overrule window for delegated votes, false otherwise
+    */
+    function withinOverruleWindow(uint256 _voteId) public view voteExists(_voteId) returns (bool) {
```
sohkai, line 400: I'm wondering if we should just make all the getters `external`; I don't think we'll see much inheritance in practice with the size limit, and there are always the internal methods exposed anyway.

On that note, maybe we should make all the internal ones `private` too, to close this off entirely from extension?

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+
+        Vote storage vote_ = votes[voteId];
+        vote_.startDate = getTimestamp64();
+        vote_.snapshotBlock = snapshotBlock;
+        vote_.overruleWindow = overruleWindow;
+        vote_.supportRequiredPct = supportRequiredPct;
+        vote_.minAcceptQuorumPct = minAcceptQuorumPct;
+        vote_.votingPower = votingPower;
+        vote_.executionScript = _executionScript;
+        vote_.actionId = _newAgreementAction(voteId, msg.sender, bytes(_metadata));
```
sohkai, line 426: I might move this down to its own line and add some comments, since this is going to call an external contract

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+                emit ProxyVoteFailure(_voteId, voter, msg.sender);
+            }
+        }
+    }
+
+    /**
+    * @dev Internal function for a representative to cast a vote on behalf of a voter. It assumes the queried vote exists.
+    */
+    function _voteOnBehalfOf(uint256 _voteId, bool _supports, address _voter, address _representative) internal {
+        Vote storage vote_ = votes[_voteId];
```
sohkai, line 474: It's kind of weird this has a very different signature than the other one.

Another thing I'm thinking about is just taking the `Vote` as a storage variable, as extra SLOADs are so expensive now.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+        _closeAgreementAction(vote_.actionId);
+
+        bytes memory input = new bytes(0); // TODO: Consider input for voting scripts
+        address[] memory blacklist = new address[](1);
+        blacklist[0] = address(_getAgreement());
+        runScript(vote_.executionScript, input, blacklist);
+        emit ExecuteVote(_voteId);
+    }
+
+    /**
```
sohkai, line 506: More of a meta note, but I would recommend putting one of those big headers for these disputable actions (perhaps even before the internal functions), to better group them.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+        vote_.pauseDuration = getTimestamp64().sub(vote_.pausedAt);
+        emit CancelVote(_voteId);
+    }
+
+    /**
+    * @dev Void an entry
+    * @param _voteId Identification number of the entry to be voided
+    */
+    function _onDisputableActionVoided(uint256 _voteId) internal {
+        _onDisputableActionAllowed(_voteId);
```
sohkai, line 550: Haha this is a little bit misleading, especially with the natspec.

Perhaps we can write some comments here to explain why we're choosing `allowed`.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+        if (!_isValuePct(vote_.yea, vote_.votingPower, vote_.minAcceptQuorumPct)) {
+            return false;
+        }
+
+        // If the vote cannot proceed due to an Agreement dispute, it cannot be executed
+        if (!_canProceedAgreementAction(vote_.actionId)) {
+            return false;
+        }
+
+        // If non of the above conditions are met, it can be executed
```
sohkai, line 589:
```suggestion
        // If none of the above conditions are met, it can be executed
```

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+        if (!_isValuePct(vote_.yea, totalVotes, vote_.supportRequiredPct)) {
+            return false;
+        }
+
+        // If the vote has not reached min quorum, it cannot be executed
+        if (!_isValuePct(vote_.yea, vote_.votingPower, vote_.minAcceptQuorumPct)) {
+            return false;
+        }
+
+        // If the vote cannot proceed due to an Agreement dispute, it cannot be executed
```
sohkai, line 584: Let's leave another comment why we want to double check with the Agreement, even though we already have all the local state available at this point to tell.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+    * @return True if the given vote can be executed, false otherwise
+    */
+    function _canExecute(Vote storage vote_) internal view returns (bool) {
+        // If vote is already executed, it cannot be executed again
+        if (vote_.executed) {
+            return false;
+        }
+
+        // If the vote is paused or cancelled, it cannot be executed
+        if (_isPausedOrCancelled(vote_)) {
```
sohkai, line 564: Checking the paused state seems kind of redundant; the only time this check would be important with the other one is when the vote is cancelled, right?

apps/voting-disputable/contracts/DisputableVoting.sol
sohkai, line 598: The `_canProceedAgreementAction()` action may be a bit too expensive to do on every vote and I wonder if we can just check against the local state, e.g. with `_isVoteOpen()`.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+    */
+    function _voteEndDate(Vote storage vote_) internal view returns (uint64) {
+        uint64 endDate = vote_.startDate.add(voteTime);
+        DisputableStatus status = vote_.disputableStatus;
+
+        if (status == DisputableStatus.Cancelled) {
+            return vote_.pausedAt.add(vote_.pauseDuration);
+        }
+
+        return endDate.add(vote_.pauseDuration);
```
sohkai, line 676: We can optimize this a bit :).

But also feels a bit weird to not handle the `Paused` state at all. This is only used with `_withinOverruleWindow()`, so I wonder if it's something we should inline into that to bundle the assumptions together with where that function is applied (e.g. it never makes sense for `_withinOverruleWindow()` to return true in a paused state).

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+    /**
+    * @dev Internal function to cast a vote. It assumes the queried vote exists.
+    */
+    function _vote(uint256 _voteId, bool _supports, address _voter) internal {
+        // This could re-enter, though we can assume the governance token is not malicious
+        Vote storage vote_ = votes[_voteId];
+        _castVote(vote_, _voteId, _supports, _voter);
+        _overwriteCasterIfNecessary(vote_, _voter);
+
+        if (_canExecute(vote_)) {
```
sohkai, line 445: I think we were discussing removing early execution; did we settle on removing it from this version?

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+    }
+
+    /**
+    * @notice Create a new vote about "`_metadata`"
+    * @param _executionScript EVM script to be executed on approval
+    * @param _metadata Vote metadata
+    * @param _castVote Whether to also cast newly created vote
+    * @return voteId id for newly created vote
+    */
+    function newVote(bytes _executionScript, string _metadata, bool _castVote)
```
sohkai, line 180: Not sure if we want to support the `castVote` mechanism here, or if `forward()` should do the same.

I think we shouldn't (especially on forward, as that was only meant for early execution).

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+
+    // Forwarding fns
+
+    /**
+    * @notice Creates a vote to execute the desired action, and casts a support vote if possible
+    * @dev IForwarder interface conformance
+    *      Disputable apps require to be the initial step in the forwarding chain
+    * @param _evmScript Start vote with script
+    */
+    function forward(bytes _evmScript) public {
```
sohkai, line 231: I assume we'll be updating this in the future with the bytes?

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+ * SPDX-License-Identifier:    GPL-3.0-or-later
+ */
+
+pragma solidity 0.4.24;
+
+import "@aragon/os/contracts/apps/AragonApp.sol";
+import "@aragon/os/contracts/common/IForwarder.sol";
+import "@aragon/os/contracts/lib/math/SafeMath.sol";
+import "@aragon/os/contracts/lib/math/SafeMath64.sol";
+import "@aragon/os/contracts/apps/disputable/DisputableApp.sol";
```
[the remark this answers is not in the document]
facuspagnuolo: I like the idea :)

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,745 @@
...
+
+    modifier voteExists(uint256 _voteId) {
+        require(_voteExists(_voteId), ERROR_NO_VOTE);
+        _;
+    }
+
+    /**
+    * @notice Initialize Voting app with `_token.symbol(): string` for governance, minimum support of `@formatPct(_supportRequiredPct)`%, minimum acceptance quorum of `@formatPct(_minAcceptQuorumPct)`%, and a voting duration of `@transformTime(_voteTime)`
+    * @param _token MiniMeToken Address that will be used as governance token
+    * @param _supportRequiredPct Percentage of yeas in casted votes for a vote to succeed (expressed as a percentage of 10^18; eg. 10^16 = 1%, 10^18 = 100%)
```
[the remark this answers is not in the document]
facuspagnuolo: I swear I always forget about this one 🙈

apps/voting-disputable/.solcover.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+    norpc: true,
+    copyPackages: ['@aragon/os', '@aragon/apps-vault'],
+    skipFiles: [
+        'test',
+        '@aragon/os',
+        '@aragon/apps-vault',
+    ]
```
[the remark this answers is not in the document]
facuspagnuolo: We don't!

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: I agree about the `external` ones, but I wouldn't commit to the `internal`s. Why would you avoid people extending this app?

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Totally!

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: I think we did, but happy to discuss if you think it's a bad idea or if there is another alternative around it

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Yes, when the new forwarding interface comes :)

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: I think it's a matter of whether we want to fully trust the cache or not. This case, compared to executing the vote is different since it is not calling the Agreement (the execute calls `closeAgreementAction`), so we may be allowing someone to vote when it is actually not possible. We can decide to fully trust the cache tho.

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Actually `_isVoteOpen` is already handling this case

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Note that it is also being used in `_isVoteOpen`

facuspagnuolo: `Paused` is being handled when doing `endDate.add(vote_.pauseDuration)`. Note that the pause duration is being set when the vote is resumed from being paused.

facuspagnuolo:
Thanks for the review @sohkai! I addressed almost all your comments, note that I changed a little bit the assumptions we make about the disputable status cache. Added a few lines commenting about this as well.

apps/voting-disputable/contracts/DisputableVoting.sol
sohkai: Two reasons:

1. Extensions are naturally difficult to create and audit, as you have less control over the upstream contract (you're probably better off just copying and updating the original contract)
2. We're most likely not leaving a lot of room for extension anyway, so we should do what we can to make this one cheaper / lighter

But yes, I think between `internal` and `private` solidity doesn't generate different code, so it's probably fine to leave as `internal` (but then the current `private` methods should probably just be made `internal`).

apps/voting-disputable/contracts/DisputableVoting.sol
sohkai: Nope, let's lift it as an assumption / something to confirm.

With disputables, any concept of early execution feels harmful.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -543,13 +470,76 @@ contract DisputableVoting is DisputableApp {
...
+        vote_.snapshotBlock = snapshotBlock;
+        vote_.overruleWindow = overruleWindow;
+        vote_.supportRequiredPct = supportRequiredPct;
+        vote_.minAcceptQuorumPct = minAcceptQuorumPct;
+        vote_.votingPower = votingPower;
+        vote_.executionScript = _executionScript;
+        vote_.disputableStatus = DisputableStatus.Active;
+
+        // Notify the Agreement app tied to the current voting app about the vote created.
+        // This is mandatory to make votes disputable, we then store its reference on the Agreement app.
```
sohkai, line 506:
```suggestion
        // This is mandatory to make the vote disputable, by storing a reference to it on the Agreement app.
```

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -543,13 +470,76 @@ contract DisputableVoting is DisputableApp {
...
+        vote_.actionId = _newAgreementAction(voteId, msg.sender, bytes(_metadata));
+
+        emit StartVote(voteId, msg.sender, _metadata);
+    }
+
+    /**
+    * @dev Internal function to cast a vote. It assumes the queried vote exists.
```
sohkai, line 513: If we have this assumption, I wonder if we should prefix with `unsafe`?

Or perhaps better, is just to use the `Vote storage` variable.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -581,12 +566,10 @@ contract DisputableVoting is DisputableApp {
...
-        // If the vote cannot proceed due to an Agreement dispute, it cannot be executed
-        if (!_canProceedAgreementAction(vote_.actionId)) {
-            return false;
-        }
-
-        // If none of the above conditions are met, it can be executed
+        // If none of the above conditions are met, it can be executed.
+        // We are not checking if the vote can proceed against the Agreement app since it will
+        // be tried to be closed when executed. This means the close call will fail in the Agreement
+        // side if the vote's status cache on the voting app was outdated for some weird reason.
```
sohkai, line 572:
```suggestion
        // At this point, we are not checking if the vote can proceed against the Agreement app.
        // This will be done on the vote's execution, as it tries to close the action with the Agreement app.
        // This means the close call can fail in the Agreement's side if the status cache here was outdated
        // for some weird reason.
```

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -543,13 +470,76 @@ contract DisputableVoting is DisputableApp {
...
+        // This could re-enter, though we can assume the governance token is not malicious
+        Vote storage vote_ = votes[_voteId];
+        _castVote(vote_, _voteId, _supports, _voter);
+        _overwriteCasterIfNecessary(vote_, _voter);
+    }
+
+    /**
+    * @dev Internal function for a representative to cast a vote on behalf of many voters. It assumes the queried vote exists.
+    */
+    function _voteOnBehalfOf(address[] _voters, uint256 _voteId, bool _supports) internal {
```
sohkai, line 525: Still kind of find it weird that this interface isn't

```
_voteOnBehalfOf(uint256 _voteId, bool _supports, address[] _voters)
```

To align it with the previous one :).

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -543,13 +470,76 @@ contract DisputableVoting is DisputableApp {
...
+        vote_.actionId = _newAgreementAction(voteId, msg.sender, bytes(_metadata));
+
+        emit StartVote(voteId, msg.sender, _metadata);
+    }
+
+    /**
+    * @dev Internal function to cast a vote. It assumes the queried vote exists.
+    */
+    function _vote(uint256 _voteId, bool _supports, address _voter) internal {
+        // This could re-enter, though we can assume the governance token is not malicious
```
sohkai, line 516: Not sure if this comment is useful anymore; I think this must've been applied to somewhere else where we were calling the token?

But perhaps better is to just state this at the initialization time—we expect to completely trust the attached token when doing any calls.

apps/voting-disputable/contracts/DisputableVoting.sol
sohkai: Right... so my problem with `endDate.add(vote_.pauseDuration)` is that it doesn't _really_ handled the paused state.

If the vote is paused, I honestly don't know what the end time will be, and it seems like bad practice to report it as the original end time (it most definitely will not be!).

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,734 @@
...
+    * @dev Challenge a vote
+    * @param _voteId Identification number of the vote to be challenged
+    */
+    function _onDisputableActionChallenged(uint256 _voteId, uint256 /* _challengeId */, address /* _challenger */) internal {
+        Vote storage vote_ = votes[_voteId];
+        require(_canPause(vote_), ERROR_CANNOT_PAUSE_VOTE);
+
+        vote_.disputableStatus = DisputableStatus.Paused;
+        vote_.pausedAt = getTimestamp64();
+        emit PauseVote(_voteId);
```
sohkai, line 443: Could we log the `challengeId` as well?

Having that information here will be really useful:

1. If you're syncing old events, and your vote has been challenged multiple times, the `currentChallengeId` will not reflect that old event's challengeId
2. It allows us to potentially skip a roundtrip call to `getAction()` (which causes 1) and directly call `getChallenge()`

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -397,111 +399,36 @@ contract DisputableVoting is DisputableApp {
...
     */
-    function _voteOnBehalfOf(uint256 _voteId, bool _supports, address _voter, address _representative) internal {
-        Vote storage vote_ = votes[_voteId];
-        _castVote(vote_, _voteId, _supports, _voter);
-        vote_.casters[_voter] = _representative;
-
-        emit ProxyVoteSuccess(_voteId, _voter, _representative, _supports);
+    function forward(bytes _evmScript) public {
+        require(canForward(msg.sender, _evmScript), ERROR_CANNOT_FORWARD);
+        _newVote(_evmScript, "");
```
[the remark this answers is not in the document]
facuspagnuolo: Yea, we are not preventing this from happening, it was one of the things we set in the technical document. It is expected to be the first and one only per forwarding chain but if as a developer you manage to accept forwards from another Disputable app, the framework doesn't prevent you from that.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -24,26 +25,38 @@ contract DisputableVoting is DisputableApp {
...
     enum VoterState { Absent, Yea, Nay }
 
+    /**
+    * These statuses are used only for caching purposes to avoid querying the Agreement app every time
+    * we need to check if a vote is being challenged. Note that this is not mandatory, we could check
+    * every time against the Agreement app but in order to save some gas, we are using this cache for
+    * cases when there is not a huge risk in case the cache gets outdated. Although the cache is carefully
+    * updated only within the disputable app callbacks, edge cases like an Agreement app upgrade could occur.
+    */
     enum DisputableStatus {
```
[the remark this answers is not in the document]
facuspagnuolo: Is there any use case you are thinking of? It would be a strong assumption to be honest

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -397,111 +399,36 @@ contract DisputableVoting is DisputableApp {
...
+    * @dev IForwarder interface conformance
+    * @param _sender Address of the account intending to forward an action
+    * @return True if the given address can create votes, false otherwise
     */
-    function _executeVote(uint256 _voteId) internal {
-        require(_canExecute(votes[_voteId]), ERROR_CAN_NOT_EXECUTE);
-        _unsafeExecuteVote(_voteId);
+    function canForward(address _sender, bytes) public view returns (bool) {
+        // Note that `canPerform()` implicitly does an initialization check itself
+        return canPerform(_sender, CREATE_VOTES_ROLE, arr(_sender));
```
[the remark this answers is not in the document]
facuspagnuolo: Currently, we decoupled the disputable base app from `IForwarder`, but we can add another check here.

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: We are not using a storage pointer cause we rely on the vote ID for events :/
I'm a bit hesitant about adding the `unsafe` prefix to half of the internal methods haha

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Let me know if the proposed enhancement looks better!

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,749 @@
...
+    }
+
+    // Disputable getter fns
+
+    /**
+    * @dev Tells whether a vote can be challenged or not
+    * @return True if the given vote can be challenged, false otherwise
+    */
+    function canChallenge(uint256 _voteId) external view returns (bool) {
+        return _voteExists(_voteId) && _isVoteOpen(votes[_voteId]);
```
sohkai, line 250: Wondering if we should lift the `_voteExists()` into the modifier so it reverts instead like the other getters?

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,749 @@
...
+    }
+
+    /**
+    * @dev Tell disputable information related to a vote
+    */
+    function getDisputableAction(uint256 _voteId) external view voteExists(_voteId) returns (uint64 endDate, bool challenged, bool finished) {
+        Vote storage vote_ = votes[_voteId];
+        endDate = _voteEndDate(vote_);
+        challenged = _isPaused(vote_);
+        finished = _isVoteOpen(vote_);
```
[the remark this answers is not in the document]
facuspagnuolo: Totally! However, we removed this for now :)

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,749 @@
...
+        return true;
+    }
+
+    /**
+    * @notice Creates a vote to execute the desired action, and casts a support vote if possible
+    * @dev IForwarder interface conformance
+    *      Disputable apps are required to be the initial step in the forwarding chain
+    * @param _evmScript EVM script to be executed on approval
+    */
+    function forward(bytes _evmScript) public {
```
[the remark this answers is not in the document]
facuspagnuolo: I think we should and I'm actually proposing that in https://github.com/aragon/aragonOS/pull/583

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,749 @@
...
+        return _canVote(vote_, _voter);
+    }
+
+    /**
+    * @notice Tells whether `_representative` can vote on behalf of `_voter` in vote #`_voteId` or not
+    * @dev Initialization check is implicitly provided by `voteExists()` as new votes can only be
+    *      created via `newVote()`, which requires initialization
+    * @return True if the given representative can vote on behalf of the given voter in a certain vote, false otherwise
+    */
+    function canVoteOnBehalfOf(uint256 _voteId, address _voter, address _representative) external view voteExists(_voteId) returns (bool) {
```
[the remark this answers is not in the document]
facuspagnuolo: Actually, before we had both `voteOnBehalfOf` and `voteOnBehalfOfMany` so maybe it wasn't that obvious. But I decided to drop the one for a single address, so yea, let's do it!

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: But we won't be able to emit the events then, or do you mean to receive both a `uint256 voteId` and `Voting storage vote` reference?

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,749 @@
...
+            vote_.yea = vote_.yea.add(voterStake);
+        } else {
+            vote_.nay = vote_.nay.add(voterStake);
+        }
+
+        vote_.voters[_voter] = _supports ? VoterState.Yea : VoterState.Nay;
+        emit CastVote(_voteId, _voter, _supports, voterStake);
+    }
+
+    function _overwriteCasterIfNecessary(Vote storage vote_, address _voter) private {
```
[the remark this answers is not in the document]
facuspagnuolo: I honestly like the abstraction, let's rename it then :)

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,749 @@
...
+    function _hasNotVotedYet(Vote storage vote_, address _voter) internal view returns (bool) {
+        return _voteCaster(vote_, _voter) != _voter;
+    }
+
+    /**
+    * @dev Internal function to check if a vote is still open. It assumes the queried vote exists.
+    * @return True if the given vote is open, false otherwise
+    */
+    function _isVoteOpen(Vote storage vote_) internal view returns (bool) {
+        if (_isPausedOrCancelled(vote_)) {
```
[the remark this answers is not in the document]
facuspagnuolo: Definitely, I think I mess it up when carrying these changes from the initial PoC when we tried to build an upgrade of the original version. We should remove that picture from our heads and try to think all this without being biased! (speaking for myself)

facuspagnuolo:
@sohkai thanks for the review, super straightforward! I addressed your comments, let me know your thoughts :)

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,749 @@
...
+
+    /**
+    * @notice Tells whether vote #`_voteId` is within the overrule period for delegated votes or not
+    * @dev Initialization check is implicitly provided by `voteExists()` as new votes can only be
+    *      created via `newVote(),` which requires initialization
+    * @return True if the requested vote is within the overrule window for delegated votes, false otherwise
+    */
+    function withinOverruleWindow(uint256 _voteId) external view voteExists(_voteId) returns (bool) {
+        Vote storage vote_ = votes[_voteId];
+        return _isVoteOpen(vote_) && _withinOverruleWindow(vote_);
```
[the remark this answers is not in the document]
facuspagnuolo: The idea here was to be able to have this in a single function to use it from the frontend unless we expect the frontend to perform both checks individually. What do you think?

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -111,24 +109,27 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
...
     * @param _token MiniMeToken Address that will be used as governance token
     * @param _supportRequiredPct Percentage of yeas in cast votes for a vote to succeed (expressed as a percentage of 10^18; eg. 10^16 = 1%, 10^18 = 100%)
     * @param _minAcceptQuorumPct Percentage of yeas in total possible votes for a vote to succeed (expressed as a percentage of 10^18; eg. 10^16 = 1%, 10^18 = 100%)
-    * @param _voteTime Seconds that a vote will be open for tokenholders to vote
+    * @param _voteTime Seconds that a vote will be open for token holders to vote
+    * @param _overruleWindow New overrule window in seconds
     */
-    function initialize(MiniMeToken _token, uint64 _supportRequiredPct, uint64 _minAcceptQuorumPct, uint64 _voteTime) external onlyInit {
+    function initialize(MiniMeToken _token, uint64 _supportRequiredPct, uint64 _minAcceptQuorumPct, uint64 _voteTime, uint64 _overruleWindow)
```
sohkai, line 118: Should we keep the `onlyInit` modifier?

apps/agreement/package.json
```diff
@@ -23,9 +23,7 @@
     "apm:publish:minor": "aragon apm publish minor --files public/ --prepublish-script apm:prepublish",
```
sohkai, line 23: Should we revert these changes, especially if we've rebased the base branches to be on top of `master`?

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: it is already ensured by `initialized()`, do you prefer duplicating it?

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -220,16 +219,14 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
         require(_canExecute(vote_), ERROR_CANNOT_EXECUTE);
 
         vote_.status = VoteStatus.Executed;
+        // TODO: check agreement action is not closed
```
sohkai, line 222: I wonder if we should detect this case from the Agreement—if the `msg.sender` is the action's disputable, we could ignore the `canClose()` check and immediately return if it's already been closed from the Agreement's side?

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -691,49 +695,22 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
...
-
-        return getTimestamp64() < _voteEndDate(vote_) && !_isExecuted(vote_);
+        return _isActive(vote_) && getTimestamp64() < _voteEndDate(vote_);
     }
 
     /**
     * @dev Internal function to check if a vote is within its overrule window
-    *      It assumes the pointer to the vote is valid
+    *      It assumes the pointer to the vote is valid.
+    *      This function doesn't ensure whether the vote is open or not. Note that it is always used along with `_isVoteOpen`
```
sohkai, line 710:
```suggestion
    *      This function doesn't ensure whether the vote is open or not. Note that it must always be used along with `_isVoteOpen()`
```

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -691,49 +695,22 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
...
-    * @dev Internal function to check if a voter has already voted on a certain vote
-    *      It assumes the pointer to the vote is valid
-    * @return True if the given voter has not voted on the requested vote
-    */
-    function _hasNotVotedYet(Vote storage vote_, address _voter) internal view returns (bool) {
-        return _voteCaster(vote_, _voter) != _voter;
-    }
-
     /**
     * @dev Internal function to check if a vote is still open
```
sohkai, line 699:
```suggestion
    * @dev Internal function to check if a vote is still open for voting
```

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,803 @@
...
+
+    /**
+    * @notice Vote `_supports ? 'yes' : 'no'` in vote #`_voteId`
+    * @dev Initialization check is implicitly provided by `voteExists()` as new votes can only be
+    *      created via `newVote()`, which requires initialization
+    * @param _voteId Id for vote
+    * @param _supports Whether voter supports the vote
+    */
+    function vote(uint256 _voteId, bool _supports) external voteExists(_voteId) {
+        require(_canVote(votes[_voteId], msg.sender), ERROR_CANNOT_VOTE);
```
sohkai, line 194: We might be able to move this check into `_vote()` now—it will also save an extra SLOAD :)

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,803 @@
...
+        return _isVoteOpen(votes[_voteId]);
+    }
+
+    /**
+    * @dev Tells whether vote #`_voteId` can be closed
+    * @param _voteId Id for vote
+    * @return True if the given vote can be closed
+    */
+    function canClose(uint256 _voteId) external view voteExists(_voteId) returns (bool) {
+         return !_isVoteOpen(votes[_voteId]);
```
sohkai, line 250: Want to double check this case—right now, this looks like you can close a vote once you've challenged it.

While, I believe, the Agreement does not let you do this from `closeAction()` due to the internal `_canProceed()` check, we may still want to be a bit more explicit here.

With the previous comment about making `closeAction()` potentially a bit smarter when it's called from the disputable app directly, this change shouldn't significantly impact gas in the happy path (executing a passed vote).

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,803 @@
...
+    {
+        Vote storage vote_ = votes[_voteId];
+        actionId = vote_.actionId;
+        pausedAt = vote_.pausedAt;
+        pauseDuration = vote_.pauseDuration;
+        status = vote_.status;
+    }
+
+    /**
+    * @dev Tells whether vote #`_voteId` is open
```
sohkai, line 366:
```suggestion
    * @dev Tells whether vote #`_voteId` is open for voting
```

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -0,0 +1,803 @@
...
+    *      Initialization check is implicitly provided by `voteExists()` as new votes can only be
+    *      created via `newVote()`, which requires initialization
+    * @param _voteId Id for vote
+    * @return True if the vote is still open
+    */
+    function isVoteOpen(uint256 _voteId) external view voteExists(_voteId) returns (bool) {
```
sohkai, line 372: The base voting app was a lot simpler, but with the new paused state here, I wonder if we should be more explicit with this function's name (renaming it to `isVoteOpenForVoting()` or etc.) to differentiate it from an open vote that is paused.

facuspagnuolo: I like that!

apps/voting-disputable/contracts/DisputableVoting.sol
sohkai: 🤦‍♀️ good point, we should remove our rule of placing the modifier here then (not sure if it's in the docs, but it's almost everywhere we use `initialize()`).

apps/voting-disputable/contracts/DisputableVoting.sol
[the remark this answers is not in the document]
sohkai: Right, sounds good to me. I think the frontend will know whether a vote is open before it calls this function, but doesn't hurt to include both conditions here.

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: It is an interesting approach. However, this comment was because the Agreement app must ensure actions are not closed multiple times (the locked collateral can be unlocked only once). So even if we skip the `canClose` check, `closeAction` should fail. 

Ideally, I was thinking to rename `_closeAgreementAction` to `_closeAgreementActionIfNecessary` in the base app or similar. I think you get the approach, simply to call `closeAction` if the action was not closed already.

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Yes, that's true. However, I prefer making this explicit, let's change it to state that votes can only be closed if they are still active and the vote end date has passed.

PR #1152, opened 2020-06-01
Voting: Multiple transaction steps 

delfipolito:
image


apps/voting/app/src/vote-description-utils.js
```diff
@@ -0,0 +1,115 @@
...
+          <LocalIdentityBadge
+            key={index + 1}
+            compact
+            entity={descriptionLine.value}
+          />
+        )
+      }
+      return (
+        <span key={index + 1}>
+          {descriptionLine.value.description || descriptionLine.value}
```
bpierre, line 66: We should check that `descriptionLine.value` is a string first, since it can also be an object (app or permission), making it crash (e.g. with http://localhost:3000/?#/dcl/0xbc1863a593aaebb40b4f43665de30174c9d3fe29/vote/17/).

apps/voting/app/src/components/VoteCard/VoteCard.js
```diff
@@ -89,10 +89,10 @@ function VoteCard({ vote, onOpen }) {
         )}
         {hasConnectedAccountVoted && <VotedIndicator expand={highlighted} />}
       </div>
-      <VoteText
+      <VoteDescription
```
bpierre, line 92: Nice :100:

sohkai:
> About the UI, do we have a reference of how it should look? The spacing seems a bit odd but maybe it’s just me

Agreed! I was wondering if we should use `ul` and `li` elements for it instead of `div`s, so we can apply the left margins to the whole lines.

apps/voting/app/src/vote-description-utils.js
```diff
@@ -0,0 +1,115 @@
...
+        </span>
+      )
+    })
+
+    description.unshift(app)
+  } else {
+    description = step.description ? (
```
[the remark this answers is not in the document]
delfipolito: It is used again here: https://github.com/aragon/aragon-apps/blob/559c6e7302d880a57db8824991de7c22e75eb45b/apps/voting/app/src/vote-description-utils.js#L129

apps/voting/app/src/disputable-app-logic.js
```diff
@@ -0,0 +1,54 @@
+import React, { useCallback, useMemo } from 'react'
```
sohkai, line 1: Ah, we should remove this from this PR :)

apps/voting/app/src/vote-description-utils.js
```diff
@@ -0,0 +1,135 @@
...
+      ? path.map((step, index) =>
+          renderForwardingStep(step, [index + 1], theme)
+        )
+      : ''
+  }, [path, theme])
+  return <React.Fragment>{description}</React.Fragment>
+}
+
+function renderForwardingStep(step, depth, theme) {
+  const app =
```
sohkai, line 18: Perhaps we can leave a TODO note with the app variable to eventually move to using an AppBadge for this component.

apps/voting/app/src/disputable-app-logic.js
delfipolito: haha sorry I commit this by mistake!

PR #1164, opened 2020-06-12
Voting: fix dev server autoreload

bpierre:
<AragonApi /> has moved to index.js and is now the root component. It fixes an issue where the devserver autoreload would prevent the aragonAPI connection to restore properly. The issue seems to only appear when <AragonApi /> is not the root component. I will investigate separately on the side of aragonAPI for React.

It also aligns the Voting app with the other apps in that regard.


andy-hook:
🙌 🙌

Issue #1161, opened 2020-06-10
Voting: update date format


sohkai:
We should migrate all date formats to ISO-8601 (`yyyy-mm-dd`).

ECWireless:
Anyone else working on this? If not, I can try to fix it... As far as I can tell, the only date formatting occurs in the [VoteDetail](https://github.com/aragon/aragon-apps/blob/af250ffb3fe63cb26f8e95ba2f9eb841b397c263/apps/voting/app/src/screens/VoteDetail.js#L328) component. I'll get started on this now, but in the meantime @sohkai did you spot date formatting in any other component?

sohkai:
Done with #1165.

PR #1165, opened 2020-06-13
Voting: migrate date format to ISO-8601

ECWireless:
Fixes issue #1161.

### DIFF
--- a/apps/voting/app/src/screens/VoteDetail.js
+++ b/apps/voting/app/src/screens/VoteDetail.js
@@ -30,7 +30,7 @@ import { getQuorumProgress } from '../vote-utils'
 import { VOTE_NAY, VOTE_YEA } from '../vote-types'
 import { addressesEqual } from '../web3-utils'
 
-const formatDate = date => `${format(date, 'do MMM yy, HH:mm')}`
+const formatDate = date => `${format(date, 'yyyy-MM-dd, HH:mm')}`
 
 const DEFAULT_DESCRIPTION =
   'No additional description has been provided for this proposal.'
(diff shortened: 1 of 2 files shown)


ECWireless:
@sohkai @bpierre I believe the only other date display in the Voting App is in the [Votes.js](https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/screens/Votes.js#L139) component, but I think that involves updating [DateRangePicker.js](https://github.com/aragon/aragon-ui/blob/c2dec5e9e7d7ed81d1d366f23a81d87fe05212b2/src/components/DateRangePicker/DateRangePicker.js#L197) in aragon/ui (?)

sohkai:
> I think that involves updating DateRangePicker.js in aragon/ui (?)

This is controllable via [`DateRangePicker`'s `format` prop](https://github.com/aragon/aragon-ui/tree/master/src/components/DateRangePicker#format).

We should most likely also update aragonUI's default in the future, but that would be a [breaking change](https://github.com/aragon/aragon-ui/issues/781).

ECWireless:
@sohkai  Got it! Here are screenshots of all the changes:

From [VotesDetail.js](https://github.com/aragon/aragon-apps/blob/af250ffb3fe63cb26f8e95ba2f9eb841b397c263/apps/voting/app/src/screens/VoteDetail.js#L34)...

Passed Vote

From [Votes.js](https://github.com/aragon/aragon-apps/blob/master/apps/voting/app/src/screens/Votes.js#L139) (DateRangePicker)...

DateRangePicker

This date is displayed during a pending vote, but I'm unsure if it requires reformatting (?) ...

Pending Vote

PR #1166, opened 2020-06-15
Switch out app utilities with AragonUI utilities

### LINKED ISSUE #1156: chore: prefer utilities provided by aragonUI
We currently have some utilities copied from aragonUI in a number of the apps, but we should prefer directly importing them from aragonUI instead (for example, `shortenAddress`).

ECWireless:
Fixes #1156 

This PR is a work in progress.

Utilities to import from AragonUI:
- [x] `shortenAddress()`
- [x] `addressesEqual()`
- [x] `formatTokenAmount()` in place of `formatNumber()`


ECWireless:
@sohkai These are the only duplicated utilities I could find, but let me know if there are others I can add to the list!

Also, I couldn't find any way to install the Survey app (?), so I've skipped replacing `formatAmount()`in that app for these commits.

sohkai:
> Also, I couldn't find any way to install the Survey app (?), so I've skipped replacing formatAmount()in that app for these commits.

No worries, we are removing the Survey app from this repo soon anyway :). Thanks for doing this!

apps/voting/app/src/components/SummaryRows.js
```diff
@@ -1,8 +1,7 @@
 import React from 'react'
```
sohkai, line 1: Ah, sorry for the rework—we are likely to have a conflict here against https://github.com/aragon/aragon-apps/pull/1113.

We'll merge that one first, and then it should be pretty easy to rollback these changes against Voting here to solve any conflicts :).

apps/token-manager/app/src/script.js
```diff
@@ -1,6 +1,6 @@
 import Aragon, { events } from '@aragon/api'
+import { addressesEqual }  from '@aragon/ui'
```
sohkai, line 2: Ah... this may be a problem @bpierre.

This is still run in the context of a browser, but I think we should aim for this script to be runnable in a node / etc. environment. AFAIK nothing else used by the script should cause an issue.

From what I remember, importing `@aragon/ui` may force a browser context?

apps/token-manager/app/src/screens/Holders.js
```diff
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types'
 import BN from 'bn.js'
 import { useConnectedAccount } from '@aragon/api-react'
 import {
+  addressesEqual,
```
sohkai, line 6: For style, let's move these utilities to be at the end of the import list (notice how elsewhere we place things like `textStyle()` and `useTheme()` at the end of the list).

apps/token-manager/app/src/script.js
bpierre: It should be fine, aragonUI is now compatible with non-DOM environments: https://github.com/aragon/aragon-ui/pull/699

But another thing to take care of would be to ensure that the entire library doesn’t get included in script.js. That would require using [`--experimental-scope-hoisting`](https://en.parceljs.org/cli.html#enable-experimental-scope-hoisting/tree-shaking-support), but we were having issues with it IIRC.

What would you guys think of inlining the function in every script.js using it, until we move to Parcel 2 or another bundler with a good support for tree shaking? https://github.com/aragon/aragon-ui/blob/master/src/utils/web3.js#L82-L92

PR #1169, opened 2020-06-16
Agent: update help text

### LINKED ISSUE #1162: Agent: update help text suggesting Frame
We should link to https://help.aragon.org/article/37-agent and also mention that you can use Frame through `aragonCLI` and the in-app console.

ECWireless:
Fixes #1162 

@sohkai ended up leaving the [InstallFrame.js](https://github.com/aragon/aragon-apps/blob/master/apps/agent/app/src/components/InstallFrame.js) component (and [modal](https://github.com/aragon/aragon-apps/blob/master/apps/agent/app/src/components/FrameModal.js)) in for now, but will delete if you don't think either are necessary to keep around.

Agent Help Text


apps/agent/app/src/components/AgentHelp.js
```diff
@@ -0,0 +1,54 @@
...
+                margin-bottom: ${0.5 * GU}px;
+              `}
+            >
+              The Agent app can be used to interact with external contracts.
+            </div>
+            <div>
+              <Link
+                css={`
+                  color: ${theme.indicatorContent};
+                `}
```
Evalir, line 44: This color doesn't exist in aragonUI, so this is why it shows up as pink in the example. :) You probably meant [`tagIndicatorContent`](https://ui.aragon.org/colors/). That said though, why are we switching colors?

ECWireless: To be completely honest, I just copied what appeared in the [Agent modal](https://github.com/aragon/aragon-apps/blob/master/apps/agent/app/src/components/FrameModal.js#L37), so I have no reason for switching colors other than that lol

Evalir: Ah haha I see! That component was made long ago, and the theme colors have changed a lot. :) Let's revert the color change and leave the standard `Link` color, and we can also safely remove this modal, as the "Learn More" will now be a external link, rendering it unused.

sohkai:
> Ended up leaving the InstallFrame.js component (and modal) in for now

Let's remove them with this change as well :)

apps/agent/app/src/components/AgentSvg.js
```diff
@@ -0,0 +1,229 @@
+import React from 'react';
```
sohkai, line 1: Let's optimize this into a single asset, perhaps @Evalir you can help @ECWireless do this?

ECWireless: I merely ran the SVG through [svg-to-react-cli](https://www.npmjs.com/package/svg-to-react-cli) originally, but I believe it's optimized in my latest commit (?) @Evalir definitely wouldn't mind help if this is wrong though!

Evalir: Yup, it is better now. :)

apps/agent/app/src/components/AgentHelp.js
```diff
@@ -25,19 +21,24 @@ function InstallFrame() {
             align-items: center;
           `}
         >
-          <div>
-            <FrameSvg />
+          <div
+            css={`
+              width: 46px;
+              height: 46px;
```
Evalir, line 27: As a rule we avoid setting sizes with numbers, and always use the `GU` import with corresponds to 8 pixels, like this: `width: ${6 * GU}px`. This way we're more consistent. Let's GUify this!

ECWireless: Ah, I see. Done!

PR #1171, opened 2020-06-17
Tokens: Move renderEntry hook calls into its own component & fix crash

### LINKED ISSUE #1170: Tokens: App crashes when switching to the last page of DataView
If I'm in the Token Manager app from an org that has more than 10 token holders, such as the [Saint Fame DAO](https://mainnet.aragon.org/#/sf/0xd5ef0004c4589e5e53f3e6e5838b363796ea186a/), the App will crash due to a DataView hooks error.

**How to reproduce**
- Go to an org with more than 10 token holders
- Open the Token Manager App
- Go to the last page of the DataView; The following error will be thrown:

Screen Shot 2020-06-16 at 11 54 35 PM

Evalir:
Moves renderEntry hook calls and & UI depending on these into its own component, to avoid breaking the rules of hooks. Fixes #1170 

Right now there's a bug that's breaking Token Manager apps with more than 10 holders (the pagination threshold) live on Mainnet; when switching to the last tab, a React error will appear, saying that a different amount of hooks were rendered, compared to the previous render. This breaks the rules of hooks, and subsequently, the app throws.

The reason this happens is actually documented on the [aragonUI docs](https://ui.aragon.org/data-view/) for the `DataView` component: the props marked with `render` should never hold hooks as they're called conditionally. The solution is to move the breaking parts of this function to a component.

### DIFF
--- a/apps/token-manager/app/src/screens/Holders.js
+++ b/apps/token-manager/app/src/screens/Holders.js
@@ -63,9 +63,6 @@ function Holders({
           fields={groupMode ? ['Owner'] : ['Holder', 'Balance']}
           entries={mappedEntries}
           renderEntry={([address, balance, vestings]) => {
-            const theme = useTheme()
-            const { totalLocked } = useTotalVestedTokensInfo(vestings)
-
             const isCurrentUser = addressesEqual(address, connectedAccount)
 
             const values = [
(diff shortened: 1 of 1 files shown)


apps/token-manager/app/src/screens/Holders.js
```diff
@@ -224,4 +207,31 @@ function EntryActions({
   )
 }
 
+function TokenAmount({ balance, tokenDecimals, vestings }) {
```
Evalir, line 210: Moved the things above directly into this component; left it on the same file as it's related to the DataView only, but if preferred, we can move it.

PR #1172, opened 2020-06-17
Fix CI

bingen:


apps/agent/contracts/test/mocks/AgentMock.sol
```diff
@@ -0,0 +1,10 @@
+pragma solidity ^0.4.24;
+
+import "../../Agent.sol";
+
+
+contract AgentMock is Agent {
+    function isValidSignatureBytes32(bytes32 _hash, bytes memory _signature) public view returns (bytes4) {
```
sohkai, line 7: Could we leave a comment on why this is necessary?

future-apps/payroll/test/contracts/Payroll_gas_costs.test.js
```diff
@@ -35,27 +35,30 @@ contract('Payroll gas costs', ([owner, employee, anotherEmployee]) => {
...
 
         console.log('cumulativeGasUsed:', cumulativeGasUsed)
-        assert.isAtMost(cumulativeGasUsed, 337000, 'payout gas cost for a single allowed token should be ~338k')
+        assert.isAtMost(cumulativeGasUsed, gas * 1000, `payout gas cost for a single allowed token should be ~${gas}k`)
       })
     })
 
     context('when there are two allowed token', function () {
-      it('expends ~49k gas in setting a allocation tokens per token', async () => {
+      const gas1 = 49
```
sohkai, line 50: Not a huge fan of these variable names (if we even need them at all)

PR #1176, opened 2020-06-23
Agreement: Integrate transaction fees

### LINKED ISSUE #1186: Agreement newAction payments
With the [last implementation](https://github.com/aragon/aragon-apps/pull/1176) of Agreement app, on new action creation the transaction fees oracle is called to fetch the token and the amount to be paid and then the transfer (token or ETH) is performed:

https://github.com/aragon/aragon-apps/blob/agreements_transaction_fees_integration/apps/agreement/contracts/Agreement.sol#L742

If we want the fees to be proportionally distributed to jurors and the governor share to be paid, the beneficiary should be the `CourtSubscriptions` module of Aragon Court, but for proper accounting [`payFees`](https://github.com/aragon/aragon-court/blob/master/contracts/subscriptions/CourtSubscriptions.sol#L145) should be used in that contract, instead of the regular token (or ETH) transfer. This has some problems:

1- It creates a strong and maybe unnecessary dependency between `Agreement` and `CourtSubscriptions`. The new model for transaction fees on new actions was meant to be different and independent of previous model of subscriptions. At the very least `Agreement` should have the definition of an interface for the beneficiary that includes that `payFees` signature.
2- The function `payFees` on `CourtSubscriptions` also fetches the information of the token and amount. That sounds redundant.

Some possible solutions:

- While we could keep the interfaces separate, in our implementation we could have `CourtSubscriptions` implement the `ITransactionFeesOracle` interface. That would somehow mitigate [2].
- For [1], we could create a small contract wrapper to be the beneficiary and to re-route funds to `CourtSubscriptions`. The problem is that it would involve another “manual” step if the token is neither ETH nor ERC223/ERC777 token with a fallback mechanism, plus the approval overhead.
- Kind of a mix of the two above, the oracle could always be the beneficiary, with a defined endpoint to be called for the payment, and then we could have our implementation to re-direct the funds to `CourtSubscriptions`, but “outside” the interface definition.
- We could try to leverage ERC165 to check if the beneficiary returned by the transaction fees oracle implements `payFees`, and in case not, do the regular transfer. This would soften the dependency explained in [1].

bingen:
Relies on https://github.com/aragon/aragonOS/pull/586
Closes #1186


apps/agreement/test/agreement/agreement_new_action.js
```diff
@@ -39,90 +43,156 @@ contract('Agreement', ([_, owner, submitter, someone]) => {
...
                   })
-
-                  it('emits an event', async () => {
-                    const { receipt, actionId } = await disputable.newAction({ submitter, actionContext, stake, sign })
-                    const logs = decodeEventsOfType(receipt, disputable.abi, AGREEMENT_EVENTS.ACTION_SUBMITTED)
-
-                    assertAmountOfEvents({ logs }, AGREEMENT_EVENTS.ACTION_SUBMITTED, 1)
-                    assertEvent({ logs }, AGREEMENT_EVENTS.ACTION_SUBMITTED, { actionId })
+                }
+                context('when the transaction fees module is set', () => {
```
facuspagnuolo, line 134:
```suggestion
                }
                
                context('when the transaction fees module is set', () => {
```

apps/agreement/test/agreement/agreement_new_action.js
```diff
@@ -39,90 +43,156 @@ contract('Agreement', ([_, owner, submitter, someone]) => {
...
-
-                    const actionData = await disputable.getAction(actionId)
-                    assert.equal(actionData.disputable, disputable.disputable.address, 'disputable does not match')
-                    assert.equal(actionData.submitter, submitter, 'submitter does not match')
-                    assert.equal(actionData.context, actionContext, 'action context does not match')
-                    assert.isFalse(actionData.closed, 'action state does not match')
-                    assertBn(actionData.settingId, currentSettingId, 'setting ID does not match')
-                    assertBn(actionData.disputableActionId, disputableActionId, 'disputable action ID does not match')
-                    assertBn(actionData.collateralRequirementId, currentCollateralId, 'action collateral ID does not match')
+                const newActionFlow = (collateralTokenTransactionFeeAmount) => {
```
facuspagnuolo, line 46: Can we rename this to `transactionFeeAmount` simply?

apps/agreement/test/agreement/agreement_new_action.js
```diff
@@ -39,90 +43,156 @@ contract('Agreement', ([_, owner, submitter, someone]) => {
...
-                    const { canClose, canChallenge, canSettle, canDispute, canClaimSettlement, canRuleDispute } = await disputable.getAllowedPaths(actionId)
-                    assert.isTrue(canClose, 'action cannot be closed')
-                    assert.isTrue(canChallenge, 'action cannot be challenged')
-
-                    assert.isFalse(canSettle, 'action can be settled')
-                    assert.isFalse(canDispute, 'action can be disputed')
-                    assert.isFalse(canRuleDispute, 'action dispute can be ruled')
-                    assert.isFalse(canClaimSettlement, 'action settlement can be claimed')
+                  context('when the transaction fee is not zero', () => {
+                    const feeAmount = bigExp(15, 18)
```
facuspagnuolo, line 140: Let's rename it to `transactionFeeAmount` to make it clearer

apps/agreement/test/agreement/agreement_new_action.js
```diff
@@ -156,6 +226,9 @@ contract('Agreement', ([_, owner, submitter, someone]) => {
 
         context('when the app is unregistered', () => {
           beforeEach('mark as unregistered', async () => {
+            await deployer.arbitrator.setModule(deployer.TRANSACTION_FEES_MODULE, deployer.transactionFeesOracle.address)
+            const { collateralToken } = disputable
+            await deployer.transactionFeesOracle.setFee('0x', collateralToken.address, 0)
```
facuspagnuolo, line 231: Why do you need to set the module here? Isn't it already being set in the deployer?

apps/agreement/contracts/Agreement.sol
```diff
@@ -698,6 +707,28 @@ contract Agreement is IAgreement, AragonApp {
 
     // Internal fns
 
+    function _payTransactionFees(uint256 settingId, IDisputable _disputable, Staking _staking, address _submitter) internal {
+        // Get fees
+        IArbitrator arbitrator = settings[settingId].arbitrator;
+        address transactionFeesOracleAddress = arbitrator.getModule(TRANSACTION_FEES_MODULE);
+        require(transactionFeesOracleAddress != address(0), ERROR_TX_FEES_MODULE_NOT_FOUND);
```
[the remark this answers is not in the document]
bingen: Yes, good point!

apps/agreement/contracts/Agreement.sol
```diff
@@ -698,6 +707,28 @@ contract Agreement is IAgreement, AragonApp {
...
+        ITransactionFeesOracle transactionFeesOracle = ITransactionFeesOracle(transactionFeesOracleAddress);
+        bytes32 appId = _disputable.appId();
+        (ERC20 token, uint256 amount, address beneficiary) = transactionFeesOracle.getFee(appId);
+
+        if (amount == 0) {
+            return;
+        }
+
+        // Pay fees
+        if (token == _staking.token()) {
```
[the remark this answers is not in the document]
bingen: I don’t get this one. `_staking` is coming from the cached staking pool in `requirement`. How is this different to what we do in `_lockBalance` call inside `newAction`?

apps/agreement/test/agreement/agreement_new_action.js
bingen: Ok. I named this way because the param only takes a non-zero value when the transaction fees token is staking collateral token.

facuspagnuolo:
Let's rebase this against master :)

apps/agreement/contracts/Agreement.sol
```diff
@@ -137,9 +140,18 @@ contract Agreement is IAgreement, AragonApp {
...
     * @param _arbitrator Address of the IArbitrator that will be used to resolve disputes
+    * @param _aragonAppFeesCashier Transaction fees oracle to fetch fees for new actions
     * @param _stakingFactory Staking factory to be used for the collateral staking pools
     */
-    function initialize(string _title, bytes _content, IArbitrator _arbitrator, StakingFactory _stakingFactory) external {
+    function initialize(
+        string _title,
+        bytes _content,
+        IArbitrator _arbitrator,
+        IAragonAppFeesCashier _aragonAppFeesCashier,
```
sohkai, line 150: Just wondering, but is there a way to get this off of the `IArbitrator` interface, so there's less chances of getting it wrong?

apps/agreement/contracts/test/mocks/disputable/AragonAppFeesCashierMock.sol
```diff
@@ -0,0 +1,61 @@
+pragma solidity ^0.4.24;
+
+import "@aragon/os/contracts/common/EtherTokenConstant.sol";
+import "@aragon/os/contracts/common/IsContract.sol";
+import "@aragon/os/contracts/lib/arbitration/IAragonAppFeesCashier.sol";
+
+
+contract AragonAppFeesCashierMock is IAragonAppFeesCashier, EtherTokenConstant, IsContract {
+    string private constant ERROR_WRONG_TOKEN = "SFO_WRONG_TOKEN";
```
sohkai, line 9: Just curious but what does `SFO` stand for?

apps/agreement/contracts/Agreement.sol
```diff
@@ -270,12 +291,17 @@ contract Agreement is IAgreement, AragonApp {
...
-        _lockBalance(requirement.staking, _submitter, requirement.actionAmount);
-        // TODO: pay court transaction fees
+        Staking staking = requirement.staking;
+        _lockBalance(staking, _submitter, requirement.actionAmount);
+
+        // Pay court transaction fees
+        IDisputable disputable = IDisputable(msg.sender);
 
         uint256 id = nextActionId++;
+        _payAppFees(currentSettingId, disputable, staking, _submitter, id);
```
sohkai, line 301: Just wondering if we shouldn't do this last?

We may also want to do this after emitting the `ActionSubmitted` event—if someone's looking through the transaction's logs, it may more sense if they see the `ActionSubmitted` even before any fee paid events.

apps/agreement/contracts/Agreement.sol
```diff
@@ -712,6 +744,46 @@ contract Agreement is IAgreement, AragonApp {
...
+
+        // Get staking pool
+        Staking staking;
+        if (token == _staking.token()) {
+            staking = _staking;
+        } else {
+            staking = stakingFactory.getOrCreateInstance(token);
+        }
+
+        // Get amount from staking
```
sohkai, line 778:
```suggestion
        // Pull required fee amount from staking pool
```

apps/agreement/contracts/Agreement.sol
```diff
@@ -712,6 +744,46 @@ contract Agreement is IAgreement, AragonApp {
...
+        _lockBalance(staking, _submitter, amount);
+        _slashBalance(staking, _submitter, address(this), amount);
```
sohkai, line 780: Does staking have a way to approve only a certain amount of funds to a lock manager?

apps/agreement/contracts/Agreement.sol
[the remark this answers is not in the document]
sohkai: Since we need to call `stakingFactory.getOrCreateInstance(token)` (a second call) in the worst case, would it be an optimization to just always call it (so it's always one call)?

apps/agreement/contracts/test/mocks/disputable/AragonAppFeesCashierMock.sol
bingen: “San Francisco International Airport”? ;-P This was carried over from “Subscription Fees Oracle”, will fix it.

apps/agreement/contracts/Agreement.sol
bingen: Yes, there’s the allowance concept for that, so owner allows a maximum amount to be locked by the lock manager.

apps/agreement/contracts/Agreement.sol
bingen: We have been thinking about it and didn’t find a clean solution, that’s why we went this route. The main problem would be upgrading the `Controller` in Aragon Court.

apps/agreement/test/agreement/agreement_initialize.js
```diff
@@ -61,12 +68,13 @@ contract('Agreement', ([_, EOA]) => {
         assertEvent({ logs }, AGREEMENT_EVENTS.SETTING_CHANGED, { settingId: currentSettingId })
       })
 
-      it('initializes the first setting with the given title, content and arbitrator', async () => {
+      it('initializes the first setting with the given title, content, arbitrator and transaction fees oracle', async () => {
```
facuspagnuolo, line 71:
```suggestion
      it('initializes the first setting with the given title, content, arbitrator and app fees cashier', async () => {
```

apps/agreement/test/agreement/agreement_initialize.js
```diff
@@ -61,12 +68,13 @@ contract('Agreement', ([_, EOA]) => {
...
         const setting = await agreement.getSetting(1)
 
         assert.equal(setting.title, title, 'title does not match')
         assert.equal(setting.content, content, 'content does not match')
         assert.equal(setting.arbitrator, arbitrator.address, 'arbitrator does not match')
+        assert.equal(setting.aragonAppFeesCashier, aragonAppFeesCashier.address, 'aragonAppFeesCashier does not match')
```
facuspagnuolo, line 77:
```suggestion
        assert.equal(setting.aragonAppFeesCashier, aragonAppFeesCashier.address, 'aragon app fees cashier does not match')
```

apps/agreement/test/helpers/utils/errors.js
```diff
@@ -26,6 +27,7 @@ const AGREEMENT_ERRORS = {
   ERROR_CHALLENGER_FINISHED_EVIDENCE: 'AGR_CHALLENGER_FINISHED_EVIDENCE',
   ERROR_STAKING_FACTORY_NOT_CONTRACT: 'AGR_STAKING_FACTORY_NOT_CONTRACT',
   ERROR_ARBITRATOR_NOT_CONTRACT: 'AGR_ARBITRATOR_NOT_CONTRACT',
+  ERROR_TX_FEES_ORACLE_NOT_CONTRACT: 'AGR_TX_FEES_ORACLE_NOT_CONTRACT',
```
facuspagnuolo, line 30: Let's rename this one

apps/agreement/test/helpers/utils/deployer.js
```diff
@@ -213,6 +223,14 @@ class AgreementDeployer {
     return collateralToken
   }
 
+  async deployAragonAppFeesCashier() {
+    // transaction fees module
```
facuspagnuolo, line 227:
```suggestion
```

apps/agreement/test/helpers/wrappers/agreement.js
```diff
@@ -275,6 +277,17 @@ class AgreementWrapper {
     return token.approveAndCall(to, amount, EMPTY_DATA, { from })
   }
 
+  async allowManager({ token, owner, amount}) {
```
facuspagnuolo, line 280: Nit: I think we use always `user` instead of `owner` for these helpers

apps/agreement/test/agreement/agreement_new_action.js
```diff
@@ -156,6 +234,7 @@ contract('Agreement', ([_, owner, submitter, someone]) => {
 
         context('when the app is unregistered', () => {
           beforeEach('mark as unregistered', async () => {
+            await disputable.changeSetting({ aragonAppFeesCashierAddress: ZERO_ADDRESS, from: owner })
```
facuspagnuolo, line 237: Do we actually need this one?

apps/agreement/test/agreement/agreement_new_action.js
```diff
@@ -88,42 +138,70 @@ contract('Agreement', ([_, owner, submitter, someone]) => {
...
+
+                      context('when the transaction fee payment doesn’t succeed', () => {
+                        it('reverts', async () => {
+                          await assertRevert(disputable.newAction({ submitter, actionContext, stake, sign }), AGREEMENT_ERRORS.STAKING_NOT_ENOUGH_BALANCE)
+                        })
+                      })
+                    })
+
+                    context('when the transaction fee token is not the staking token', () => {
+                      let token
```
facuspagnuolo, line 174:
```suggestion
                      let token
                      
```

apps/agreement/test/agreement/agreement_new_action.js
```diff
@@ -88,42 +138,70 @@ contract('Agreement', ([_, owner, submitter, someone]) => {
...
+                      context('when the transaction fee payment succeeds', () => {
+                        beforeEach('stake and allow manager for new staking pool for transaction fees', async () => {
+                          await disputable.stake({ token, amount: appFeeAmount, user: submitter })
+                          await disputable.allowManager({ token, owner: submitter, amount: appFeeAmount })
+                        })
+
+                        newActionFlow(bn(0))
+                      })
+
+                      context('when the transaction fee payment doesn’t succeed', () => {
```
facuspagnuolo, line 189:
```suggestion
                      context('when the transaction fee payment does not succeed', () => {
```

apps/agreement/test/agreement/agreement_new_action.js
bingen: Yes, otherwise it tries to pay for fees and it complains about not enough balance.

apps/agreement/contracts/Agreement.sol
```diff
@@ -241,7 +241,8 @@ contract Agreement is IAgreement, AragonApp {
     * @notice Update Agreement to title "`_title`" and content "`_content`", with arbitrator `_arbitrator`
     * @dev Initialization check is implicitly provided by the `auth()` modifier
     * @param _arbitrator Address of the IArbitrator that will be used to resolve disputes
-    * @param _aragonAppFeesCashier Transaction fees oracle to fetch fees for new actions
+    * @param _aragonAppFeesCashier Cashier contract to handle app transaction fees for new actions
+
```
sohkai, line 245: Super nitpick, but it looks like there's an extra newline

apps/agreement/contracts/Agreement.sol
```diff
@@ -307,7 +308,7 @@ contract Agreement is IAgreement, AragonApp {
 
         emit ActionSubmitted(id, msg.sender);
 
-        // Pay court transaction fees
+        // Pay app transaction fees
```
sohkai, line 311: I'm wondering if we can come up with a better comment for this, as "app transaction fees" is a bit vague.

Maybe "action submission fees" is a better concept?

apps/agreement/contracts/Agreement.sol
```diff
@@ -76,6 +78,7 @@ contract Agreement is IAgreement, AragonApp {
         string title;
         bytes content;
         IArbitrator arbitrator;
+        IAragonAppFeesCashier aragonAppFeesCashier;                   // Arbitrator module to fetch fees for new actions from
```
sohkai, line 81:
```suggestion
        IAragonAppFeesCashier aragonAppFeesCashier;                   // Cashier to deposit new action transaction fees (usually linked to the selected arbitrator)
```

apps/agreement/contracts/Agreement.sol
```diff
@@ -712,6 +745,46 @@ contract Agreement is IAgreement, AragonApp {
...
+    * @param _disputable Address of the Disputable app, used to determine fees
+    * @param _staking Staking pool for the ERC20 token to transfer fees from
+    * @param _submitter Address of the user that has submitted the action
+    * @param _actionId Identification number of the action to be paid for
+    */
+    function _payAppFees(uint256 _settingId, IDisputable _disputable, Staking _staking, address _submitter, uint256 _actionId) internal {
+        // Get fees
+        Setting storage setting = _getSetting(_settingId);
+        IAragonAppFeesCashier aragonAppFeesCashier = setting.aragonAppFeesCashier;
+        if (address(aragonAppFeesCashier) == address(0)) {
```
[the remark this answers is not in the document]
bingen: Not a big fan of casting address zero to a contract/interface, because it’s a “fake” casting and could potentially lead to issues, as it won’t actually have any of the assumed methods, but for this comparison it doesn’t seem important.

apps/agreement/contracts/Agreement.sol
```diff
@@ -712,6 +745,46 @@ contract Agreement is IAgreement, AragonApp {
...
+            staking = stakingFactory.getOrCreateInstance(token);
+        }
+
+        // Pull required fee amount from staking pool
+        _lockBalance(staking, _submitter, amount);
+        _slashBalance(staking, _submitter, address(this), amount);
+
+        // Pay fees
+        _approveFor(token, address(aragonAppFeesCashier), amount);
+        aragonAppFeesCashier.payAppFees(appId, abi.encodePacked(_actionId));
```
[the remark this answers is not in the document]
bingen: I think it’s an overkill, we just want to pass an action id. The app id makes total sense to have it separate, like all the other methods.

apps/agreement/contracts/Agreement.sol
```diff
@@ -712,6 +744,40 @@ contract Agreement is IAgreement, AragonApp {
...
     // Internal fns
 
+    /**
+    * @dev Pay transactions fees required for new actions
+    * @param _settingId Identification number of the setting being queried
+    * @param _disputable Address of the Disputable app, used to determine fees
+    * @param _submitter Address of the user that has submitted the action
+    * @param _actionId Identification number of the action to be paid for
+    */
+    function _payAppFees(uint256 _settingId, IDisputable _disputable, address _submitter, uint256 _actionId) internal {
```
sohkai, line 754: I'm wondering if we should take a storage pointer instead, since most of our internal functions now use them directly.

apps/agreement/contracts/Agreement.sol
```diff
@@ -307,7 +307,7 @@ contract Agreement is IAgreement, AragonApp {
         emit ActionSubmitted(id, msg.sender);
 
         // Pay action submission fees
-        _payAppFees(currentSettingId, disputable, _submitter, id);
+        _payAppFees(id, action);
```
sohkai, line 310: Actually I'm not sure this is better! Now we're doing a lot more SLOADs from the `action` storage pointer.

In my comment, I meant to push the settings as a storage pointer—sorry if I was unclear!

PR #1177, opened 2020-06-24
Fix finance USD conversion mechanism

Evalir:
Fixes a problem that arose with the move to using [bn.js everywhere](https://github.com/aragon/aragon-apps/commit/b90ebc47c722dcfa16b74b2fc7a9ec12193adb8f#diff-e842cf2b89a742592b21eeebff9f7086R79), where the conversion from Token Amount to USD was being done incorrectly, as bn.js itself can't handle decimals, and they were being passed onto the function. This adds a few things:

- A function to "shift" numbers a certain amount of decimals safely and counting the amount of places it was shifted (`toDecimal` and `fromDecimal` exist, but required a bit of hacking to make them work for this, so I implemented another version)
- A function to transform the passed token taking into account its decimals and the exchange rate correctly, using just bn.js and not relying on pure Javascript numbers.
- Moves all conversion related things to `conversion-utils.js` to remove bloat from the `Balances.js` file.

Few screenshots from the AAVE Org (where the problem arose due to WBTC breaking the previous implementation):
Screen Shot 2020-06-23 at 9 59 32 PM
Screen Shot 2020-06-23 at 9 59 43 PM

### DIFF
--- a/apps/finance/app/src/lib/conversion-utils.js
+++ b/apps/finance/app/src/lib/conversion-utils.js
@@ -1,0 +1,117 @@
+import { useEffect, useRef, useState } from 'react'
+import BN from 'bn.js'
+
+const CONVERT_API_RETRY_DELAY = 2 * 1000
+const CONVERT_API_RETRY_DELAY_MAX = 60 * 1000
+
+const USD_DECIMALS = new BN('2')
+
+function convertRatesUrl(symbolsQuery) {
+  return `https://min-api.cryptocompare.com/data/price?fsym=USD&tsyms=${symbolsQuery}`
+}
+
+function formatConvertRate(convertRate, decimals) {
+  const [whole = '', dec = ''] = convertRate.split('.')
+  const parsedWhole = whole.replace(/^0*/, '')
+  const parsedDec = dec.replace(/0*$/, '')
+  // parsedWhole could be empty,
+  // so in this case, we wanna remove leading zeros.
+  const fullyParsedDec = parsedWhole ? parsedDec : parsedDec.replace(/^0*/, '')
+
+  // Even if we remove leading zeroes from the decimal
+  // part, we want to count as if we "shifted" them
+  const decimalsToShift = decimals.sub(new BN(parsedDec.length.toString()))
+  // Apart from always considering the USD decimals (2),
+  // if there's the strange case that the above is negative,
+  // we take it as a carry as we know we already shifted to far,
+  // and will compensate by shifting the token amount by this much
+  const carryAmount =
+    decimalsToShift.toNumber() < 0
+      ? decimalsToShift.add(USD_DECIMALS)
+      : USD_DECIMALS
+  // The remaining total amount to shift through bn.js to avoid overflow.
+  const amountToShift = new BN('10').pow(
+    decimalsToShift.toNumber() > 0 ? decimalsToShift : new BN('0')
+  )
+
+  // Finish shifting the whole number through BN.js to avoid overflow,
+  return [
+    new BN(`${parsedWhole}${fullyParsedDec}`).mul(amountToShift),
+    carryAmount,
+  ]
+}
+
+export function getConvertedAmount(amount, convertRate, decimals) {
+  const [formattedConvertRate, carryAmount] = formatConvertRate(
+    convertRate.toString(),
+    decimals
+  )
+
+  // Get the actual precision we need to re-add when calculations are over
+  const precisionTarget = new BN('10').pow(decimals.sub(USD_DECIMALS))
+  const convertedAmount = amount
+    // Shift the amount to take into account the USD decimals
+    // + any leftover
+    .mul(new BN('10').pow(carryAmount))
+    // Actually convert to an USD rate
+    .div(formattedConvertRate)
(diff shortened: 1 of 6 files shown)


apps/finance/app/src/lib/conversion-utils.js
```diff
@@ -0,0 +1,117 @@
...
+  // part, we want to count as if we "shifted" them
+  const decimalsToShift = decimals.sub(new BN(parsedDec.length.toString()))
+  // Apart from always considering the USD decimals (2),
+  // if there's the strange case that the above is negative,
+  // we take it as a carry as we know we already shifted to far,
+  // and will compensate by shifting the token amount by this much
+  const carryAmount =
+    decimalsToShift.toNumber() < 0
+      ? decimalsToShift.add(USD_DECIMALS)
+      : USD_DECIMALS
```
Evalir, line 31: I'll admit I haven't found practical examples, but I'm just making sure! 😃

Evalir:
This sounds good! Wondering on how would we scale up the conversion ratio up to the proper number of decimals without potentially running into overflow issues though? It's really why there's a lot of string manipulation here.

apps/finance/app/src/lib/conversion-utils.js
```diff
@@ -0,0 +1,117 @@
...
+    .div(formattedConvertRate)
+    // Return it to its original precision
+    // Note that we don't have to subtract the "extra carry"
+    // as it's undone during the division
+    .mul(precisionTarget)
+
+  return convertedAmount
+}
+
+export function useConvertRates(symbols) {
```
sohkai, line 66: Nitpick, but maybe we should move this hook to the `components` folder as it can only be used in a React context (all the other `/lib` utilities are pure in context).

apps/finance/app/src/lib/conversion-utils.js
```diff
@@ -0,0 +1,117 @@
...
+  const convertedAmount = amount
+    // Shift the amount to take into account the USD decimals
+    // + any leftover
+    .mul(new BN('10').pow(carryAmount))
+    // Actually convert to an USD rate
+    .div(formattedConvertRate)
+    // Return it to its original precision
+    // Note that we don't have to subtract the "extra carry"
+    // as it's undone during the division
+    .mul(precisionTarget)
```
sohkai, line 61: Not sure if we need to consider any USD decimals here, as our "USD" converted amount is still in the original token's decimals after the conversion.

Wondering if we can simplify everything down to:

```
const [whole = '', dec = ''] = convertRate.split('.')
const parsedDec = dec.replace(/0*$/, '')
const carryAmount = parsedDec.length
const rate = `${whole}${parsedDec}`.replace(/^0*/, '')

return amount
  .mul(new BN('10').pow(carryAmount))
  .div(new BN(rate))
```

As we don't need to use the decimals at all.

bpierre: I think we could completely cut the string manipulations [by multiplying the two values and relying on `formatTokenAmount()`](https://github.com/aragon/aragon-apps/blob/fix-finance-usd-conversion-no-string-manipulation/apps/finance/app/src/lib/conversion-utils.js#L13-L22). What do you guys think?

PR: https://github.com/aragon/aragon-apps/pull/1180

Evalir: My only worry with the above is overflow issues on many tokens due to the ceiling for safe numbers in Javascript (`Number.MAX_SAFE_INTEGER`) is around `9*10^15`, having 3 decimals less than most tokens, which have 18. This is really the only reason why I considered doing this string manipulation, rather than just shifting the decimals by normal math.

bpierre: Yes, the loss of precision with `Number` is exactly why are using BN.js, and the proposed solution is taking care of the token decimals… but I realize now that `rate` itself, once scaled up, [could also easily become too large](https://github.com/aragon/aragon-apps/pull/1180/files#r445459722) :-(

I am wondering if it would be easier to implement `TokenAmount.convert()` directly? We would use a mix of string splitting for the rate, and `BigInt` calculations otherwise.

Evalir: > could also easily become too large :-(

Yeah, and the solutions are not pretty :( In general these conversions are a headache when maintaining them in BigInts.

> I am wondering if it would be easier to implement TokenAmount.convert() directly? We would use a mix of string splitting for the rate, and BigInt calculations otherwise.

I think it could be good! it's also pretty high leverage as doing these conversions is both hard to get 100% right, and at the same time extremely common to do.

bpierre: I was thinking that we could add it to `TokenAmount` directly and use it from here, but we could merge this first to make sure it gets fixed quick yes.

I added the commit with the Jest configuration.

What would you think of:

1. Perhaps simplify a bit the process, following @sohkai’s comment.
2. Add some tests to make sure everything works fine.
3. Merge it until `TokenAmount.format()` is ready?

Evalir: Sounds good to me! Let's do that.

PR #1178, opened 2020-06-24
Agreements voting app: 1 – New vote detail layout

delfipolito:
image


apps/voting/app/src/components/VoteDetail/VoteInfoBoxes.js
```diff
@@ -0,0 +1,193 @@
...
+  const votesYeaVotersSize = safeDiv(yea, totalVotes)
+  const votesNayVotersSize = safeDiv(nay, totalVotes)
+
+  return (
+    <div
+      css={`
+        margin-top: ${2 * GU}px;
+        display: grid;
+        grid-template-columns: ${compact ? 'auto' : 'auto auto auto'};
+        grid-template-rows: ${compact ? 'auto auto auto' : 'auto'};
```
andy-hook, line 39:
```suggestion
        grid-auto-flow: ${compact ? 'row' : 'column'};
```
This should also work if you prefer :)

apps/voting/app/src/components/VoteDetail/VoteInfoBoxes.js
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  return (
+    <React.Fragment>
+      <VoteStatus vote={vote} />
+      <div
+        css={`
+          margin-top: ${1 * GU}px;
+          display: inline-grid;
+          grid-template-columns: auto auto;
```
andy-hook, line 169:
```suggestion
          grid-auto-flow: column;
```
Also useful here if you want to add more items without having to explicitly define the count with template syntax.

apps/voting/app/src/components/VoteDetail/VoteInfoBoxes.js
```diff
@@ -0,0 +1,193 @@
...
+      {executionTransaction && (
+        <div>
+          <TransactionBadge
+            networkType={network && network.type}
+            transaction={executionTransaction}
+            css={`
+              margin-top: ${1 * GU}px;
+            `}
+          />
+        </div>
```
andy-hook, line 187: The `margin-top` doesn't seem to be being applied to `TransactionBadge`, perhaps because the component doesn't support passing them down yet. In the meantime maybe we could move it to the outer `<div>`

apps/voting/app/src/screens/VoteDetail.js
```diff
@@ -151,139 +179,36 @@ function VoteDetail({ vote, onBack, onVote, onExecute }) {
...
-                onVoteYes={handleVoteYes}
-                vote={vote}
-              />
-            </section>
-          </Box>
-        }
-        secondary={
-          <React.Fragment>
-            <Box heading="Status">
-              <Status vote={vote} />
```
andy-hook, line 210: Now that `VoteInfoBoxes` has replaced this, perhaps we could remove the `<Status>` component code from this view?

delfipolito: Of course! I deleted it  later but it slipped in this PR😅. Thanks! :)

apps/voting/app/src/components/VoteDetail/VoteInfoBoxes.js
delfipolito: Good catch!

apps/voting/app/src/screens/VoteDetail.js
```diff
@@ -80,68 +81,95 @@ function VoteDetail({ vote, onBack, onVote, onExecute }) {
...
+                        align-items: flex-start;
+                      `}
+                    >
+                      <LocalIdentityBadge
+                        connectedAccount={addressesEqual(
+                          creator,
+                          connectedAccount
+                        )}
+                        entity={creator}
+                      />
```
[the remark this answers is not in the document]
delfipolito: Do you think? I wouldn't use <Field> here cause is not an input component or a form. And is consistent with the description component used above. What do you mean with visual consistency?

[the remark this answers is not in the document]
andy-hook: I was thinking that this combination of a title with control/content is very common and having that margin between them be consistent by using the `<Field>` would be good, but looking now the `<Field>` only gives a `<Label>` which isn’t semantically correct in this context. Let’s ignore this one and I’ll think about how we can provide a component for making this easier outside of forms in the future :)

apps/voting/app/src/components/VoteDetail/VoteInfoBoxes.js
```diff
@@ -0,0 +1,192 @@
...
+function VoteInfoBoxes({ vote }) {
+  const theme = useTheme()
+  const { layoutName } = useLayout()
+  const compact = layoutName === 'small'
+  const { numData } = vote
+  const { minAcceptQuorum, supportRequired, yea, nay } = numData
+  const quorumProgress = getQuorumProgress(vote)
+  const totalVotes = yea + nay
+  const votesYeaVotersSize = safeDiv(yea, totalVotes)
+  const votesNayVotersSize = safeDiv(nay, totalVotes)
```
andy-hook, line 31: There's some overlap between what `VoteDetail` needs and what is being destructured here. Just a suggestion but it might be easier to pass these down directly rather than taking the whole `vote` object and doing some of this twice.

apps/voting/app/src/components/VoteDetail/VoteInfoBoxes.js
```diff
@@ -0,0 +1,192 @@
...
+              ${textStyle('body2')};
+            `}
+          >
+            {round(quorumProgress * 100, 2)}%{' '}
+            <span
+              css={`
+                color: ${theme.surfaceContentSecondary};
+              `}
+            >
+              (>{round(minAcceptQuorum * 100, 2)}% needed)
```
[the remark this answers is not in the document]
delfipolito: Yes! Super good solution! But I didn't want to touch this part of the code cause is taken care here in this [open PR](https://github.com/aragon/aragon-apps/pull/1113). Maybe I should improve it anyway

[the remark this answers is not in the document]
andy-hook: That's fair! let's improve this later after things have settled then :)

PR #1179, opened 2020-06-24
Agreements voting app: 2 – Disputable action status component

delfipolito:
image

### DIFF
--- a/apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
+++ b/apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
@@ -51,7 +51,21 @@ function DisputableActionStatus({ vote, connectedAccount }) {
         {new Date().getTime() > vote.disputable.action.endDate ? (
           formatDate(vote.disputable.action.endDate)
         ) : (
-          <Timer end={new Date(vote.disputable.action.endDate)} />
+          <div
+            css={`
+              display: inline-flex;
+            `}
+          >
+            <Timer end={new Date(vote.disputable.action.endDate)} />{' '}
+            <div
+              css={`
+                padding-left: ${1 * GU}px;
+                color: ${theme.contentSecondary};
+              `}
+            >
+              (48h)
+            </div>
+          </div>
         )}
       </Item>
       <Item>
--- a/apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
+++ b/apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
@@ -11,9 +11,16 @@ import {
 } from '@aragon/ui'
 import { format } from 'date-fns'
 import { getAgreement } from '../../agreementsMockData'
-import { hasDispute } from '../../disputable-utils'
 
 const formatDate = date => `${format(date, 'yyyy-MM-dd, HH:mm')}`
+
+function hasDispute(vote) {
+  return (
+    vote.disputable &&
+    vote.disputable.action &&
+    vote.disputable.action.challenge
+  )
+}
 
 function DisputableActionStatus({ vote }) {
   //TODO: get agreement and vote real data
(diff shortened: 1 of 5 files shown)


apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -0,0 +1,116 @@
...
+  const theme = useTheme()
+  const agreement = getAgreement()
+
+  return (
+    <React.Fragment>
+      <Box heading="Disputable Action Status">
+        <Item>
+          <Label>Status</Label>
+          <div>DisputableStatusTag</div>
+        </Item>
```
andy-hook, line 26: I think `<Field>` would work nicely for these and keep us visually consistent, it comes with it's own margin too so should slot straight in :)

apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -0,0 +1,116 @@
...
+            css={`
+              display: flex;
+              align-items: center;
+            `}
+          >
+            {vote.disputable.action.collateral.challengeAmount}{' '}
+            {vote.disputable.action.collateral.collateralToken}
+            <span
+              css={`
+                padding-left: ${1 * GU}px;
```
andy-hook, line 39:
```suggestion
                display: inline-flex;
                padding-left: ${1 * GU}px;
```

Should make this icon center nicely against the text

apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -0,0 +1,116 @@
...
+        <Item>
+          <Label>Agreement</Label>
+          <div
+            css={`
+              ${textStyle('body2')};
+              color: ${theme.link};
+            `}
+          >
+            {agreement.agreementTitle}
+          </div>
```
andy-hook, line 59: Might be a good use case for the `<Link>` component

apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -0,0 +1,116 @@
...
+                padding-left: ${1 * GU}px;
+              `}
+            >
+              <IconLock size="small" />
+            </span>
+          </div>
+        </Item>
+        <Item>
+          <Label>Challenge period</Label>
+          <Timer end={new Date(Date.now() + 48 * 1000 * 60 * 60)} />
```
andy-hook, line 48: Even though this is temp it might still be worth creating this obj outside of the render, or inserting via the mock data.

apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -0,0 +1,116 @@
...
+import { getAgreement } from '../../agreementsMockData'
+import { addressesEqual } from '../../web3-utils'
+
+function DisputableActionStatus({ vote, connectedAccount }) {
+  //TODO: get agreement and vote real data
+  const theme = useTheme()
+  const agreement = getAgreement()
+
+  return (
+    <React.Fragment>
```
andy-hook, line 21: Looks like the `<Box>` is acting as the root node so we can make this even tidier by removing this

apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -0,0 +1,120 @@
...
+              display: inline-flex;
+              padding-left: ${1 * GU}px;
+            `}
+          >
+            <IconLock size="small" />
+          </span>
+        </div>
+      </Item>
+      <Item>
+        <Label>Challenge period</Label>
```
andy-hook, line 50: Just checking, in the designs this can be either "Settlement period" or "Challenge period" depending on the status, do we need to worry about that now?

apps/voting/app/src/screens/VoteDetail.js
```diff
@@ -50,6 +52,10 @@ function VoteDetail({ vote, onBack, onVote, onExecute }) {
     numData,
     voteId,
   } = vote
+
+  // TODO: get real disputable vote info. this is temporary
+  vote.disputable = getDisputableVoteById(vote.voteId)
```
andy-hook, line 57: It's a TODO so ignore if it's not relevant but creating a new object rather than mutating would be preferred:

```const disputableVote = {...vote, disputable: getDisputableVoteById(vote.voteId)}```

Comment (delfipolito) [reply apps/voting/app/src/screens/VoteDetail.js:57]: This is temporary. This function wont exist and the info of vote.disputable will come with the vote all along from the script. So it wont mutate. Is just a temporary fix so I can show diferent mock data on each vote.

File: apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js (hunk 6 above)
Comment (delfipolito) [reply apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js:50]: Yes, I am planning on adding this validation on a later PR :)

File: apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js [hunk 8]
```diff
@@ -0,0 +1,120 @@
...
+  Link,
+  textStyle,
+  Timer,
+  useTheme,
+} from '@aragon/ui'
+import { format } from 'date-fns'
+import { getAgreement } from '../../agreementsMockData'
+import { addressesEqual } from '../../web3-utils'
+
+const formatDate = date => `${format(date, 'yyyy-MM-dd, HH:mm')}`
```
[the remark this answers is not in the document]
delfipolito: Yes we are using ISO-8601

apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -0,0 +1,120 @@
...
+
+function DisputableActionStatus({ vote, connectedAccount }) {
```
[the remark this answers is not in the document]
delfipolito: Sorry, they must have been there from when I split the PR into multiple ones. I will need them later but I will add them when necessary :)

apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -0,0 +1,120 @@
...
+      <Item>
+        <Label>Action collateral locked</Label>
+        <div
+          css={`
+            display: flex;
+            align-items: center;
+          `}
+        >
+          {vote.disputable.action.collateral.challengeAmount}{' '}
+          {vote.disputable.action.collateral.collateralToken}
```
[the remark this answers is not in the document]
delfipolito: I wouldn't use the TokenBadge here cause is an amount of tokens followed by the token symbol. I think the icon here would look weird. What do you think?

apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -0,0 +1,120 @@
...
+        </Link>
+      </Item>
+      {vote.disputable && vote.disputable.action && (
+        <Item>
+          <Label>Dispute</Label>
+          <Link
+            css={`
+              ${textStyle('body2')};
+            `}
+            href={`https://court.aragon.org/disputes/${vote.disputable.action.currentChallengeId}`}
```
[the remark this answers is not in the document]
delfipolito: As far as I know, is not referenced anywhere else. Am I right @dizzypaty? 
Is a good idea if it is used in other places of the app :)

apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -0,0 +1,120 @@
...
+  return (
+    <div
+      css={`
+        margin-bottom: ${3 * GU}px;
+      `}
+    >
+      {children}
+    </div>
+  )
+}
```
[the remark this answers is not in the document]
delfipolito: Good idea! 
Why do you think they should be in a list <ul>?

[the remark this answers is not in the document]
andy-hook: Was just thinking as they are the list of details that make up an action status, but not a big deal! :)

apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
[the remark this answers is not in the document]
andy-hook: I think it's present in the designs, and standardising the way we present token amounts would be a good idea. But up to you, we can always address it later.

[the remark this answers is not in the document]
delfipolito: you re right! i will add it :)

PR #1180, opened 2020-06-25
Finance USD conversion mechanism − no string manipulation

bpierre:


apps/finance/app/.babelrc
```diff
@@ -5,13 +5,24 @@
       {
         "modules": false,
         "useBuiltIns": "entry",
-        "core-js": 3,
-        "shippedProposals": true,
+        "corejs": 3,
```
sohkai, line 8: 🤦‍♂️ all of our configs with corejs are wrong haha.

Do you mind also fixing this in `aragon/aragon`, `aragon-react-boilerplate`, and `your-first-aragon-app` (I think that's all?)?

apps/finance/app/src/components/Balances.js
```diff
@@ -18,9 +18,10 @@ function useBalanceItems(balances) {
         return {
           address,
           amount,
-          convertedAmount: convertRates[symbol]
-            ? getConvertedAmount(amount, convertRates[symbol], decimals)
-            : new BN('-1'),
+          amountConverted:
+            amount && decimals && convertRates[symbol]
+              ? convertAmount(amount, decimals, 1 / convertRates[symbol])
```
sohkai, line 23: Note that we can also get a `convertRate` that is > 1 here (potentially, many times greater).

apps/finance/app/src/lib/conversion-utils.js
```diff
@@ -1,66 +1,34 @@
...
+ * @param {Object} [options] options passed to formatTokenAmount()
+ * @returns {string} the formatted amount converted
+ */
+export function convertAmount(amount, decimals, rate, options) {
+  amount = new BN(String(amount))
+  decimals = parseInt(String(decimals), 10)
+  return formatTokenAmount(
+    amount
+      .mul(new BN(10).pow(new BN(CONVERT_PRECISION)))
+      .mul(new BN(rate * 10 ** CONVERT_PRECISION)),
```
sohkai, line 28: Hmm, not entirely sure on this; you're going to get an error if `rate * 10 ** CONVERT_PRECISION` becomes an imprecise number.

apps/finance/app/.babelrc
Evalir: Done for [`aragon/aragon`](https://github.com/aragon/aragon/pull/1453), [`aragon-react-boilerplate`](https://github.com/aragon/aragon-react-boilerplate/pull/112) and [`your-first-aragon-app`](https://github.com/aragon/your-first-aragon-app/pull/19).

bpierre:
@sohkai You are right, it wouldn’t work since the rate can be larger than this. Closing this PR, I moved the Jest config back to https://github.com/aragon/aragon-apps/pull/1177.

PR #1181, opened 2020-06-25
Agreements voting app: 3 – Disputable status label

delfipolito:
Disputable status label to use in disputable action status component and on the vote cards on the main screen

image


apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -22,7 +23,7 @@ function DisputableActionStatus({ vote, connectedAccount }) {
       <Box heading="Disputable Action Status">
         <Item>
           <Label>Status</Label>
-          <div>DisputableStatusTag</div>
+          <DisputableStatusLabel vote={vote} />
```
andy-hook, line 26: With the changes to use `Map` below, you could now pass the status symbol to this component directly and guard if necessary:

```
const disputableStatus = vote.disputable && DISPUTABLE_VOTE_STATUS.get(vote.disputable.status)

......

{disputableStatus && <DisputableStatusLabel status={disputableStatus} />}
```

As we only care about the `status` we can avoid passing the entire vote which makes it a bit easier to reason about and keeps all of our concerns clear.

apps/voting/app/src/components/DisputableStatusLabel.js
```diff
@@ -0,0 +1,82 @@
...
+    background: null,
+    label: '',
+    Icon: null,
+    color: null,
+  }
+}
+
+function DisputableStatusLabel({ vote }) {
+  const theme = useTheme()
+  const status = getDisputableVoteStatus(vote)
```
andy-hook, line 64: As mentioned, we could only accept `status` as the prop as it's all we need to determine the style of our `Tag`. Also be sure to add propTypes for the 4 that it accepts.

apps/voting/app/src/components/DisputableStatusLabel.js
```diff
@@ -0,0 +1,82 @@
...
+      Icon: IconClock,
+      color: null,
+    }
+  }
+  return {
+    background: null,
+    label: '',
+    Icon: null,
+    color: null,
+  }
```
andy-hook, line 59: Since we've controlled our input props to be one of the expected 4 values, we could potentially convert this to be an `object` and use the statuses as keys to access these style rules.

apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -28,12 +30,16 @@ function DisputableActionStatus({ vote }) {
...
+    vote.disputable && DISPUTABLE_VOTE_STATUSES.get(vote.disputable.status)
 
   return (
     <Box heading="Disputable Action Status">
       <ul>
         <Item label="Status">
-          <div>DisputableStatusTag</div>
+          {disputableStatus && (
+            <DisputableStatusLabel status={disputableStatus} />
+          )}
```
andy-hook, line 42: Looking good! 🚀

apps/voting/app/src/components/DisputableStatusLabel.js
```diff
@@ -0,0 +1,78 @@
...
+    ],
+    [
+      DISPUTABLE_VOTE_STATUSES.get('Cancelled'),
+      {
+        background: String(theme.surfaceUnder),
+        label: 'Cancelled',
+        Icon: IconClosed,
+        color: String(theme.disabledContent),
+      },
+    ],
```
andy-hook, line 35: We're on the right track! I think we could simplify this even more though! How about we provide a function outside of the component without the string conversions? because we've already filtered the `Symbol`s we can use them as keys directly.

something like:

```
function getAttributes(status, theme) {
  const attributes = {
    [VOTE_STATUS_ACTIVE]: {
      label: 'Scheduled',
      Icon: IconClock,
    },
    [VOTE_STATUS_PAUSED]: {
      background: theme.warningSurface,
      color: theme.warningSurfaceContent,
      label: 'Challenged',
      Icon: IconAttention,
    },
    ....
  }

  return attributes[status]
}

```

apps/voting/app/src/components/DisputableStatusLabel.js
```diff
@@ -0,0 +1,78 @@
...
+      {
+        background: String(theme.warningSurface),
+        label: 'Challenged',
+        Icon: IconAttention,
+        color: String(theme.warningSurfaceContent),
+      },
+    ],
+  ])
+
+  const { Icon, background, color, label } = statusAttributes.get(status)
```
andy-hook, line 56: And here we could call the previous function and pass it the `theme` as well:

`const { Icon, background, color, label } = getAttributes(status, theme)`

apps/voting/app/src/components/DisputableStatusLabel.js
```diff
@@ -0,0 +1,78 @@
...
+  return (
+    <Tag
+      background={background}
+      color={color}
+      mode="indicator"
+      label={label}
+      icon={<Icon size="small" />}
+    />
+  )
+}
```
andy-hook, line 67: Now we can perform our string conversion here if we want:

```
return (
    <Tag
      background={background && `${background}`}
      color={color && `${color}`}
      mode="indicator"
      label={label}
      icon={<Icon size="small" />}
    />
  )
```

apps/voting/app/src/components/DisputableStatusLabel.js
```diff
@@ -0,0 +1,78 @@
...
+
+DisputableStatusLabel.propTypes = {
+  status: PropTypes.oneOf([
+    DISPUTABLE_VOTE_STATUSES.get('Active'),
+    DISPUTABLE_VOTE_STATUSES.get('Cancelled'),
+    DISPUTABLE_VOTE_STATUSES.get('Closed'),
+    DISPUTABLE_VOTE_STATUSES.get('Paused'),
+  ]),
+}
```
andy-hook, line 76: Because we've already filtered the symbols based on what the "api" gave us in `DisptuableActionStatus`, we can just provide the symbols we accept directly:

```
DisputableStatusLabel.propTypes = {
  status: PropTypes.oneOf([
    VOTE_STATUS_ACTIVE,
    VOTE_STATUS_PAUSED,
    VOTE_STATUS_CANCELLED,
    VOTE_STATUS_CLOSED,
  ]),
}
```

PR #1182, opened 2020-06-25
Agreements voting app: 4 – Disputable actions

delfipolito:
All buttons for the avaiable disputable actions depending on the vote status and connected account

image

### DIFF
--- a/apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
+++ b/apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
@@ -36,23 +36,17 @@ function DisputableActionStatus({ vote }) {
 
   return (
     <Box heading="Disputable Action Status">
-      <Item>
-        <Label>Status</Label>
-        <DisputableStatusLabel vote={vote} />
-      </Item>
-      <Item>
-        <Label>Action collateral locked</Label>
-        <div
-          css={`
-            display: flex;
-            align-items: center;
-          `}
-        >
-          {vote.disputable.action.collateral.challengeAmount}{' '}
-          {vote.disputable.action.collateral.collateralToken}
-          <span
+      <ul>
+        <Item label="Status">
+          {disputableStatus && (
+            <DisputableStatusLabel status={disputableStatus} />
+          )}
+        </Item>
+        <Item label="Action collateral locked">
+          <div
             css={`
-              padding-left: ${1 * GU}px;
+              display: flex;
+              align-items: center;
             `}
           >
             {challengeAmount} {collateralToken}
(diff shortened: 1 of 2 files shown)


apps/voting/app/src/components/VoteDetail/DisputableActions.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react'
+import { Button, GU, textStyle, useTheme } from '@aragon/ui'
+import {
+  VOTE_STATUS_PAUSED,
+  VOTE_STATUS_ACTIVE,
+  VOTE_STATUS_CANCELLED,
+  VOTE_STATUS_CLOSED,
+} from '../../disputable-vote-statuses'
+import { getDisputableVoteStatus } from '../../disputable-utils'
```
andy-hook, line 9: I don't think we need this anymore, and since it's been removed it might be causing an import resolution error.

apps/voting/app/src/components/VoteDetail/DisputableActions.js
```diff
@@ -0,0 +1,45 @@
...
+      )
+    }
+  }
+  if (status === VOTE_STATUS_ACTIVE) {
+    if (addressesEqual(submitter, connectedAccount)) {
+      return <Button mode="strong" wide label="Cancel vote" />
+    }
+    return <Button mode="strong" wide label="Challenge vote" />
+  }
+  return <Button mode="strong" wide label="Review details" />
```
andy-hook, line 42: Now this is subjective but for me I found this a little bit hard to follow. My initial thought would be to do something like:

```javascript
const connectedAccountIsSubmitter = addressesEqual(
    submitter,
    connectedAccount
  )

  if (status === VOTE_STATUS_PAUSED && connectedAccountIsSubmitter) {
    return (
      <React.Fragment>
        <Button
          css={`
            margin-bottom: ${2 * GU}px;
          `}
          mode="strong"
          wide
          label="Accept settlement"
        />
        <Button mode="normal" wide label="Raise dispute to court" />
      </React.Fragment>
    )
  }

  if (status === VOTE_STATUS_ACTIVE) {
    return connectedAccountIsSubmitter ? (
      <Button mode="strong" wide label="Cancel vote" />
    ) : (
      <Button mode="strong" wide label="Challenge vote" />
    )
  }

  return <Button mode="strong" wide label="Review details" />
```

But let me know what you think.

apps/voting/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -100,6 +101,14 @@ function DisputableActionStatus({ vote }) {
             trail.
           </Info>
         </Item>
+        <Item>
```
andy-hook, line 104: This `Item` is adding an additional margin on this last element, making the space feel a little larger than it should.

image

We can omit the margin for the last item, either with `:not(:last-child)` or a `Item + Item` selection in the styling (I think the second option is preferred)

PR #1183, opened 2020-06-26
Agent: support ERC-1155

sohkai:
Adds support for [ERC-1155 tokens](https://eips.ethereum.org/EIPS/eip-1155) to the Agent.

Also found a couple other things I've updated with this change:

- When uninitialized, report false for ERC-165, ERC-721, and ERC-1155
- Update docstrings / code style for ERC interfaces to closely match their finalized states on eip.org

### DIFF
--- a/apps/agent/contracts/Agent.sol
+++ b/apps/agent/contracts/Agent.sol
@@ -38,7 +38,7 @@ contract Agent is IERC165, IERC721Receiver, IERC1155Receiver, ERC1271Bytes, IFor
 
     bytes4 private constant ERC165_INTERFACE_ID = 0x01ffc9a7;
     bytes4 private constant ERC721_RECEIVED_INTERFACE_ID = 0x150b7a02; // bytes4(keccak256("onERC721Received(address,address,uint256,bytes)"))
-    bytes4 private constant ERC1155_RECEIVER_INTERFACE_ID= 0x4e2312e0; // IERC721Receiver.onERC1155Received.selector ^ IERC721Receiver.onERC1155BatchReceived.selector
+    bytes4 private constant ERC1155_RECEIVER_INTERFACE_ID = 0x4e2312e0; // IERC721Receiver.onERC1155Received.selector ^ IERC721Receiver.onERC1155BatchReceived.selector
     bytes4 private constant ERC1155_SINGLE_RECEIVED_INTERFACE_ID = 0xf23a6e61; // bytes4(keccak256("onERC1155Received(address,address,uint256,uint256,bytes)"))
     bytes4 private constant ERC1155_BATCH_RECEIVED_INTERFACE_ID = 0xbc197c81; // bytes4(keccak256("onERC1155BatchReceived(address,address,uint256[],uint256[],bytes)"))
 
--- a/apps/agent/contracts/Agent.sol
+++ b/apps/agent/contracts/Agent.sol
@@ -313,7 +313,7 @@ contract Agent is IERC165, IERC721Receiver, ERC1271Bytes, IForwarder, IsContract
     }
 
     function _protectedTokenIndex(address _token) internal view returns (uint256) {
-        for (uint i = 0; i < protectedTokens.length; i++) {
+        for (uint256 i = 0; i < protectedTokens.length; i++) {
             if (protectedTokens[i] == _token) {
               return i;
             }
--- a/apps/agent/contracts/Agent.sol
+++ b/apps/agent/contracts/Agent.sol
@@ -6,15 +6,16 @@ pragma solidity 0.4.24;
 
 import "./SignatureValidator.sol";
 import "./standards/IERC165.sol";
-import "./standards/ERC1271.sol";
 import "./standards/IERC721Receiver.sol";
+import "./standards/IERC1155Receiver.sol";
+import "./standards/ERC1271.sol";
 
 import "@aragon/apps-vault/contracts/Vault.sol";
 
 import "@aragon/os/contracts/common/IForwarder.sol";
 
 
-contract Agent is IERC165, IERC721Receiver, ERC1271Bytes, IForwarder, IsContract, Vault {
+contract Agent is IERC165, IERC721Receiver, IERC1155Receiver, ERC1271Bytes, IForwarder, IsContract, Vault {
     /* Hardcoded constants to save gas
     bytes32 public constant EXECUTE_ROLE = keccak256("EXECUTE_ROLE");
     bytes32 public constant SAFE_EXECUTE_ROLE = keccak256("SAFE_EXECUTE_ROLE");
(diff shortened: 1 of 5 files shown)


apps/agent/contracts/standards/IERC1155Receiver.sol
```diff
@@ -0,0 +1,35 @@
...
+interface IERC1155Receiver {
+    /**
+    * @notice Handle the receipt of a single ERC1155 token type.
+    * @dev An ERC1155-compliant smart contract MUST call this function on the token recipient contract, at the end of a `safeTransferFrom` after the balance has been updated.
+    *      This function MUST return `bytes4(keccak256("onERC1155Received(address,address,uint256,uint256,bytes)"))` (i.e. 0xf23a6e61) if it accepts the transfer.
+    *      This function MUST revert if it rejects the transfer.
+    *      Return of any other value than the prescribed keccak256 generated value MUST result in the transaction being reverted by the caller.
+    * @param _operator The address which initiated the transfer (i.e. msg.sender)
+    * @param _from The address which previously owned the token
+    * @param _id The ID of the token being transferred
```
facuspagnuolo, line 14: In many places, the EIP refers to this field as "token type", not always tho, but I think it is clearer that way

```suggestion
    * @param _id The ID of the token type being transferred
```

apps/agent/contracts/standards/IERC1155Receiver.sol
```diff
@@ -0,0 +1,35 @@
...
+    * @dev An ERC1155-compliant smart contract MUST call this function on the token recipient contract, at the end of a `safeBatchTransferFrom` after the balances have been updated.
+    *      This function MUST return `bytes4(keccak256("onERC1155BatchReceived(address,address,uint256[],uint256[],bytes)"))` (i.e. 0xbc197c81) if it accepts the transfer(s).
+    *      This function MUST revert if it rejects the transfer(s).
+    *      Return of any other value than the prescribed keccak256 generated value MUST result in the transaction being reverted by the caller.
+    * @param _operator The address which initiated the batch transfer (i.e. msg.sender)
+    * @param _from The address which previously owned the token
+    * @param _ids An array containing ids of each token being transferred (order and length must match _values array)
+    * @param _values An array containing amounts of each token being transferred (order and length must match _ids array)
+    * @param _data Additional data with no specified format
+    * @return `bytes4(keccak256("onERC1155BatchReceived(address,address,uint256[],uint256[],bytes)"))`
```
facuspagnuolo, line 32: Same here, I'd add "if it accepts the transfer(s), any other value otherwise"

apps/agent/contracts/standards/IERC1155Receiver.sol
```diff
@@ -0,0 +1,35 @@
...
+    * @dev An ERC1155-compliant smart contract MUST call this function on the token recipient contract, at the end of a `safeTransferFrom` after the balance has been updated.
+    *      This function MUST return `bytes4(keccak256("onERC1155Received(address,address,uint256,uint256,bytes)"))` (i.e. 0xf23a6e61) if it accepts the transfer.
+    *      This function MUST revert if it rejects the transfer.
+    *      Return of any other value than the prescribed keccak256 generated value MUST result in the transaction being reverted by the caller.
+    * @param _operator The address which initiated the transfer (i.e. msg.sender)
+    * @param _from The address which previously owned the token
+    * @param _id The ID of the token being transferred
+    * @param _value The amount of tokens being transferred
+    * @param _data Additional data with no specified format
+    * @return `bytes4(keccak256("onERC1155Received(address,address,uint256,uint256,bytes)"))`
```
facuspagnuolo, line 17: I would add "if it accepts the transfer, any other value otherwise"

apps/agent/contracts/standards/IERC1155Receiver.sol
```diff
@@ -0,0 +1,35 @@
+pragma solidity 0.4.24;
+
+
+// See https://eips.ethereum.org/EIPS/eip-1155
```
facuspagnuolo, line 4:
```suggestion
/**
* @title ERC-1155 Multi Token Receiver
* @dev See https://eips.ethereum.org/EIPS/eip-1155
* @dev Note: The ERC-165 identifier for this interface is `0x4e2312e0`.
*/
```

apps/agent/contracts/standards/IERC1155Receiver.sol
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+    * @notice Handle the receipt of multiple ERC1155 token types.
+    * @dev An ERC1155-compliant smart contract MUST call this function on the token recipient contract, at the end of a `safeBatchTransferFrom` after the balances have been updated.
+    *      This function MUST return `bytes4(keccak256("onERC1155BatchReceived(address,address,uint256[],uint256[],bytes)"))` (i.e. 0xbc197c81) if it accepts the transfer(s).
+    *      This function MUST revert if it rejects the transfer(s).
+    *      Return of any other value than the prescribed keccak256 generated value MUST result in the transaction being reverted by the caller.
+    * @param _operator The address which initiated the batch transfer (i.e. msg.sender)
+    * @param _from The address which previously owned the token
+    * @param _ids An array containing ids of each token being transferred (order and length must match _values array)
```
facuspagnuolo, line 29:
```suggestion
    * @param _ids An array containing ids of each token type being transferred (order and length must match _values array)
```

apps/agent/contracts/Agent.sol
```diff
@@ -37,6 +38,9 @@ contract Agent is IERC165, IERC721Receiver, ERC1271Bytes, IForwarder, IsContract
 
     bytes4 private constant ERC165_INTERFACE_ID = 0x01ffc9a7;
     bytes4 private constant ERC721_RECEIVED_INTERFACE_ID = 0x150b7a02; // bytes4(keccak256("onERC721Received(address,address,uint256,bytes)"))
+    bytes4 private constant ERC1155_RECEIVER_INTERFACE_ID= 0x4e2312e0; // IERC721Receiver.onERC1155Received.selector ^ IERC721Receiver.onERC1155BatchReceived.selector
```
facuspagnuolo, line 41:
```suggestion
    bytes4 private constant ERC1155_RECEIVER_INTERFACE_ID = 0x4e2312e0; // IERC721Receiver.onERC1155Received.selector ^ IERC721Receiver.onERC1155BatchReceived.selector
```

apps/agent/test/agent_shared.js
```diff
@@ -172,6 +194,72 @@ module.exports = (
...
+          assert.isTrue(await agent.supportsInterface(ERC721_RECEIVED_INTERFACE_ID))
+        })
+
+        it('supports ERC1155Receiver interface', async () => {
+          assert.isTrue(await agent.supportsInterface(ERC1155_RECEIVER_INTERFACE_ID))
+        })
+
+        it("doesn't support any other interface", async () => {
+          assert.isFalse(await agent.supportsInterface('0x12345678'))
+          assert.isFalse(await agent.supportsInterface('0x'))
```
facuspagnuolo, line 217: Requested by EIP 165
```suggestion
          assert.isFalse(await agent.supportsInterface('0x'))
          assert.isFalse(await agent.supportsInterface('0xffffffff'))
```

apps/agent/test/agent_shared.js
```diff
@@ -172,6 +194,72 @@ module.exports = (
...
+          const [token, operator, from] = accounts
+          const id = 5
+          const value = 10
+          const data = '0x12345678'
+
+          const callbackReturn = await agent.onERC1155Received.call(operator, from, id, value, data)
+          assert.equal(callbackReturn, ERC1155_SINGLE_RECEIVED_INTERFACE_ID, 'expected ERC1155 single receipt magic return')
+
+          const receipt = await agent.onERC1155Received(operator, from, id, value, data)
+          assertAmountOfEvents(receipt, 'ReceiveERC1155')
```
facuspagnuolo, line 246: We can add a second assert here to check the logged values

apps/agent/test/agent_shared.js
```diff
@@ -172,6 +194,72 @@ module.exports = (
...
+          const [token, operator, from] = accounts
+          const ids = [5, 6]
+          const values = [10, 11]
+          const data = '0x12345678'
+
+          const callbackReturn = await agent.onERC1155BatchReceived.call(operator, from, ids, values, data)
+          assert.equal(callbackReturn, ERC1155_BATCH_RECEIVED_INTERFACE_ID, 'expected ERC1155 batch receipt magic return')
+
+          const receipt = await agent.onERC1155BatchReceived(operator, from, ids, values, data)
+          assertAmountOfEvents(receipt, 'ReceiveERC1155', ids.length)
```
facuspagnuolo, line 259: Same here

apps/agent/contracts/standards/IERC1155Receiver.sol
sohkai: Agreed, but these were the comment strings in the finalized EIP so I'd prefer to leave it as-is :).

apps/agent/test/agent_shared.js
sohkai: Was thinking that... but then stopped because we haven't done this elsewhere 😅.

Agree that we should definitely do this though; do the new contracts include tests for the event parameters?

facuspagnuolo: I think we should, otherwise this is untested logic. In fact, I always do test the event params haha

PR #1185, opened 2020-06-30
Agreements voting app: 5- Reflect disputable status on vote cards

delfipolito:
image

### DIFF
--- a/apps/voting/app/src/components/VoteCard/VoteCard.js
+++ b/apps/voting/app/src/components/VoteCard/VoteCard.js
@@ -84,7 +84,6 @@ function VoteCard({ vote, onOpen }) {
   const disputableStatus =
     vote.disputable && DISPUTABLE_VOTE_STATUSES.get(vote.disputable.status)
 
-  //TODO: update this part of the code once status function is updated
   const border = getCardBorderColor(
     DISPUTABLE_VOTE_STATUSES.get(vote.disputable.status),
     theme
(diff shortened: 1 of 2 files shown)


apps/voting/app/src/components/VoteCard/VoteCard.js
```diff
@@ -56,6 +81,14 @@ function VoteCard({ vote, onOpen }) {
   // “highlighted” means either focused or hovered
   const [highlighted, setHighlighted] = useState(false)
 
+  const disputableStatus =
+    vote.disputable && DISPUTABLE_VOTE_STATUSES.get(vote.disputable.status)
+
+  const border = getCardBorderColor(
+    DISPUTABLE_VOTE_STATUSES.get(vote.disputable.status),
```
andy-hook, line 88: Think we can just pass `disputableStatus`

apps/voting/app/src/components/VoteCard/VoteCard.js
```diff
@@ -56,6 +81,14 @@ function VoteCard({ vote, onOpen }) {
   // “highlighted” means either focused or hovered
   const [highlighted, setHighlighted] = useState(false)
 
+  const disputableStatus =
+    vote.disputable && DISPUTABLE_VOTE_STATUSES.get(vote.disputable.status)
```
andy-hook, line 85: Something to consider when this goes into the disputable app – if the disputable property is always available and with a status, we could remove the guards.

apps/voting/app/src/components/VoteCard/VoteCard.js
```diff
@@ -110,10 +144,16 @@ function VoteCard({ vote, onOpen }) {
...
+          open ? (
+            <Timer end={endDate} maxUnits={4} />
+          ) : (
+            <VoteStatus vote={vote} />
+          )
         ) : (
-          <VoteStatus vote={vote} />
+          disputableStatus && (
+            <DisputableStatusLabel status={disputableStatus} />
+          )
```
andy-hook, line 156: Same thing here, if we always have the status (and it's not being delivered async or something) then no need to check :)

PR #1187, opened 2020-07-02
Disputable Voting: Disallow multiple challenges

facuspagnuolo:


facuspagnuolo:
If the action is disputed, unless the Agreement app or Aragon Court are malicious no, it cannot happen. Aragon Court needs to wait a minimum number of Court terms to provide a ruling. 

However, if the action is settled I think it should be possible to do everything in the same block: create, challenge and settle.

PR #1188, opened 2020-07-03
Agreement: Simplify dispute metadata construction

bingen:
Besides using abi.encodePacked, there’s no need for a semi-colon as
fields are fixed length.


apps/agreement/contracts/Agreement.sol
```diff
@@ -383,7 +380,7 @@ contract Agreement is IAgreement, AragonApp {
         require(msg.sender == submitter, ERROR_SENDER_NOT_ALLOWED);
 
         IArbitrator arbitrator = _getArbitratorFor(action);
-        bytes memory metadata = _buildDisputeMetadata(action);
+        bytes memory metadata = abi.encodePacked(appId(), action.currentChallengeId);
```
sohkai, line 383: I do still like having a `:`, if only to make things a bit more obvious when looking at the raw bytes.

Would we be able to do this with `abi.encodePacked(appId(), ":", action.currentChallengeId)`?

sohkai: Oh nevermind, I see https://github.com/aragon/aragon-apps/pull/1189!

How much worse in gas is it to do avoid the `:`?

facuspagnuolo: I'll re-open the other one, I was inclined due to the gas costs, but actually I like the other one better. Let's check the gas costs diff and decide!

bingen: Yes, that’s #1189

bingen: Do you always remember by heart that `3a` corresponds to the colon? :astonished:

PR #1189, opened 2020-07-03
Agreement: Simplify dispute metadata construction

bingen:
To choose between this one and #1188 (personally I like better that one).


facuspagnuolo:
Closing in favor of https://github.com/aragon/aragon-apps/pull/1188

facuspagnuolo:
Re-opening! Let's analyze the gas costs diff and decide :)

bingen:
Well, gas if we don’t store it I guess it’s not a big deal, but if we did we would waste 20k for just a “:”. But besides that, I find it simpler, cleaner and more consistent with how data is passed around in Ethereum and how we have done it so far (I don’t remember using the colon anywhere else). That extra colon seems completely unnecessary to me. If we need to display this somewhere, the frontend can easily parse it, as fields are fixed length.

bingen:
Another (minor) argument: [here](https://github.com/aragon/court-subgraph/blob/master/helpers/disputable.ts#L14) and [here](https://github.com/aragon/court-subgraph/blob/master/helpers/disputable.ts#L17) we are doing an implicit rounding, as the length is odd.

bingen:
Closed in favor of #1188

Issue #1120, opened 2020-04-09
Agent, finance: Filtering updates & general improvements


Evalir:
Through spring cleaning we have found a few improvements we can implement into Agent's filtering, and some general improvements these two apps can get. Here's the compiled list:

Agent:
- [ ] Filtering: Replace the current way of filtering dates for [individual date filtering](https://github.com/aragon/aragon-apps/pull/1102#discussion_r406286552).
- [ ] Filtering: Move the page reset to 0 to an effect.
- [ ] Filtering: Make transaction types be provided by the filtering.
- [ ] CSV: Use `useToast` 🥂on Agent CSV export.

Finance & Agent:
- [ ] Properly marshall transaction type from the script itself, instead of having to compare types on the frontend.
- [ ] Use LocalIdentity implementation from Voting, which has more utilities.

PR #1193, opened 2020-07-06
Agent: Filtering Updates

Evalir:
Addresses the first 3 points of the pending [Agent](https://github.com/aragon/aragon-apps/issues/1120) upgrades.

### DIFF
--- a/apps/agent/app/src/components/Transactions.js
+++ b/apps/agent/app/src/components/Transactions.js
@@ -26,7 +26,6 @@ import {
 import { useIdentity } from './IdentityManager/IdentityManager'
 import LocalIdentityBadge from './LocalIdentityBadge/LocalIdentityBadge'
 import TransactionFilters from './TransactionFilters'
-import { TRANSACTION_TYPES_LABELS } from '../transaction-types'
 import useDownloadData from './useDownloadData'
 import useFilteredTransactions from './useFilteredTransactions'
 import { ISO_SHORT_FORMAT, ISO_LONG_FORMAT } from '../lib/date-utils'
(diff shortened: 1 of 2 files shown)


Evalir:
@sohkai @bpierre this is ready for review! Can use some feedback as it turned out to be pretty short. :)

apps/agent/app/src/components/useFilteredTransactions.js
```diff
@@ -48,15 +54,17 @@ function useFilteredTransactions({ transactions, tokens }) {
           return false
         }
 
-        // Exclude by date range
+        // filter separately by start and end date
```
andy-hook, line 57:
```suggestion
        // Filter separately by start and end date
```

😅

apps/agent/app/src/components/useFilteredTransactions.js
```diff
@@ -100,6 +109,7 @@ function useFilteredTransactions({ transactions, tokens }) {
     selectedToken,
     selectedTransactionType,
     symbols,
+    transactionTypes: TRANSACTION_TYPES_LABELS,
```
andy-hook, line 112: Is this likely to be altered within this hook in the future? It might be simpler to import it directly in `<TransactionFilters/>`

apps/agent/app/src/components/Transactions.js
```diff
@@ -57,8 +56,13 @@ const Transactions = React.memo(function Transactions({
     selectedToken,
     selectedTransactionType,
     symbols,
+    transactionTypes,
   } = useFilteredTransactions({ transactions, tokens })
 
+  const transactionLabels = useMemo(() => Object.values(transactionTypes), [
+    transactionTypes,
+  ])
```
andy-hook, line 64: Related to previous – If we decide to import `TRANSACTION_TYPES_LABELS` in `<TransactionFilters/>` we could move this there as well.

apps/agent/app/src/components/useFilteredTransactions.js
Evalir: Not quite; the reasoning here is to have a single source of truth regarding the filtering and the types. We could import it from `<TransactionFilters/>`, but it'd be weird to have the types coming from one place, and the logic for these from another.

apps/agent/app/src/components/Transactions.js
Evalir: Yup; definitely not the prettiest thing there, but the reasoning is stated above. :)

PR #1194, opened 2020-07-08
Disputable voting app frontend

delfipolito:
image

The first commit is just the voting app like it is on master.
The second one is the disputable changes from the PRs already approved.
And the other two is the new stuff. The changes on the voting app so it works here.

To try it out: 
1- Include this line ```[appIds['VotingDisputable']]: {
    script: '/script.js',
    start_url: '/index.html',
  },``` on enviroment.js on the client after [this line](https://github.com/aragon/aragon/blob/master/src/environment.js#L65)
2- Run the app
3- Run the client with: ` ARAGON_APP_LOCATOR=0x705b5084c67966bb8e4640b28bab7a1e51e03d209d84e3a04d2a4f7415f93b34 npm run start:rinkeby `


apps/voting-disputable/app/src/components/DetailedDescription.js
```diff
@@ -0,0 +1,130 @@
+import React, { useMemo } from 'react'
```
andy-hook, line 1: Nitpick: We can remove `useMemo` as it's not used

apps/voting-disputable/app/src/components/DisputableStatusLabel.js
```diff
@@ -0,0 +1,74 @@
...
+  textStyle,
+  useTheme,
+} from '@aragon/ui'
+import {
+  VOTE_STATUS_ACTIVE,
+  VOTE_STATUS_CANCELLED,
+  VOTE_STATUS_CLOSED,
+  VOTE_STATUS_PAUSED,
+  DISPUTABLE_VOTE_STATUSES,
+} from '../disputable-vote-statuses'
```
andy-hook, line 19: Nitpick: `GU` `textStyle` `DISPUTABLE_VOTE_STATUSES` can be removed

apps/voting-disputable/app/src/components/DisputableStatusLabel.js
```diff
@@ -0,0 +1,74 @@
...
+
+function getAttributes(status, theme) {
+  const attributes = {
+    [VOTE_STATUS_ACTIVE]: {
+      label: 'Scheduled',
+      Icon: IconClock,
+    },
+    [VOTE_STATUS_CANCELLED]: {
+      background: String(theme.surfaceUnder),
```
andy-hook, line 28: We can remove these string conversions as they are already being performed in the template string when passed to the component.

apps/voting-disputable/app/now.json
```diff
@@ -1,10 +1,14 @@
 {
-  "version": 2,
+  "version": 1,
```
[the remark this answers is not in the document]
delfipolito: Cause the file was copied from the voting app wich is on version 2, but this is the version 1 of the disputable voting app

apps/voting-disputable/app/src/components/DisputableStatusLabel.js
delfipolito: I passed down like this cause theme.surfaceUnder is an extended String object, and on this Tag component is seen as an object instead of a normal string and throughs an error

apps/voting-disputable/app/now.json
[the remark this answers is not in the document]
andy-hook: Gotcha! in that case we should double check the `version` in `package.json` is accurate. afaik the `now.json` config is for the [vercel deployments](https://vercel.com/docs/configuration#project/version) which at a glance mostly run version 2.

apps/voting-disputable/app/src/components/DisputableStatusLabel.js
andy-hook: Interesting! the template string that's being passed to the component at the moment should already be converting this so you shouldn't be seeing any errors by removing them 🤔 .

I did find an error in this file with an icon import though which might be causing the problems! We're importing `IconClosed` which should probably be `IconClose`

apps/voting-disputable/app/src/components/DisputableStatusLabel.js
```diff
@@ -0,0 +1,74 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+import {
+  GU,
+  IconAttention,
+  IconClock,
+  IconClosed,
```
andy-hook, line 6: This one is probably causing some problems! I think it should be `IconClose`

apps/voting-disputable/app/src/components/VoteDetail/VoteInfoBoxes.js
```diff
@@ -0,0 +1,186 @@
...
+          }
+          css={`
+            height: 100%;
+          `}
+        >
+          <SummaryWithPercentages
+            positiveSize={votesYeaVotersSize}
+            requiredSize={supportRequired}
+          />
+        </Box>
```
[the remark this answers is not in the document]
delfipolito: You're right, good idea!

apps/voting-disputable/app/src/components/VoteDetail/DisputableActionStatus.js
```diff
@@ -0,0 +1,145 @@
...
+          )}
+        </Item>
+        <Item label="Action collateral locked">
+          <div
+            css={`
+              display: flex;
+              align-items: center;
+            `}
+          >
+            {challengeAmount} {collateralToken}
```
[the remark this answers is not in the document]
delfipolito: Yes I will add it in a separate PR :)

apps/voting-disputable/app/src/components/DisputableStatusLabel.js
delfipolito: Yes I modified it, thanks :)

PR #1195, opened 2020-07-08
Disputable Voting: Improve tests

facuspagnuolo:
This PR includes:
- Re-organization of the previous existing tests in new files organizing them by features.
- Minor extra test cases and contextualization grouping by scenarios
- A new bunch of helper methods to avoid duplicating logic between these test files


apps/voting-disputable/test/contracts/voting_vote.js
```diff
@@ -0,0 +1,283 @@
...
+            const { nays } = await getVoteState(voting, voteId)
+            const voterState = await voting.getVoterState(voteId, holder29)
+
+            assertBn(nays, bigExp(29, 18), 'nay vote should have been counted')
+            assert.equal(voterState, VOTER_STATE.NAY, 'holder29 should have nay voter status')
+          })
+
+          it('holder can modify vote', async () => {
+            await voting.vote(voteId, true, { from: holder29 })
+            await voting.vote(voteId, false, { from: holder29 })
```
bingen, line 113: It would be nice to make a check here too.

apps/voting-disputable/test/contracts/voting_vote.js
```diff
@@ -0,0 +1,283 @@
...
+      const result1 = await voting.isValuePct(0, 0, pct(50))
+      assert.equal(result1, false, 'total 0 should always return false')
+
+      const result2 = await voting.isValuePct(1, 0, pct(50))
+      assert.equal(result2, false, 'total 0 should always return false')
+    })
+
+    it('tests value = 0', async () => {
+      const result1 = await voting.isValuePct(0, 10, pct(50))
+      assert.equal(result1, false, 'value 0 should false if pct is non-zero')
```
bingen, line 261: “should return”

apps/voting-disputable/test/contracts/voting_execution.js
```diff
@@ -0,0 +1,120 @@
...
+        it('cannot execute vote if not enough quorum met', async () => {
+          await voting.vote(voteId, true, { from: holder20 })
+          await voting.mockIncreaseTime(VOTE_DURATION)
+
+          await assertRevert(voting.executeVote(voteId), VOTING_ERRORS.VOTING_CANNOT_EXECUTE)
+        })
+
+        it('cannot execute vote if not support met', async () => {
+          await voting.vote(voteId, false, { from: holder29 })
+          await voting.vote(voteId, false, { from: holder20 })
```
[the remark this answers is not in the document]
facuspagnuolo: True, but it provides more clarity about not meeting the required support

apps/voting-disputable/test/contracts/voting_settings.js
```diff
@@ -0,0 +1,212 @@
...
+          await deployer.token.generateTokens(holder51, bigExp(51, 18))
+          const { voteId } = await createVote({ voting, from: holder51 })
+          await voting.changeSupportRequiredPct(pct(70), { from })
+
+          // With previous required support at 50%, vote should be approved
+          // with new quorum at 70% it shouldn't have, but since min quorum is snapshotted
+          // it will succeed
+
+          await voting.vote(voteId, true, { from: holder51 })
+          await voting.vote(voteId, true, { from: holder20 })
```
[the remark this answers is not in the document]
facuspagnuolo: True. Funny fact that I only organized these tests, they've been like this for a long time 😂

facuspagnuolo:
Thanks @bingen for reviewing! I addressed all your comments :)

apps/voting-disputable/test/helpers/voting.js
```diff
@@ -1,14 +1,43 @@
...
 
-const pct = x => bigExp(x, 16)
+const EMPTY_SCRIPT = '0x00000001'
 
-const getVoteState = async (voting, id) => {
-  const isOpen = await voting.isVoteOpen(id)
-  const { executed, startDate, snapshotBlock, supportRequired, minAcceptQuorum, voteOverruleWindow, earlyExecution, yea, nay, votingPower, script } = await voting.getVote(id)
-  return { isOpen, isExecuted: executed, startDate, snapshotBlock, support: supportRequired, quorum: minAcceptQuorum, overruleWindow: voteOverruleWindow, earlyExecution, yeas: yea, nays: nay, votingPower, script }
-}
+module.exports = (web3, artifacts) => {
```
sohkai, line 7: I wonder if we should avoid this pattern of injecting the truffle environment transparently, so it's more obvious how to both use the module and how the module is used in a test.

For example:

- It's not immediately clear which imports require this injection
- It can lead to some awkward situations like

```js
const { pct } = require('../helpers/voting')()
```

It would be more clear if we only surfaced direct exports from a module, especially as you can't do the `require('module')()` trick when importing ES modules. For helpers that specifically need the environment, we could migrate them to a `helpers/web3/` folder or etc. and expose a default export from there to hydrate all the internals that assume the environment to exist.

apps/voting-disputable/test/helpers/voting.js
```diff
@@ -1,14 +1,43 @@
 const { bigExp } = require('@aragon/apps-agreement/test/helpers/lib/numbers')
+const { encodeCallScript } = require('@aragon/contract-test-helpers/evmScript')
+const { decodeEventsOfType } = require('@aragon/apps-agreement/test/helpers/lib/decodeEvent')
```
sohkai, line 3: This is honestly getting ridiculous—these should all be coming from the same place (https://github.com/aragon/contract-helpers/issues/27).

apps/voting-disputable/test/helpers/deployer.js
```diff
@@ -0,0 +1,147 @@
+const { pct } = require('./voting')()
```
sohkai, line 1: So I know you really like deployer classes, but is there a reason to prefer this over a simplified version with just functions that don't encapsulate any state?

I'm definitely of the mindset that the less magic and abstraction there are in tests, especially in terms of hidden state, the better :).

facuspagnuolo: I found this one relatively simple compared to the agreement app, for example, so with a bunch of functions, it was straightforward to test. But I agree we must find a rationale and follow the same approach always.

apps/voting-disputable/test/helpers/voting.js
facuspagnuolo: I know, I'm working on a first step to start getting these fixed

apps/voting-disputable/test/helpers/voting.js
facuspagnuolo: Totally agreed! This one was definitely a workaround, based on the new proposed changes on https://github.com/aragon/contract-helpers/pull/35 we should be able to fix this soon

apps/voting-disputable/test/helpers/voting.js
facuspagnuolo: Opened https://github.com/aragon/contract-helpers/pull/35 and https://github.com/aragon/aragon-apps/pull/1198 on top of it to remove all the duplicated logic from Agreements, I promise to update disputable voting app when these two are merged.

PR #1198, opened 2020-07-09
[Blocked] Agreement: Use test helpers

facuspagnuolo:
Requires https://github.com/aragon/contract-helpers/pull/35

This PR removes all the test helpers logic and reuses the one provided in `@aragon/contract-helpers`
I have tested this linking the dependency PR locally and it worked fine, we need to merge and publish and update the dependency version here before merging.


sohkai:
Closing for https://github.com/aragon/aragon-apps/pull/1218.

PR #1199, opened 2020-07-09
Agreement: Add try/catch mechanism for Disputable callbacks

bingen:


facuspagnuolo:
I think we should update this branch cause it is carrying some diff that was already merged

apps/agreement/test/agreement/agreement_settlement.js
```diff
@@ -1,10 +1,12 @@
 const { assertBn } = require('../helpers/assert/assertBn')
 const { assertRevert } = require('../helpers/assert/assertThrow')
-const { assertEvent, assertAmountOfEvents } = require('../helpers/assert/assertEvent')
+const { assertEvent, assertAmountOfEvents, assertAmountOfRawEvents } = require('../helpers/assert/assertEvent')
 const { AGREEMENT_ERRORS } = require('../helpers/utils/errors')
-const { AGREEMENT_EVENTS } = require('../helpers/utils/events')
+const { AGREEMENT_EVENTS, DISPUTABLE_EVENTS } = require('../helpers/utils/events')
 const { CHALLENGES_STATE, RULINGS } = require('../helpers/utils/enums')
 
+const Disputable = artifacts.require('DisputableAppMock')
```
facuspagnuolo, line 8: You can probably define:
```js
  get disputableABI() {
    return this.disputable.abi
  }
```

similar to how we are doing [here](https://github.com/aragon/aragon-apps/blob/master/apps/agreement/test/helpers/wrappers/agreement.js#L20-L22) for the Agreement contract

sohkai:
https://github.com/aragon/aragon-apps/pull/1213 replaces this PR, which was merged onto the wrong branch.

PR #1200, opened 2020-07-09
Agent: replicate rate conversion strategy from #1177

Evalir:
Replicates conversion strategy used in #1177.


apps/agent/app/src/components/Balances.js
```diff
@@ -63,18 +14,19 @@ function useBalanceItems(balances) {
...
-      address,
-      amount,
-      convertedAmount: convertRates[symbol]
-        ? amount.divn(convertRates[symbol])
-        : new BN(-1),
-      decimals,
-      symbol,
-      verified,
-    }))
+    return balances.map(({ address, amount, decimals, symbol, verified }) => {
```
sohkai, line 17: Small nitpick: leaving the implicit return (`({`) would have made these changes smaller :).

PR #1205, opened 2020-07-14
Disputable Voting: Support quiet ending

facuspagnuolo:
Fixes https://github.com/aragon/aragon/issues/836

> Quiet Ending Voting extends the duration of a vote by X amount if the outcome is flipped within Y amount of time before the end of the vote.

This PR proposed the following implementation:

image


apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -690,7 +739,7 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
     * @return True if the given voter can participate a certain vote
     */
     function _canVote(Vote storage vote_, address _voter) internal view returns (bool) {
-        return _isVoteOpenForVoting(vote_) && _hasVotingPower(vote_, _voter);
+        return _isVoteOpenForVoting(vote_) && _hasVotingPower(vote_, _voter) && _voteCaster(vote_, _voter) != _voter;
```
bingen, line 742: This means that voters can’t change their cast votes,  right? Some discussion rings a bell about this, but I don’t remember it. What was the rationale for this change?

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -703,12 +752,12 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
...
     /**
-    * @dev Internal function to check if a voter has already voted on a certain vote
+    * @dev Internal function to check if the vote for a voter has already been casted
     *      It assumes the pointer to the vote is valid
-    * @return True if the given voter has not voted on the requested vote
+    * @return True if a vote for the given voter has already been casted
     */
-    function _hasNotVotedYet(Vote storage vote_, address _voter) internal view returns (bool) {
-        return _voteCaster(vote_, _voter) != _voter;
+    function _hasCastedVote(Vote storage vote_, address _voter) internal view returns (bool) {
```
bingen, line 759: I’ll leave it for someone English native (@sohkai ?), but I think it’s “Cast” instead of “Casted”.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -703,12 +752,12 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
...
+        return _voterState(vote_, _voter) != VoterState.Absent;
```
bingen, line 760: This changes also means that representatives can’t change votes either, which is consistent with holders not being able.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -863,4 +973,23 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
...
+    * @dev Internal function to compute the start date of time duration from the original vote end date
+    *      It considers the pause duration only if the vote has "returned" from being paused,
+    *      and if the pause occurred before the start date of the time duration being queried
+    *      It assumes the pointer to the vote is valid
+    *
+    *                                                  [   queried duration   ]
+    *      [   vote active    ][   vote paused    ][   .   vote active        ]
+    *      ^                                           ^                      ^
+    *      |                                           |                      |
+    *  vote starts                            duration start date         vote ends
```
bingen, line 987: Nice!

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -385,6 +418,7 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
         nay = vote_.nay;
         votingPower = vote_.votingPower;
         script = vote_.executionScript;
+        // TODO: cannot add quiet ending related variables, let's use checkpointed settings instead of copying all variables
```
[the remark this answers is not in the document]
facuspagnuolo: Yes, but I'm planning to solve it by checkpointing the settings, it will be more efficient

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -464,7 +498,7 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
     */
     function withinOverruleWindow(uint256 _voteId) external view voteExists(_voteId) returns (bool) {
         Vote storage vote_ = votes[_voteId];
-        return _isVoteOpenForVoting(vote_) && _withinOverruleWindow(vote_);
+        return _isVoteOpenForVoting(vote_) && _withinOverrulePeriod(vote_);
```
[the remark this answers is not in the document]
facuspagnuolo: No, thanks for the heads up

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -834,23 +926,41 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
...
+        vote_.yea = yeas;
+        vote_.nay = nays;
         vote_.voters[_voter] = _supports ? VoterState.Yea : VoterState.Nay;
         emit CastVote(_voteId, _voter, _supports, voterStake);
+
+        // Avoid checking flips for quiet ending if first vote or outside quiet ending period
+        if (voterStake < yeas.add(nays) && _withinQuietEndingPeriod(vote_)) {
+            bool isAccepted = _isAccepted(yeas, yeas.add(nays), votingPower, supportRequired, minimumAcceptanceQuorum);
+            if (wasAccepted != isAccepted) {
+                vote_.quietEndingExtendedSeconds = vote_.quietEndingExtendedSeconds.add(vote_.quietEndingExtension);
```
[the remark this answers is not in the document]
facuspagnuolo: Yes, but that is the main purpose of quiet ending

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -834,23 +926,41 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
...
+            nays = nays.add(voterStake);
         }
 
+        vote_.yea = yeas;
+        vote_.nay = nays;
         vote_.voters[_voter] = _supports ? VoterState.Yea : VoterState.Nay;
         emit CastVote(_voteId, _voter, _supports, voterStake);
+
+        // Avoid checking flips for quiet ending if first vote or outside quiet ending period
+        if (voterStake < yeas.add(nays) && _withinQuietEndingPeriod(vote_)) {
```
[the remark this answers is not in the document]
facuspagnuolo: True, let me remove that :)

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Yea haha, we already agreed on that, but sometimes I forget about it haha

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Yes, simply because since multiple extensions are allowed, if we allow voters to change their vote, then they could extend votes indefinitely

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Exactly, since multiple extensions are allowed, if we allow representatives to change their vote, then they could extend votes indefinitely due to quiet ending

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -834,23 +926,41 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
...
+        uint64 minimumAcceptanceQuorum = vote_.minAcceptQuorumPct;
+        bool wasAccepted = _isAccepted(yeas, yeas.add(nays), votingPower, supportRequired, minimumAcceptanceQuorum);
 
         // If voter had previously voted, decrease count
-        if (state == VoterState.Yea) {
-            vote_.yea = vote_.yea.sub(voterStake);
-        } else if (state == VoterState.Nay) {
-            vote_.nay = vote_.nay.sub(voterStake);
+        // Note that votes can be changed only during the overrule window
+        VoterState previousVoterState = _voterState(vote_, _voter);
```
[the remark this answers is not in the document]
facuspagnuolo: I tried it but I honestly don't think it looks cleaner. If you see a cleaner way please share!

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -587,6 +621,20 @@ contract DisputableVoting is DisputableAragonApp, IForwarder {
         emit ChangeOverruleWindow(_overruleWindow);
     }
 
+    /**
+    * @dev Internal function to change the quiet ending configuration
+    * @param _quietEndingPeriod New quiet ending period in seconds
+    * @param _quietEndingExtension New quiet ending extension in seconds
+    */
+    function _changeQuietEndingPeriod(uint64 _quietEndingPeriod, uint64 _quietEndingExtension) internal {
+        require(_quietEndingPeriod <= voteTime, ERROR_INVALID_QUIET_ENDING_PERIOD);
```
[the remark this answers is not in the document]
bingen: I see, thanks! And actually, as all those parameters can be zero, this way we can have quiet ending without overrule window.
But now I have another question: why does the quiet ending period need to be <= the quiet ending extension?
Finally, I would personally understand better those diagrams if the “Flipped” arrow was disassociated from the start of the quiet ending extension.

PR #1208, opened 2020-07-14
Disputable voting app: Show real data for status label

delfipolito:


apps/voting-disputable/app/src/components/VoteCard/VoteCard.js
```diff
@@ -41,8 +41,7 @@ function VoteCard({ vote, onOpen }) {
   } = vote
 
   //TODO: Remove this once we have real data
-  vote.disputable = getDisputableVoteById(voteId)
-
+  vote.disputable.action = getDisputableVoteById(voteId)
```
andy-hook, line 44: Tiny nitpick: should we rename this temp func to be `getMockVoteActionById` or similar? and update the mock references to be `voteAction` rather than `vote`. Really not a big deal! just took me a second to see which parts were being mocked vs real.

apps/voting-disputable/app/src/script.js
```diff
@@ -337,6 +337,22 @@ function loadVoteData(voteId) {
   )
 }
 
+function loadVoteDisputableInfo(voteId) {
+  // Wrap with retry in case the vote is somehow not present
+  return retryEvery(() =>
```
andy-hook, line 342: Genuinely curious, under which conditions would the vote not be present? while the app is syncing?

delfipolito: Sorry here it should say 'Wrap with retry in case the disputable info is somehow not present'. I already update it. Its a validation in case it fails for some reason. Like for example if the event is firing a little bit before the data is available through the .call method.

PR #1207, opened 2020-07-14
Agreements: Use new ACL version

bingen:

PR #1209, opened 2020-07-14
Agreement: Use new ACL version

bingen:
Replaces partially #1207

### DIFF
--- a/apps/agreement/contracts/Agreement.sol
+++ b/apps/agreement/contracts/Agreement.sol
@@ -1127,9 +1127,7 @@ contract Agreement is IAgreement, AragonApp {
             return false;
         }
 
-        // TODO: update with new ACL version: no need to pass challenger address by parameter
-        bytes memory params = ConversionHelpers.dangerouslyCastUintArrayToBytes(arr(_challenger));
-        return currentKernel.hasPermission(_challenger, address(_disputable), CHALLENGE_ROLE, params);
+        return currentKernel.hasPermission(_challenger, address(_disputable), CHALLENGE_ROLE, new bytes(0));
     }
 
     /**
--- a/apps/agreement/contracts/test/mocks/disputable/AragonAppFeesCashierMock.sol
+++ b/apps/agreement/contracts/test/mocks/disputable/AragonAppFeesCashierMock.sol
@@ -7,6 +7,7 @@ import "@aragon/os/contracts/lib/arbitration/IAragonAppFeesCashier.sol";
 
 contract AragonAppFeesCashierMock is IAragonAppFeesCashier, EtherTokenConstant, IsContract {
     string private constant ERROR_WRONG_TOKEN = "AAFC_WRONG_TOKEN";
+    string private constant ERROR_NON_ZERO_VALUE = "AAFC_NON_ZERO_VALUE";
 
     ERC20 internal token;
     uint256 internal amount;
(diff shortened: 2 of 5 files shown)


apps/agreement/contracts/test/mocks/disputable/AragonAppFeesCashierMock.sol
```diff
@@ -41,7 +42,8 @@ contract AragonAppFeesCashierMock is IAragonAppFeesCashier, EtherTokenConstant,
         }
     }
 
-    function payAppFees(bytes32, bytes) external {
+    function payAppFees(bytes32, bytes) external payable {
+        require(msg.value == 0, ERROR_NON_ZERO_VALUE);
```
sohkai, line 46: Maybe we should have a different mock method for this that requires ETH to be sent?

Otherwise this precludes token fees at the moment.

apps/agreement/contracts/Agreement.sol
```diff
@@ -1127,9 +1127,7 @@ contract Agreement is IAgreement, AragonApp {
             return false;
         }
 
-        // TODO: update with new ACL version: no need to pass challenger address by parameter
-        bytes memory params = ConversionHelpers.dangerouslyCastUintArrayToBytes(arr(_challenger));
-        return currentKernel.hasPermission(_challenger, address(_disputable), CHALLENGE_ROLE, params);
+        return currentKernel.hasPermission(_challenger, address(_disputable), CHALLENGE_ROLE, new bytes(0));
```
sohkai, line 1130: To double check, the reason we don't need to add the `challenger` any more is because this address is now available by default in the ACLOracle?

apps/agreement/contracts/Agreement.sol
```diff
@@ -1127,9 +1127,7 @@ contract Agreement is IAgreement, AragonApp {
...
     }
 
     /**
```
sohkai, line 1133: I think we forgot the most important change here—updating the [`canPerform()`](https://github.com/aragon/aragon-apps/blob/80e4b5935fbac4bb2635e6c21f72ecc3169485be/apps/agreement/contracts/Agreement.sol#L662) function.

facuspagnuolo: True, I think the signature was updated before but the previous implementation was kept. Pushed a fix here https://github.com/aragon/aragon-apps/pull/1212

facuspagnuolo: True, I think the signature was updated before but the previous implementation was kept. Pushed a fix here https://github.com/aragon/aragon-apps/pull/1212

apps/agreement/contracts/test/mocks/disputable/AragonAppFeesCashierMock.sol
facuspagnuolo: True, let me know @bingen if there is a reason for this change, otherwise we can address it in #1212

apps/agreement/contracts/Agreement.sol
facuspagnuolo: Exactly, we are testing that here https://github.com/aragon/aragon-apps/blob/master/apps/agreement/test/agreement/agreement_permissions.js#L153-L206

apps/agreement/contracts/test/mocks/disputable/AragonAppFeesCashierMock.sol
facuspagnuolo: Just in case https://github.com/aragon/aragonOS/pull/593

PR #1212, opened 2020-07-15
Agreement: Improve ACL oracle implementation

facuspagnuolo:


apps/agreement/package.json
```diff
@@ -24,7 +24,7 @@
     "apm:publish:patch": "aragon apm publish patch --files public/ --prepublish-script apm:prepublish"
   },
   "dependencies": {
-    "@aragon/os": "5.0.0-beta.1"
+    "@aragon/os": "^5.0.0-beta.2"
```
sohkai, line 27: Probably best to keep this pinned so we can make changes consciously :).

facuspagnuolo: Cool, I'll unpin it in the next PR I'm about to push :)

PR #1214, opened 2020-07-15
Agreement: Improve can close check for disputables

facuspagnuolo:
Related to https://github.com/aragon/aragonOS/pull/594

This PRs improve the `close` function of the Agreement app to avoid calling `canClose` if the sender is the disputable app

### DIFF
--- a/apps/agreement/contracts/Agreement.sol
+++ b/apps/agreement/contracts/Agreement.sol
@@ -325,11 +325,9 @@ contract Agreement is IAgreement, AragonApp {
     */
     function closeAction(uint256 _actionId) external {
         Action storage action = _getAction(_actionId);
-        (IDisputable disputable, CollateralRequirement storage requirement) = _getDisputableFor(action);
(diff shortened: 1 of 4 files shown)


apps/agreement/contracts/Agreement.sol
```diff
@@ -1158,10 +1160,11 @@ contract Agreement is IAgreement, AragonApp {
     /**
     * @dev Tell whether an action can be closed
     * @param _action Action instance to be queried
+    * @param _checkingFromDisputable Whether the check is being performed from the disputable itself
     * @return True if the action can be closed, false otherwise
     */
-    function _canClose(Action storage _action) internal view returns (bool) {
-        return _canProceed(_action) && _action.disputable.canClose(_action.disputableActionId);
+    function _canClose(Action storage _action, bool _checkingFromDisputable) internal view returns (bool) {
```
bingen, line 1166: Inside this function we already have the disputable, so I think we could get rid of `_checkingFromDisputable` param and check it inside. For the external view function it’s not a problem, actually the opposite, more consistent.

PR #1215, opened 2020-07-15
Agreement: Support payable actions

facuspagnuolo:
This PR basically implements the new forwarding interface (payable + context) in the disputable app used to test the agreement app, forcing the later to handle ETH fees.


apps/agreement/contracts/example/RegistryApp.sol
```diff
@@ -121,23 +122,29 @@ contract Registry is DisputableAragonApp {
...
     * @return True if the given address can submit actions, false otherwise
     */
-    function canForward(address _sender, bytes _data) public view returns (bool) {
-        (bytes32 id,) = _decodeData(_data);
-        return canPerform(_sender, REGISTER_ENTRY_ROLE, arr(id));
+    function canForward(address _sender, bytes _data) external view returns (bool) {
+        return _canForward(_sender, _data);
+    }
+
+    // TODO: how are we going to implement this from the disputable apps?
```
facuspagnuolo, line 145: This is an important thing that came up we will need to figure out soon

apps/agreement/contracts/Agreement.sol
```diff
@@ -773,16 +775,23 @@ contract Agreement is IAgreement, AragonApp {
             return;
         }
 
-        // Get staking pool
-        Staking staking = stakingFactory.getOrCreateInstance(token);
-
-        // Pull required fee amount from staking pool
-        _lockBalance(staking, _submitter, amount);
-        _slashBalance(staking, _submitter, address(this), amount);
+        uint256 sentEth = 0;
```
bingen, line 778: Why do we need this?

apps/agreement/contracts/example/RegistryApp.sol
```diff
@@ -121,23 +122,29 @@ contract Registry is DisputableAragonApp {
     * @notice Schedule a new entry
     * @dev IForwarder interface conformance
     * @param _data Data requested to be registered
+    * @param _context Link to a human-readable text giving context for the given action
     */
-    function forward(bytes memory _data) public {
```
bingen, line 125: Why are we removing the `memory` keyword?

apps/agreement/contracts/example/RegistryApp.sol
```diff
@@ -121,23 +122,29 @@ contract Registry is DisputableAragonApp {
...
-        require(canForward(msg.sender, _data), ERROR_CANNOT_REGISTER);
+    function forward(bytes _data, bytes _context) external payable {
```
bingen, line 127: Isn’t first parameter an EVM script here? That’s confusing to me. Why do we need forwarding then?

apps/agreement/contracts/example/RegistryApp.sol
facuspagnuolo: It is calldata now that it is external :)

apps/agreement/contracts/Agreement.sol
facuspagnuolo: Simply to support both ETH or ERC20 fees

apps/agreement/contracts/example/RegistryApp.sol
facuspagnuolo: Yes it is, but I'm playing with it in the Registry app. Although it is a sample app, we can actually not make it forwardable

apps/agreement/contracts/example/RegistryApp.sol
bingen: Oh, of course, haha! :facepalm:

apps/agreement/contracts/Agreement.sol
bingen: But can’t you use `msg.value` everywhere?

facuspagnuolo: True 😅

PR #1217, opened 2020-07-16
Tokens: Fix token SyncIndicator position

Evalir:
Seems we hadn't updated the Token Manager app for allocating the correct spacing in between the sync indicator and the helpscout icon.

Before:
Screen Shot 2020-07-16 at 5 29 47 PM

After:
Screen Shot 2020-07-16 at 5 29 40 PM


Evalir:
Whoops! Indeed!

PR #1222, opened 2020-07-19
chore: migrate to yarn workspaces

sohkai:
Removes some old install scripts and greatly simplifies the overall repo management and set up.

Improves the cold bootstrap time to around one minute (less than one if yarn caches are hot), compared to 10s of minutes previously, on a 2018 6 core Mbp.

The only behaviour I removed is to not install app frontends by default when a normal (non-CI) user clones and installs the repo. I think this is ok—it was excessively to install all the frontends separately anyway.

-----

Also:

- Updates all the apps' `@aragon/os` dependencies to be on `4.4.0`
- Removes CI for the Token Manager's frontend tests (see more comprehensive CI with Github Actions in https://github.com/aragon/aragon-apps/pull/1223)


sohkai:
> I think there are a few lock files missing tho: agreement and vault

These are actually not needed anymore; they're captured all in the root yarn.lock due to how workspaces work :).

The only yarn.locks we need are those in the app frontends, as they're not managed by the yarn workspace (thinking about how to do this, but deciding it's not really worth it).

facuspagnuolo:
I think there is some issue with the version we upgrade for the solidity coverage

PR #1223, opened 2020-07-19
CI: Add Github Actions CI

sohkai:
For convenience, builds on top of https://github.com/aragon/aragon-apps/pull/1222.

Adds scoped, per-app Github Action jobs.

Assuming this works as anticipated, CI should only be running in a scoped matter based on the files actually being changed in each PR. This also allows us to more easily control our CI environment to include testing and linting in each app's frontend.

(Alright Github, third time's the charm for us to do this migration here)

TODO:

- [ ] Add scheduled option for running all tests every night
- [ ] Add coverage reporting (either coveralls or codecov config)


sohkai:
Going to close this for now; we're going to split this repo up into individual applications and then the actions CI presented here won't be as useful.

PR #1227, opened 2020-07-21
DisputableVoting: Checkpoint settings

facuspagnuolo:
It is recommended to review per commit

### DIFF
--- a/apps/voting-disputable/contracts/DisputableVoting.sol
+++ b/apps/voting-disputable/contracts/DisputableVoting.sol
@@ -78,7 +78,7 @@ contract DisputableVoting is IForwarder, DisputableAragonApp {
         uint64 quietEndingExtension;
     }
 
-    struct CastVote {
+    struct VoteCast {
         VoterState state;
         address caster;
     }
@@ -96,7 +96,7 @@ contract DisputableVoting is IForwarder, DisputableAragonApp {
         uint64 pauseDuration;                               // Duration in seconds while the vote has been paused
         uint64 quietEndingExtendedSeconds;                  // Total number of seconds a vote was extended due to quiet ending
         bytes executionScript;                              // EVM script attached to the vote
-        mapping (address => CastVote) castVotes;            // Cast votes information indexed by voter address: status and caster address
+        mapping (address => VoteCast) castVotes;            // Cast votes information indexed by voter address: status and caster address
     }
 
     uint64 public voteTime;                                 // Duration of each vote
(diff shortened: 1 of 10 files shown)


apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -68,58 +69,57 @@ contract DisputableVoting is IForwarder, DisputableAragonApp {
...
-
-    uint64 public overruleWindow;
-    mapping (address => address) internal representatives;
+    uint256 public votesLength;                             // Number of existing votes created
+    mapping (uint256 => Vote) internal votes;               // List of votes indexed by ID
+    mapping (address => address) internal representatives;  // List of representatives indexed by voter address
 
     event StartVote(uint256 indexed voteId, address indexed creator, bytes context);
-    event CastVote(uint256 indexed voteId, address indexed voter, bool supports, uint256 stake);
+    event VoteCast(uint256 indexed voteId, address indexed voter, bool supports, uint256 stake);
```
bingen, line 114: Why this change? It seems to be the opposite order of the other events (verb + direct object), like `StartVote, `ChangeSupportRequired`, etc.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -398,111 +388,123 @@ contract DisputableVoting is IForwarder, DisputableAragonApp {
...
+    * @return caster Address of the caster of the voter's vote
     */
-    function getVoteDisputableInfo(uint256 _voteId)
-        external
-        view
-        voteExists(_voteId)
-        returns (uint256 actionId, uint64 pausedAt, uint64 pauseDuration, VoteStatus status)
-    {
+    function getCastVote(uint256 _voteId, address _voter) external view voteExists(_voteId) returns (VoterState state, address caster) {
         Vote storage vote_ = votes[_voteId];
```
bingen: Should we check that the vote exists first?

apps/voting-disputable/truffle.js
```diff
@@ -1,7 +1,7 @@
 const TruffleConfig = require('@aragon/truffle-config-v5/truffle-config')
 
 TruffleConfig.compilers.solc.version = '0.4.24'
-TruffleConfig.compilers.solc.settings.optimizer.runs = 100
+TruffleConfig.compilers.solc.settings.optimizer.runs = 1000
```
bingen, line 4: :heart:

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: The problem was the struct, I can probably swap the struct then

PR #1229, opened 2020-07-23
Agreement: Implement Staking ILockManager

bingen:
Depends on: https://github.com/aragon/staking/pull/47

Merge after #1226 (and change PR base)

### DIFF
--- a/apps/agreement/contracts/Agreement.sol
+++ b/apps/agreement/contracts/Agreement.sol
@@ -6,7 +6,7 @@ pragma solidity 0.4.24;
 
 import "@aragon/os/contracts/apps/AragonApp.sol";
 import "@aragon/os/contracts/apps/disputable/IAgreement.sol";
-import "@aragon/os/contracts/apps/disputable/IDisputable.sol";
+import "@aragon/os/contracts/apps/disputable/DisputableAragonApp.sol";
 import "@aragon/os/contracts/common/ConversionHelpers.sol";
 import "@aragon/os/contracts/common/SafeERC20.sol";
 import "@aragon/os/contracts/common/TimeHelpers.sol";
@@ -83,7 +83,7 @@ contract Agreement is IAgreement, AragonApp {
     }
 
     struct Action {
-        IDisputable disputable;             // Address of the disputable that created the action
+        DisputableAragonApp disputable;     // Address of the disputable that created the action
         uint256 disputableActionId;         // Identification number of the disputable action in the context of the disputable instance
         uint256 collateralRequirementId;    // Identification number of the collateral requirements for the given action
         uint256 settingId;                  // Identification number of the agreement setting for the given action
@@ -186,7 +186,7 @@ contract Agreement is IAgreement, AragonApp {
         DisputableInfo storage disputableInfo = disputableInfos[_disputableAddress];
         _ensureInactiveDisputable(disputableInfo);
 
-        IDisputable disputable = IDisputable(_disputableAddress);
+        DisputableAragonApp disputable = DisputableAragonApp(_disputableAddress);
         disputableInfo.activated = true;
         emit DisputableAppActivated(disputable);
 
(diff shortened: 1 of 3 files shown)


apps/agreement/contracts/Agreement.sol
```diff
@@ -16,9 +16,10 @@ import "@aragon/os/contracts/lib/math/SafeMath.sol";
 import "@aragon/os/contracts/lib/math/SafeMath64.sol";
 import "@aragon/staking/contracts/Staking.sol";
 import "@aragon/staking/contracts/StakingFactory.sol";
+import "@aragon/staking/contracts/locking/ILockManager.sol";
 
 
-contract Agreement is IAgreement, AragonApp {
+contract Agreement is IAgreement, AragonApp, ILockManager {
```
sohkai, line 22: Minor nitpick, but can we move `ILockManager` to be first?

It seems fairly well isolated and independent from the other two, which are a lot more heavy.

apps/agreement/contracts/Agreement.sol
```diff
@@ -752,6 +753,17 @@ contract Agreement is IAgreement, AragonApp {
         return _isDisputed(challenge);
     }
 
+    /**
+    * @notice Check if a certain amount for a certain address can be unlocked
```
sohkai, line 757: Reading this out of context, this description sounds a bit vague and disconnected from the staking pool.

apps/agreement/contracts/Agreement.sol
```diff
@@ -752,6 +753,17 @@ contract Agreement is IAgreement, AragonApp {
...
+    * @dev It always returns false because if we allow owners to unlock by themselves,
+    *      then the unlock call on closing or settling actions could fail due to
+    *      insufficient balance.
```
sohkai, line 760: Just to double check, if the Agreement calls unlock itself, as the manager, this function won't be checked?

apps/agreement/test/agreement/agreement_lock_manager.js
```diff
@@ -0,0 +1,82 @@
+const deployer = require('../helpers/utils/deployer')(web3, artifacts)
+const { injectWeb3, injectArtifacts } = require('@aragon/contract-helpers-test')
```
sohkai, line 2: Not opposed to leaving this here, but since we don't seem to use any part of the `contract-helpers-test` module, we could remove it.

apps/agreement/test/agreement/agreement_lock_manager.js
```diff
@@ -0,0 +1,82 @@
...
+    disputable = await deployer.deployAndInitializeDisputableWrapper({ owner, activate: false, submitters: [submitter], challengers: [challenger] })
+    agreement = disputable.agreement
+    collateralToken = disputable.collateralToken
+    actionCollateral = disputable.actionCollateral
+  })
+
+  describe('lock manager interface', () => {
+    let actionId
+
+    context('can’t unlock', () => {
```
sohkai, line 22:
```suggestion
    context("can't unlock", () => {
```

Might be a bit easier 😄

apps/agreement/test/agreement/agreement_lock_manager.js
```diff
@@ -0,0 +1,82 @@
...
+      const cannot_unlock = async (user) => {
```
sohkai, line 23:
```suggestion
      const cannotUnlock = async (user) => {
```

apps/agreement/contracts/Agreement.sol
bingen: Sure, I’ll try. Do we have a standard criteria for this?

apps/agreement/test/agreement/agreement_lock_manager.js
bingen: hm, ok, but ask @bpierre about it. He convinced me to use ’, “ and ” a while ago :wink:

apps/agreement/contracts/Agreement.sol
bingen: Yes, that’s correct. As you can see [here](https://github.com/aragon/staking/blob/master/contracts/Staking.sol#L621), there is a shortcut for that case.

PR #1230, opened 2020-07-24
DisputableVoting: Fix quiet ending extension

facuspagnuolo:


apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -618,7 +617,6 @@ contract DisputableVoting is IForwarder, DisputableAragonApp {
     */
     function _changeQuietEndingPeriod(Setting storage _setting, uint64 _quietEndingPeriod, uint64 _quietEndingExtension) internal {
         require(_quietEndingPeriod <= voteTime, ERROR_INVALID_QUIET_ENDING_PERIOD);
-        require(_quietEndingExtension <= _quietEndingPeriod, ERROR_INVALID_QUIET_ENDING_EXTENSION);
```
bingen, line 621: But maybe it would be reasonable to keep it shorter than vote time?

PR #1231, opened 2020-07-24
Agreement: Ignore re-closing requests

facuspagnuolo:
Related to https://github.com/aragon/aragonOS/pull/594

### DIFF
--- a/apps/agreement/contracts/Agreement.sol
+++ b/apps/agreement/contracts/Agreement.sol
@@ -315,21 +315,23 @@ contract Agreement is IAgreement, AragonApp {
     /**
     * @notice Close action #`_actionId`
     * @dev If allowed by the originating Disputable app, this function allows users to close actions that are not:
-    *      - Closed
     *      - Currently challenged
     *      - Ruled as voided
     *      - Ruled in favour of the submitter
+    *      It ignores if the action is already closed, without re-closing it
     *      Initialization check is implicitly provided by `_canClose()` as disputable actions can be created only
     *      via `newAction()` which already requires initialization implicitly through `activate()`
     * @param _actionId Identification number of the action to be closed
     */
     function closeAction(uint256 _actionId) external {
         Action storage action = _getAction(_actionId);
-        require(_canClose(action), ERROR_CANNOT_CLOSE_ACTION);
+        require(_canClose(action, true), ERROR_CANNOT_CLOSE_ACTION);
 
-        (, CollateralRequirement storage requirement) = _getDisputableFor(action);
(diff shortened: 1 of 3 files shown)


apps/agreement/contracts/Agreement.sol
```diff
@@ -315,21 +315,23 @@ contract Agreement is IAgreement, AragonApp {
...
     *      - Ruled in favour of the submitter
+    *      It ignores if the action is already closed, without re-closing it
     *      Initialization check is implicitly provided by `_canClose()` as disputable actions can be created only
     *      via `newAction()` which already requires initialization implicitly through `activate()`
     * @param _actionId Identification number of the action to be closed
     */
     function closeAction(uint256 _actionId) external {
         Action storage action = _getAction(_actionId);
-        require(_canClose(action), ERROR_CANNOT_CLOSE_ACTION);
+        require(_canClose(action, true), ERROR_CANNOT_CLOSE_ACTION);
```
sohkai, line 328: I find the `ignoreIfClosed` parameter a bit difficult to understand, partly because its name is a somewhat vague.

However, I also wonder if we shouldn't just move this require down into the `if (!action.closed)` block, and avoid all the complexity. If the action is closed, we don't need to do any more validation on it other than to skip it.

facuspagnuolo: Addressed this in the last commit, let me know your thoughts. Note that I had to change some tests since it affects some related behavior

apps/agreement/contracts/Agreement.sol
```diff
@@ -315,18 +315,21 @@ contract Agreement is IAgreement, AragonApp {
...
+    *      It ignores if the action is already closed, without re-closing it
     *      Initialization check is implicitly provided by `_canClose()` as disputable actions can be created only
     *      via `newAction()` which already requires initialization implicitly through `activate()`
     * @param _actionId Identification number of the action to be closed
     */
     function closeAction(uint256 _actionId) external {
         Action storage action = _getAction(_actionId);
-        require(_canClose(action), ERROR_CANNOT_CLOSE_ACTION);
+        if (action.closed) {
+            return;
```
sohkai, line 329: To double check, we were discussing whether we should return a flag here to mark if you had actually closed the action in this request or not.

However, this is actually not that useful:

1. We only unlock tokens from the staking pool, so the caller usually doesn't receive any tokens directly from this call
2. For a contract that really wanted to know, the `closed` status is available from `getAction()` and should be called first (similar to the PR in aragonOS that we closed)

facuspagnuolo: Yea, I finally decided no to add it to be honest.

PR #1232, opened 2020-07-27
agreements: Rollback payable actions

bingen:
This PR essentially reverts https://github.com/aragon/aragon-apps/pull/1215
The reason for the change is mainly that we are going to use the Staking pool to pay for fees. The initial reason for supporting ETH was to improve UX by not needing an extra tx for approval, but as long as there’s enough funds and allowance in Staking, UX for paying fees is also quite good, so we expect that either stable coins or “native” tokens will be used.


apps/agreement/contracts/Agreement.sol
```diff
@@ -281,7 +281,7 @@ contract Agreement is IAgreement, AragonApp {
     * @param _context Link to a human-readable text providing context for the given action
```
sohkai, line 281: Looking at #1215, should we also remove some of the unused constants now?

apps/agreement/test/agreement/agreement_new_action.js
```diff
@@ -216,7 +216,7 @@ contract('Agreement', ([_, owner, submitter, someone]) => {
                               const sentEth = bn(1)
```
sohkai, line 216: Should we also remove the `value` parameter in [`itHandlesNewActionsCorrectly()`](https://github.com/aragon/aragon-apps/blob/forward_non_payable/apps/agreement/test/agreement/agreement_new_action.js#L66)?

bingen:
> Should we also rollback the changes to the RegistryApp?

I think the changes to the RegistryApp in the original PR were unrelated. `forward` function wasn’t actually payable there.

bingen:
> LGTM! The other change we can revert is removing the `value` parameter from the `forward` method of the Disputable wrapper :)

We cannot do that, if we want to keep testing newActions sent along with ETH, which have to revert. I think it’s good to keep those tests to make sure no ETH can be leaked.

PR #1236, opened 2020-07-28
DisputableVoting: Update latest Agreement and aragonOS versions

facuspagnuolo:


apps/voting-disputable/package.json
```diff
@@ -24,16 +25,17 @@
...
     "@aragon/apps-shared-migrations": "1.0.0",
-    "@aragon/cli": "^7.1.3",
+    "@aragon/cli": "^7.0.0",
     "@aragon/contract-helpers-test": "^0.1.0-rc.2",
     "@aragon/minime": "^1.0.0",
-    "@aragon/staking": "^0.2.3",
-    "@aragon/truffle-config-v5": "^1.0.0",
+    "@aragon/staking": "^0.2.4",
+    "@aragon/truffle-config-v5": "^1.0.1",
+    "chai": "^4.2.0",
```
bingen, line 38: Curious about this, why do we need it now here explicitly?

facuspagnuolo: it's a peer dependency of the contract-helpers-test library now :)

PR #1240, opened 2020-07-31
DisputableVoting: Add tests to check role constants

bingen:


bingen:
>  Coverage decreased (-0.2%) to 97.338% 
Hm, how come a PR that only adds tests decreases the coverage?? :thinking:

PR #1243, opened 2020-08-04
Agreement: Rollback ACL v2 related changes

facuspagnuolo:
Related to https://github.com/aragon/aragonOS/pull/606 and https://github.com/aragon/aragonOS/pull/607


apps/agreement/contracts/Agreement.sol
```diff
@@ -1176,7 +1179,8 @@ contract Agreement is ILockManager, IAgreement, AragonApp {
             return false;
         }
 
-        return currentKernel.hasPermission(_challenger, address(_disputable), CHALLENGE_ROLE, new bytes(0));
+        bytes memory params = ConversionHelpers.dangerouslyCastUintArrayToBytes(arr(_challenger));
+        return currentKernel.hasPermission(_challenger, address(_disputable), CHALLENGE_ROLE, params);
```
sohkai, line 1183: This part is a bit obfuscated to understand why we need to push the challenger in again; perhaps it'd be useful to leave a note or two about why we still need this and how it'll be removed in the future?

PR #1092, opened 2020-03-07
Agent: move to buidler

ajsantander:
Upgrade Agent to be based on buidler-aragon.

PR #1245, opened 2020-08-06
Agent: Move to buidler

0xGabi:
Replace #1092 

This PR applies the changes to move the Agent app to use buidler as a developer tool and the Aragon Buidler plugin.

Changes are straight forward now that tests were migrated to truffle5. Once we get one of the apps merged I can move fast and replicate the same for the rest as is a mechanical process.

Note I also start using `yarn` as default. Let me know what you think and if I may rollback the change.

A few open questions we can address on following PRs are:
1. Migrate `solium` rules to use `solhint`
2. Decide about the rule we have on the node engine (currently we are fixed to node 10) with the migration to buidler I think we can safely move to node 12.


apps/agent/contracts/test/TestImports.sol
```diff
@@ -1,8 +1,12 @@
 pragma solidity 0.4.24;
 
 import "@aragon/os/contracts/acl/ACL.sol";
+import "@aragon/os/contracts/apm/APMRegistry.sol";
```
0xGabi, line 4: The new imports are needed to deploy the Aragon bases on the local ganache chain we ran with the buidler plugin.

apps/agent/scripts/ganache-cli.sh
```diff
@@ -0,0 +1,80 @@
+#!/usr/bin/env bash
```
0xGabi, line 1: This file will live on `@aragon/contract-helpers-test`. I include it locally until we publish the new version.

apps/agent/buidler.config.js
```diff
@@ -0,0 +1,42 @@
+const { usePlugin } = require('@nomiclabs/buidler/config')
+const hooks = require('./scripts/buidler-hooks')
+
+usePlugin('@aragon/buidler-aragon')
```
sohkai, line 4: Rather than use the buidler-aragon plugin, I was wondering if we shouldn't move to a vanilla buidler set up for now? That way we might be able to use buidlerevm or other defaults rather than configure our own ganache chain / etc.

I'm not particularly concerned at the moment about providing a good frontend developer experience for others in these apps.

apps/agent/arapp.json
```diff
@@ -1,28 +1,25 @@
 {
+  "appName": "agent.aragonpm.eth",
```
sohkai, line 2: Similarly, I was thinking of removing aragonCLI as a dependency, but still using it to publish these frontends for now.

apps/agent/buidler.config.js
0xGabi: As discussed offline, this is a good step forward.

0xGabi:
I'll merge with the main buidler branch where I will aggregate the rest of the apps changes.

PR #1246, opened 2020-08-13
Agreement: Implement app fees cashier sync

facuspagnuolo:
This PR adds a way to sync the agreement app with the Arbitrator's app fees cashier in case this changes.


apps/agreement/test/agreement/agreement_new_action.js
```diff
@@ -154,17 +150,23 @@ contract('Agreement', ([_, owner, submitter, someone]) => {
                     })
                   }
 
-                  context('when the transaction fees module is set', () => {
-                    context('when the transaction fee is zero', () => {
+                  context('when the app fees cashier module is set', () => {
+                    let aragonAppFeesCashier
+
+                    beforeEach('load app fees cashier', async () => {
```
sohkai, line 156: Theoretically this doesn't need to be a `beforeEach()`, right?

apps/agreement/contracts/Agreement.sol
```diff
@@ -246,20 +247,37 @@ contract Agreement is ILockManager, IAgreement, AragonApp {
...
-        _newSetting(_arbitrator, _aragonAppFeesCashier, _title, _content);
+        _newSetting(_arbitrator, _setAppFeesCashier, _title, _content);
+    }
+
+    /**
+    * @notice Sync app fees cashier address
+    * @dev The app fees cashier address is being cached in the contract, in case the arbitrator changes the address
+    *      this permission-less function allows anyone to have it synced
+    */
+    function syncAppFeesCashier() external {
```
sohkai, line 271: So as it is now, you could:

- Initialize or create a new setting without setting the app cashier
- Then anyone can permissionlessly call `syncAppFeesCashier()` to sync it

While I don't think this behaviour is incorrect—again, I don't see any situation where you would not want to be synced—perhaps this points us to prefer removing the `_setAppFeesCashier` flag entirely?

apps/agreement/test/agreement/agreement_new_action.js
facuspagnuolo: We do, some of these tests are testing different settings with the arbitrator, therefore we cannot trust it is not going to change

apps/agreement/contracts/Agreement.sol
facuspagnuolo: As discussed offline we agreed on having the flag to allow users not to be linked to the Arbitrator's subscription module

PR #1247, opened 2020-08-19
Agreement: Update aragonOS to v5.0.0-beta.8

facuspagnuolo:
This PR makes the current Agreement app compatible with the latest changes introduced in aragonOS v5.0.0-beta.8

PR #1248, opened 2020-08-19
Agreement: Security-related bug fixes

facuspagnuolo:
**MERGE AFTER #1247**

This PR includes:
- Disable funds recovery
- Avoid front-running agreement signatures


izqui (approved):

apps/agreement/test/agreement/agreement_initialize.js
```diff
@@ -70,6 +70,10 @@ contract('Agreement', ([_, EOA]) => {
         const { recipient: expectedCashier } = await arbitrator.getSubscriptionFees(ZERO_ADDRESS)
         assert.equal(setting.aragonAppFeesCashier, expectedCashier, 'aragon app fees cashier does not match')
       })
+
+      it('does not allow recovering funds', async () => {
```
sohkai, line 74: We may want to test this with a slightly more complicated example on a mock org.

You can set an EOA as the 'default vault', and then try to call `transferToVault()`.

apps/agreement/contracts/Agreement.sol
```diff
@@ -289,15 +290,18 @@ contract Agreement is ILockManager, IAgreement, IArbitrable, IACLOracle, AragonA
     }
 
     /**
-    * @notice Sign the agreement
+    * @notice Sign the agreement up-to setting #`_settingId`
+    * @param _settingId Last setting ID the user is agreeing with, including all past settings
```
sohkai, line 294: I would say it's not "all past settings"; they're only signing for this particular version (the past ones don't matter).

apps/agreement/contracts/Agreement.sol
```diff
@@ -289,15 +290,18 @@ contract Agreement is ILockManager, IAgreement, IArbitrable, IACLOracle, AragonA
...
     */
-    function sign() external isInitialized {
-        uint256 currentSettingId = _getCurrentSettingId();
+    function sign(uint256 _settingId) external isInitialized {
         uint256 lastSettingIdSigned = lastSettingSignedBy[msg.sender];
-        require(lastSettingIdSigned < currentSettingId, ERROR_SIGNER_ALREADY_SIGNED);
+        require(lastSettingIdSigned < _settingId, ERROR_SIGNER_ALREADY_SIGNED);
+
+        uint256 currentSettingId = _getCurrentSettingId();
+        require(_settingId <= currentSettingId, ERROR_INVALID_SIGNING_SETTING);
```
sohkai, line 301:
```suggestion
        require(_settingId < nextSettingId, ERROR_INVALID_SIGNING_SETTING);
```

This would be marginally cheaper

sohkai (approved):
LGTM as well, just small optimizations

apps/agreement/test/agreement/agreement_initialize.js
facuspagnuolo: I couldn't set the default vault to an EOA, but I did use `transferToVault` with using a sample ERC20 :)

sohkai: Good point; the Kernel checks if the app address is a contract!

PR #1249, opened 2020-08-19
DisputableVoting: Update aragonOS to v5.0.0-beta.8

facuspagnuolo:


apps/voting-disputable/test/contracts/voting_disputable.js
```diff
@@ -24,7 +24,7 @@ contract('Voting disputable', ([_, owner, representative, voter10, voter20, vote
     collateralToken = await agreementDeployer.deployCollateralToken()
     votingDeployer.previousDeploy = agreementDeployer.previousDeploy
 
-    await agreement.sign(voter40)
+    await agreement.sign({ from: voter40 })
```
sohkai, line 27: I'm wondering if we should primarily use a mock Agreement for these tests, and leave tests with the full Agreement as an integration test?

In any case, this will probably break once we merge with master since `sign()`'s signature has changed.

sohkai (approved):
Other than the note on the tests, the changes look good to me!

apps/voting-disputable/test/contracts/voting_disputable.js
facuspagnuolo: Let me merge and explore this on top of the buidler changes :)

PR #1251, opened 2020-08-20
Agreement: Update staking to v0.2.9

facuspagnuolo:


apps/agreement/package.json
```diff
@@ -13,21 +13,22 @@
...
-    "test": "buidler test",
-    "test:gas": "REPORT_GAS=true buidler test --network localhost",
+    "test": "yarn compile:staking && buidler test",
+    "test:gas": "REPORT_GAS=true yarn test --network localhost",
     "coverage": "buidler coverage --network coverage",
     "abi:extract": "buidler-extract --output abi/ --keys abi"
   },
   "dependencies": {
-    "@aragon/os": "5.0.0-beta.8"
+    "@aragon/minime": "1.0.0",
```
sohkai, line 26: Can we move minime back to a devDependency?

sohkai (approved):
🙌 🙌 🙌 Looks great!

Just wasn't sure about placing it in the `staking/` dir directly (as opposed to `contract-mocks/` or `mocks/` or etc) but 🤷‍♂️.

Want to ensure that it's clear these are only used for the tests :).

apps/agreement/package.json
facuspagnuolo: Definitely!

facuspagnuolo:
@sohkai I decided to use `mocks` instead :)

PR #1252, opened 2020-08-20
DisputableVoting: Improve test suite to use a real Agreement instance

facuspagnuolo:


apps/voting-disputable/test/helpers/deployer.js
```diff
@@ -75,21 +98,35 @@ class VotingDeployer {
...
-    const contractName = `DisputableVoting${agreement ? 'Mock' : 'WithoutAgreementMock'}`
-    const Voting = this._getContract(contractName)
+  async deployAgreement(options = {}) {
+    const owner = options.owner || await this._getSender()
+    if (!this.dao) await this.deployDAO(owner)
+    if (!this.base) await this.deployBase(options)
+
+    const { currentTimestamp } = { ...DEFAULT_VOTING_INITIALIZATION_PARAMS, ...options }
+    agreementDeployer.previousDeploy = { ...agreementDeployer.previousDeploy, dao: this.dao, acl: this.acl, owner }
+    const agreement = await agreementDeployer.deployAndInitializeAgreementWrapper({ ...options, owner, currentTimestamp })
```
sohkai, line 122: I'm not sure about this—doesn't this move all of our tests to be integration tests, since we're using the real agreement?

We should definitely have a suite of these with a real agreement, but I was wondering if it would be easier for us to test using a mocked version for the unit tests?

PR #1256, opened 2020-08-25
[Internal audit] DisputableVoting

sohkai:
Comments left inline for d6bc944, by severity.

Some other small breaking changes I pushed already:

- 6698380: updates a parameter's name for `ChangeRepresentative`
- aa87e87: updates changes to the quiet ending configuration to be "configuration" rather than just "period".


apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -783,6 +814,8 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
         }
 
         // If the vote is still open, it cannot be executed
+        // Note: I think this does not correctly check for the quiet ending flip (and thereby extension)
+        //       If the vote was flipped but not extended through another vote, this will continually stay false, locking the vote from being executed
```
sohkai, line 818: The scenario:

1. Vote gets flipped in the last round
2. Time goes by and nobody votes during the extra extension (which is also "missing" / not registered in state)
3. The vote can't be executed, as `_hasEnded()` will continue to stay false, despite the "phantom" extension having been quiet.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -347,6 +367,8 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     */
     function canClose(uint256 _voteId) external view returns (bool) {
         Vote storage vote_ = _getVote(_voteId);
+        // Note: do we check for isExecuted() because the original call to close may fail (but not revert) for some reason?
+        // Note: wondering if we should expose a permissionless method to close any cancelled vote (and that way we can also set their status to be cancelled)?
         return (_isActive(vote_) || _isExecuted(vote_)) && _hasEnded(vote_);
```
sohkai, line 372: This is also affected by above—submitters wouldn't be able to manually close the vote from the Agreement's side.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -250,6 +269,7 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     */
     function vote(uint256 _voteId, bool _supports) external {
         Vote storage vote_ = _getVote(_voteId);
+        // Note: I think you can vote past the last quiet ending period (if no one else has already or if that last person did not again flip the vote)
```
sohkai, line 272: Similar to the scenario above, but instead of trying to close the vote, I believe anyone who hasn't voted yet can still vote despite the "phantom" extension having gone by quietly. Note that if this person has enough say to flip the vote again, someone else can do the same later.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -1022,8 +1061,10 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     * @param _voter Address of principal voter
     * @param _caster Address of vote caster, if voting via representative
     */
+    // Note: not sure about pushing these to be private 🤷‍♂️
     function _castVote(Vote storage _vote, uint256 _voteId, bool _supports, address _voter, address _caster) private {
         Setting storage setting = settings[_vote.settingId];
+        // Note: if we move the quiet ending check at the end here, we could optimize this so it doesn't get calculated outside of the quiet ending period
```
sohkai, line 1067: I think this and the next note are important, and related to the issue we have with `_hasEnded()`. It would be ideal if we updated the quiet ending state as a modifier / precondition to the rest of functions, so that we effectively work with the latest state down the line.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -77,6 +81,8 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
         // Must be <= supportRequiredPct to avoid votes being impossible to pass
         uint64 minAcceptQuorumPct;
 
+        // Note: we could re-work these concepts to be against the start of the vote rather than the end
+        //       E.g. overrule window is more like "allowed time for representatives to vote"
```
sohkai, line 85: We discussed this briefly offline, but I think we should give a go at this "refactor" to see if it can simplify the code and our mental model of the timings.

The best description I have for these values' behaviour is "the amount of voting-eligible time before/after which X happens", where X is one of:

- Time allowed for representatives to vote on others' behalf
- Time to take the initial quiet ending snapshot

----

Flipping it like this gets a bit harder to conceptualize for those who don't want those settings (e.g. "I don't want quiet ending" --> "your snapshot time should be the same as your vote time" rather than 0), but I think it's worth it.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -100,23 +106,31 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
...
         uint64 startDate;                                   // Datetime when the vote was created
         uint64 snapshotBlock;                               // Block number used to check voting power on attached token
         uint64 pausedAt;                                    // Datetime when the vote was paused
         uint64 pauseDuration;                               // Duration of the pause (only updated once resumed)
+        // Note: maybe rename to quietEndingExtensionDuration?
         uint64 quietEndingExtendedSeconds;                  // Total number of seconds a vote was extended due to quiet ending
+        // Note: I initially found this name a bit confusing, but it makes sense now
         VoterState quietEndingSnapshotSupport;              // Snapshot of the vote's support at the beginning of the first quiet ending period
+        // Note: let's prefer to store a hash here
         bytes executionScript;                              // EVM script attached to the vote
```
sohkai, line 125: I think this will save a lot of gas, which I think is much more important now that it can get quite expensive to create an action.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -100,23 +106,31 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
...
         mapping (address => VoteCast) castVotes;            // Mapping of voter address => more information about their cast vote
     }
 
+    // Note: may be good to double check if we'd like these to be updatable and push them to the settings
```
sohkai, line 129: Something to double check with @lkngtn; being able to change the time could be useful for the future.

Even if we don't let it be changed, it is something we could store in the settings array already (just a bit worried about gas cost, but I think this flexibility could be worth it for the 2/3k extra gas on votes).

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -876,12 +912,14 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
         return wasAcceptedBeforeLastFlip != _isAccepted(_vote, setting);
     }
 
+   // Note: maybe we should rename these "within" functions as they are not really "within" (they only check a single bound, rather than two)
```
sohkai, line 915: Minor gripe about the naming that may get solved with the above "refactor"; these functions act more like "have we passed this time checkpoint yet" rather than "am I within this time bound"?

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -131,16 +145,20 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
...
     event StartVote(uint256 indexed voteId, address indexed creator, bytes context);
     event PauseVote(uint256 indexed voteId, uint256 indexed challengeId);
     event ResumeVote(uint256 indexed voteId);
     event CancelVote(uint256 indexed voteId);
     event ExecuteVote(uint256 indexed voteId);
+    // Note: maybe we should stick with Vote* or *Vote? E.g. ExtendVote()?
     event VoteQuietEndingExtension(uint256 indexed voteId, bool passing);
 
+    // Note: should we remove stake?
     event CastVote(uint256 indexed voteId, address indexed voter, bool supports, uint256 stake);
```
sohkai, line 158: Thought the `stake` is pretty useless here; we already have everything we need (voter and snapshot block) to get it from a call.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -131,16 +145,20 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
...
     event ChangeRepresentative(address indexed voter, address indexed representative);
     event ProxyVoteFailure(uint256 indexed voteId, address indexed voter, address indexed representative);
+    // Note: maybe we can roll this into CastVote if we add the caster there?
     event ProxyVoteSuccess(uint256 indexed voteId, address indexed voter, address indexed representative, bool supports);
```
sohkai, line 162: Was wondering if we really need to emit this event—CastVote could be emitted with the `caster` to provide the same information as this (saving us 2-3k gas per proxy vote)

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -347,6 +367,8 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     */
     function canClose(uint256 _voteId) external view returns (bool) {
         Vote storage vote_ = _getVote(_voteId);
+        // Note: do we check for isExecuted() because the original call to close may fail (but not revert) for some reason?
+        // Note: wondering if we should expose a permissionless method to close any cancelled vote (and that way we can also set their status to be cancelled)?
```
sohkai, line 371: These are minor, but curious what you think :).

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -490,6 +516,7 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
         if (!_canRepresentativesVote(vote_)) {
             return false;
         }
+        // Note: should we also check that the length is <= MAX_VOTES_DELEGATION_SET_LENGTH? Perhaps better to keep this as a revert, similar to a missing vote
```
sohkai, line 519: Thought this might be useful to mirror `voteOnBehalfOf()`'s behaviour.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -1022,8 +1061,10 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     * @param _voter Address of principal voter
     * @param _caster Address of vote caster, if voting via representative
     */
+    // Note: not sure about pushing these to be private 🤷‍♂️
```
sohkai, line 1064: Wasn't really sure I understood what was meant to be private, but from the current ones, I'm not convinced any really need to be private 🤷

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -100,23 +106,31 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     struct Vote {
         uint256 yea;                                        // Voting power for
         uint256 nay;                                        // Voting power against
+        // Note: rename to totalPower?
         uint256 votingPower;                                // Total voting power (based on the snapshot block)
         uint256 settingId;                                  // Identification number of the setting applicable to the vote
         uint256 actionId;                                   // Identification number of the associated disputable action on the attached Agreement
+        // Note: we may want to re-order this state to be based on localized SSTOREs
+        //       start date and snapshot block never change, whereas pausedAt, pauseDuration and status may be changed together
```
sohkai, line 114: Not sure about this; just something I was thinking for a bit. It'd be relatively minor, but some of these variables are stored or loaded together at the same time, and we'd want those to be in the same slot.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -833,6 +866,7 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     function _isAccepted(Vote storage _vote, Setting storage _setting) internal view returns (bool) {
         uint256 yeas = _vote.yea;
         uint256 nays = _vote.nay;
+        // Note: minor storage optimization might be to pull the support and quorum variables out first
```
sohkai, line 869: The reason for this is that both of these vars are uint64s and should be accessible via one SLOAD (we should make sure to order them as such!). However, since they're used like this, I'm not sure what solidity will do.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -1095,7 +1140,9 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     /**
     * @dev Copy settings from one storage pointer to another
     */
+   // Note: should we inline?
     function _copySettings(Setting storage _from, Setting storage _to) private {
+        // Note: we may want to check if solidity is smart enough to not require us to first load into local vars to optimize sstores
```
sohkai, line 1145: Similar note to above; these should be loadable (and re-storable) in 1-2 SLOADs, but since we've chunked them like this, I'm not sure the compiler is smart enough :)

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -18,6 +19,7 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     // bytes32 public constant CREATE_VOTES_ROLE = keccak256("CREATE_VOTES_ROLE");
     bytes32 public constant CREATE_VOTES_ROLE = 0xe7dcd7275292e064d090fbc5f3bd7995be23b502c1fed5cd94cfddbbdcd32bbc;
 
+    // Note: maybe we should just use CHANGE_* as a prefix?
```
sohkai, line 22: Minor nitpick, but we call the functions "change*" so might as well name the roles similarly :)

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -61,6 +63,8 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     enum VoterState { Absent, Yea, Nay }
 
     enum VoteStatus {
+        // Note: may be able to find a better name than "active". This state just includes all votes
+        // that have not been challenged or executed (thus including votes that have ended).
```
sohkai, line 67: I'm having a hard time finding a better word to describe this state, so perhaps updating the "ongoing vote" comment would be easier.

"Active" is a bit loose but actually describes this state pretty well. Other choices I didn't like very much: "Normal", "Not disrupted", "Base", "Basic" (keeping in mind that ruling as accepted brings the vote back to "Active").

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -530,6 +558,7 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     * @param _voteId Vote identifier
     * @return True if the vote is within its overrule window
     */
+   // Note: maybe better renamed as `canRepresentativesVote()`?
     function withinOverruleWindow(uint256 _voteId) external view returns (bool) {
```
sohkai, line 562: Thought that perhaps we should stick to `can*`-names rather than `isVoteOpen()` and `withinOverruleWindow()`, to align with `canExecute()`, etc. It might also allow us to deduplicate some logic (e.g. `withinOverruleWindow()` is awkwardly not the same as `_canRepresentativesVote()`).

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: > do we check for isExecuted() because the original call to close may fail (but not revert) for some reason?

I'm having a hard time trying to remember why I added this, but I think this was introduced before we added the close action step in `executeVote`, now it seems redundant

> wondering if we should expose a permissionless method to close any cancelled vote (and that way we can also set their status to be cancelled)?

About adding an extra method to allow closing canceled votes, I wouldn't add it to be honest. Users can go through the agreement app instead. I wouldn't add unnecessary methods to the interface cause we are already close to the limit.

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: I see the benefit ofc, but I wouldn't add it to be honest. Users can go through the agreement app instead. I wouldn't add unnecessary methods to the interface simply because we are already close to the limit.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -852,6 +886,8 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     * @return True if the vote has ended
     */
     function _hasEnded(Vote storage _vote) internal view returns (bool) {
+        // Note: in these cases, we should prefer getTimestamp()
+        // Note: this lies if the vote is stuck in a flip; ideally we would also take into account the extra period for the flip
```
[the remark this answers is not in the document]
facuspagnuolo: Addressed in f7a96d8b837a4e03b934a5115638b2a411db2aad

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in f7a96d8b837a4e03b934a5115638b2a411db2aad

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in f7a96d8b837a4e03b934a5115638b2a411db2aad

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in f7a96d8b837a4e03b934a5115638b2a411db2aad

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Definitely, addressed in c93188d423b91af905e0728c956021e16dcf0255

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: My main intention was to make sure we don't expose unsafe or incomplete functions to inheriting contracts, but looking at the exposed interface, it doesn't seem harmful now. Addressed in 6e2ec983f8496dadbe110a88de7fdfe728465c6f

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in d3463b54a6d1039ddf15aae2e0ebc4eb7646bbfc

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in fa29e27058b852b6ec20115a4089316212db66ca

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in eff6dec49286034c16de77d3bd4f40c67d8d56ee

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in f25c3ca02101a272f9bd6d698951f78da20c6f9c

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in 8a6dcf10ed85c12aa184ba3b06eb63d4b8b851e8

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in d131b514f74993aa434e2ffc701dc74a030a9b58

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: I'm not exactly sure what would be the alternative here

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -530,6 +550,7 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     * @param _voteId Vote identifier
     * @return True if the vote is within its overrule window
     */
+   // Note: maybe better renamed as `canRepresentativesVote()`?
```
facuspagnuolo, line 553: It is actually the other way around, during the overrule window representatives cannot vote

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -4,6 +4,7 @@
 
 pragma solidity 0.4.24;
 
+// Note: it's not necessary, but we could make the import to IAgreement explicit here
```
facuspagnuolo, line 7: I wouldn't add an unnecessary import TBH

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -679,45 +716,96 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
...
+        }
+
+        _vote.yea = yeas;
+        _vote.nay = nays;
+        castVote.state = _voterStateFor(_supports);
+        castVote.caster = _caster;
+        emit CastVote(_voteId, _voter, _supports, _caster == address(0) ? _voter : _caster);
+
+        // Note: could we move this to be at the start of `castVote` so it acts like a modifier /
+        // pre-condition, making sure we're calculating the quiet ending period correctly?
```
facuspagnuolo, line 782: Having it here allows us to work with the previous and current support values to now if the outcome was flipped actually

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: It is actually the other way around, during the overrule window representatives cannot vote

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Having it here allows us to work with the previous and current support values to now if the outcome was flipped actually. I think having this as a precondition is less intuitive, don't you think? I mean, by definition is something you would always like to check after a vote, not before

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in 9609ef02a9a73dab443906adfe0f2b383a2e2acf

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: I agree but I don't like much the other options either haha.. let's go with "Normal" for now tho. Addressed in 9a29ff9238860352b9bbd6e1c71ac08871eb194d

apps/voting-disputable/contracts/DisputableVoting.sol
sohkai: I would double check with remix to see what the compiler does here, but the ugly alternative is to first load them all into local vars and then push them back into storage. So something ugly like:

```
uint256 supportRequiredPct_ = from.supportRequiredPct;
uint256 minAcceptQuorumPct_ = from.minAcceptQuorumPct;
...

to.supportRequiredPct = uint64(supportRequiredPct_);
to.minAcceptQuorumPct = uint64(minAcceptQuorumPct_);
...
```

apps/voting-disputable/contracts/DisputableVoting.sol
sohkai: Yes exactly—so with the above change on changing the configuration to be "start date"-based, `canRepresentativesVote()` seems more meaningful than `withinOverruleWindow()`.

apps/voting-disputable/contracts/DisputableVoting.sol
sohkai: > I mean, by definition is something you would always like to check after a vote, not before

But from what I understand, not really, right? We're actually using this as a pre-condition because we only know about a quiet ending extension when someone votes (or tries to close) in the new period.

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in 8765ccf206f6d4e8ff78e51d5e84d9dd7a77810d

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Siezed 8765ccf206f6d4e8ff78e51d5e84d9dd7a77810d and renamed it :)

facuspagnuolo:
@sohkai thx for the thorough review once again! I addressed almost all your comments and answered a few of then, let me know!

apps/voting-disputable/test/contracts/voting_delegation.js
```diff
@@ -27,6 +27,8 @@ contract('Voting delegation', ([_, owner, voter, anotherVoter, thirdVoter, repre
     voting = await deployer.deployAndInitialize({ owner, minimumAcceptanceQuorum: MIN_QUORUM, requiredSupport: MIN_SUPPORT, voteDuration: VOTE_DURATION, overruleWindow: OVERRULE_WINDOW, quietEndingPeriod: QUIET_ENDING_PERIOD, quietEndingExtension: QUIET_ENDING_EXTENSION })
   })
 
+  const repeat = (x, y) => [...Array(x)].map(() => y)
```
sohkai, line 30: Note that you can replace this with just `new Array(x).fill(y)` now :)

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -509,10 +509,9 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     function canVoteOnBehalfOf(uint256 _voteId, address[] _voters, address _representative) external view returns (bool) {
         Vote storage vote_ = _getVote(_voteId);
 
-        if (!_canRepresentativesVote(vote_)) {
+        if (!_canRepresentativesVote(vote_) || _voters.length > MAX_VOTES_DELEGATION_SET_LENGTH) {
```
sohkai, line 502: Interesting, do you prefer to not revert (it does save some space 😄)?

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -44,7 +44,7 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     string private constant ERROR_CHANGE_QUORUM_TOO_BIG = "VOTING_CHANGE_QUORUM_TOO_BIG";
     string private constant ERROR_CHANGE_SUPPORT_TOO_SMALL = "VOTING_CHANGE_SUPPORT_TOO_SMALL";
     string private constant ERROR_CHANGE_SUPPORT_TOO_BIG = "VOTING_CHANGE_SUPPORT_TOO_BIG";
-    string private constant ERROR_INVALID_OVERRULE_WINDOW = "VOTING_INVALID_OVERRULE_WINDOW";
+    string private constant ERROR_INVALID_DELEGATED_VOTING_PERIOD = "VOTING_INVALID_DELEGATED_VOT_PER";
```
sohkai, line 47:
```suggestion
    string private constant ERROR_INVALID_DELEGATED_VOTING_PERIOD = "VOTING_INVALID_DLGT_VOTE_PERIOD";
```

Nitpicky, but having "period" in the error is a bit more descriptive I think :)

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -78,11 +78,9 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
         // Must be <= supportRequiredPct to avoid votes being impossible to pass
         uint64 minAcceptQuorumPct;
 
-        // Note: we could re-work these concepts to be against the start of the vote rather than the end
```
sohkai: In terms of this, should we also "flip" the current concept we have of the quietEndingPeriod to be "since the start of a vote"?

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -966,14 +961,22 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
...
     */
-    function _hasStartedOverruleWindow(Vote storage _vote, Setting storage _setting) internal view returns (bool) {
-        return getTimestamp() >= _durationStartDate(_vote, _setting.overruleWindow);
+    function _hasFinishedDelegatedVotingPeriod(Vote storage _vote, Setting storage _setting) internal view returns (bool) {
+        uint64 pausedAt = _vote.pausedAt;
+        uint64 baseDelegatedVotingPeriodEndDate = _vote.startDate.add(_setting.delegatedVotingPeriod);
+
+        // If the vote was paused before the deleted voting period ends, we need to extend it
+        uint64 actualDeletedVotingEndDate = pausedAt != 0 && pausedAt < baseDelegatedVotingPeriodEndDate
+            ? baseDelegatedVotingPeriodEndDate.add(_vote.pauseDuration)
```
sohkai, line 974: You're going to get sick of these notes, but a minor optimization here is to load `pauseDuration` along with `pausedAt`.

Note that actually these are not in the same slot at the moment (but they should be—more on this in a comment later), but once they are, accessing `_vote.pausedAt` will give us `pauseDuration` almost for free.

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: You changed my mind, let's revert :)

apps/voting-disputable/contracts/DisputableVoting.sol
sohkai: We discussed offline; actually this the quietEndingPeriod makes more sense being calculated from the original end date.

facuspagnuolo: As discussed offline, since the logic won't be simplified, we agreed on keeping this setting tied to the vote end date. Additionally, it feels clearer from a user perspective

apps/voting-disputable/contracts/DisputableVoting.sol
sohkai: See 75ede7c.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -791,8 +792,8 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
             // this snapshot is never stored.
             _vote.quietEndingSnapshotSupport = _voterStateFor(_wasAccepted);
         } else {
-            // We only store the quiet ending extension if it wasn't confirmed yet. Note that we are trusting `_canVote()` for that.
-            // If we reached this point, it means the vote's flip was confirmed but the last computed end date is in the past.
+            // We are calculating quiet ending extensions via "rolling snapshots", and so we only update the vote's cached duration once
+            // the last period is over and we've confirmed the flip.
```
sohkai, line 800: @facuspagnuolo I was still a bit confused by the old comment, so I took the liberty of rewriting it.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -716,11 +716,11 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
...
 
         Vote storage vote_ = votes[voteId];
-        vote_.status = VoteStatus.Normal;
-        vote_.startDate = getTimestamp64();
-        vote_.snapshotBlock = snapshotBlock;
         vote_.totalPower = totalPower;
         vote_.settingId = _getCurrentSettingId();
+        vote_.startDate = getTimestamp64();
+        vote_.snapshotBlock = snapshotBlock;
+        vote_.status = VoteStatus.Normal;
```
sohkai, line 726: Note that this change to move `status` to be after `snapshot` block (to align with the struct), saves us 1.5k gas on new votes :)

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -990,6 +990,7 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
     * @param _setting Setting instance applicable to the vote
     * @return True if the vote's execution delay has not finished
     */
+   // Note: seems a bit weird to say "not", even if it is a minor optimization
```
sohkai, line 993: @facuspagnuolo probably a quick one—not sure how you feel about this.

apps/voting-disputable/contracts/DisputableVoting.sol
```diff
@@ -103,15 +103,19 @@ contract DisputableVoting is IForwarderWithContext, DisputableAragonApp {
         uint256 yea;                                        // Voting power for
```
sohkai, line 103: Note that this change (75ede7c916aab8d5c6bd1c25fb228ec1574fbb74) (and the associated commit fixing the gas test: 03e1f71) appear to actually be increasing the gas.

This is as micro-optimizing as it gets, but I wanted to prefer saving gas on votes.

-----

The full analysis:

- Switching the Vote's struct order like this increases the cost of challenging by ~19k gas
- Based on the change below with pre-fetching `pauseDuration`:
  - When voting before a challenge, this adds ~50-100 gas per vote (and the same amount if representative is voting)
  - On challenged votes, this saves ~750 gas per vote (and another ~750 saved if voting representative is voting)

On a rough calculation, this means you would need ~25 votes to break even with the challenger, but the challenger is also financially motivated and should be able to afford more gas.

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Confirmed offline we do want to introduce this change, addressed in 97b53e6f9231b36c828ce1c4e9c8de0d1fce2596

apps/voting-disputable/test/contracts/voting_delegation.js
facuspagnuolo: Addressed in beacbd92213037ea26201ea834229ed5f960432b

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in 2e03ab03737f177fb5350b36746ea92252dbeb1b

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in beacbd92213037ea26201ea834229ed5f960432b

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Yup, renamed in 09317601554141b9d29c187674380e8dd22389a9

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in 798589a04a663ab217c717e52a6f73c81a3dc196

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: I tried this optimization, but it makes the contract hit the bytecode size limit... not sure if that's a good tradeoff at the end if we need to sacrifice the cost of the rest of the functions

facuspagnuolo: I checked this removing some functions and it actually increases the costs haha

apps/voting-disputable/arapp.json
```diff
@@ -38,46 +38,47 @@
...
-      "name": "Modify quiet ending configuration",
-      "id": "CHANGE_QUIET_ENDING_ROLE",
+      "name": "Modify execution delay",
+      "id": "CHANGE_EXECUTION_DELAY_ROLE",
       "params": [
-        "New quiet ending period",
-        "Current quiet ending period",
-        "New quiet ending extension",
-        "Current quiet ending extension"
+        "New execution delay"
```
sohkai, line 81: Ah, nice catches!

apps/voting-disputable/test/contracts/voting_settings.js
```diff
@@ -27,6 +27,64 @@ contract('Voting settings', ([_, owner, anyone, holder51, holder20, holder29]) =
...
+          assertAmountOfEvents(receipt, 'NewSetting')
+          assertAmountOfEvents(receipt, 'ChangeVoteTime')
+          assertEvent(receipt, 'ChangeVoteTime', { expectedArgs: { voteTime: newVoteTime } })
+        })
+
+        it('does not affect vote times', async () => {
+          const { voteId, script } = await createVote({ voting, from: holder51 })
+          await voting.changeVoteTime(newVoteTime, { from })
+
+          await voting.mockIncreaseTime(newVoteTime * 2)
```
sohkai, line 65: Not sure I understand why this is done

sohkai (commented):
💃🕺 Looking good to me 👌

apps/voting-disputable/contracts/DisputableVoting.sol
facuspagnuolo: Addressed in 798589a04a663ab217c717e52a6f73c81a3dc196

apps/voting-disputable/test/contracts/voting_settings.js
facuspagnuolo: Right above I'm changing the vote time to X, here I'm increasing the time by 2X, then I'm trying to vote to assert that the vote time change didn't affect already existing votes

PR #1264, opened 2020-08-28
Agreement: Close evidence period automatically

facuspagnuolo:
@sohkai I noticed Aragon Court actually allows you to close the evidence period automatically, would you support this or do we prefer not to avoid risks on bricking disputes?

### DIFF
--- a/apps/agreement/test/agreement/agreement_dispute.js
+++ b/apps/agreement/test/agreement/agreement_dispute.js
@@ -126,11 +126,11 @@ contract('Agreement', ([_, someone, submitter, challenger]) => {
                       assertEvent(receipt, 'EvidenceSubmitted', { index: 1, expectedArgs: { arbitrator: disputable.arbitrator, disputeId, submitter: challenger, evidence: challengeContext, finished: true }, decodeForAbi: disputable.abi })
                     })
 
-                    it('allows closing the evidence submission period manually', async () => {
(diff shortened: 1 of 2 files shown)


apps/agreement/test/agreement/agreement_dispute.js
```diff
@@ -126,11 +126,11 @@ contract('Agreement', ([_, someone, submitter, challenger]) => {
                       assertEvent(receipt, 'EvidenceSubmitted', { index: 1, expectedArgs: { arbitrator: disputable.arbitrator, disputeId, submitter: challenger, evidence: challengeContext, finished: true }, decodeForAbi: disputable.abi })
                     })
 
-                    it('allows closing the evidence submission period manually', async () => {
```
sohkai, line 129: We should probably still keep this test to manually close it though?

sohkai (approved):
Nice! I like it—I think we should just be careful about what we're expecting from `IArbitrators` though. Can we assume / enforce (via convention) that all `IArbitrables`:

- Allow closing the evidence period like this?
- Revert if trying to close an already closed evidence period?

apps/agreement/test/agreement/agreement_dispute.js
facuspagnuolo: We do have some for this case in the evidence test file

facuspagnuolo: We do have some for this case in the evidence test file

PR #1266, opened 2020-09-05
Apps: update buidler config and remove keys

sohkai:
Strips these configs of keys, and uses `ETH_KEYS` instead. It didn't make much sense to default to the aragen default on rinkeby/mainnet in most cases anyway, since you would never want to use those in a real environment.

`ETH_KEYS` supports one or more comma-delinated keys.


0xGabi (approved):
LGTM!

I agree having aragen keys was not super useful, maybe on Rinekby made a bit more sense.

PR #1269, opened 2020-09-10
Agreement: Update staking dependency

facuspagnuolo:


sohkai (approved):
LGTM, do you mind also checking the gas tests to see if we can lower them? I think we should be able to!

PR #1271, opened 2020-09-11
Finance: fix transfers layout (text overflow)

bpierre:
image

@sohkai could you please check in Safari just to be sure? (I can only do Gecko / Blink)

Fixes https://github.com/aragon/aragon/issues/1212.


sohkai (approved):
🏎 LGTM, works on Safari!

PR #1273, opened 2020-09-15
DisputableVoting: Remove agreement from executors blacklist

facuspagnuolo:
The main reason is that Disputable Voting apps won't be able to control the agreement's settings otherwise


sohkai (approved):
LGTM

PR #1275, opened 2020-09-21
DisputableVoting: Allow principals submitting votes as representatives

facuspagnuolo:
This change will allow representatives casting their votes on behalf of their principals on their votes in one single tx


sohkai (approved):
Oops, forgot to sen the review!

LGTM, good catch! In terms of the fix, this is not so much about sending votes for representatives, but moreso about a principle being able to use `voteOnBehalfOf()` with their own vote?

PR #1283, opened 2021-02-26
Voting: use BN.js everywhere #1113

eliobricenov:
This is a fork from the PR https://github.com/aragon/aragon-apps/pull/1113, created by @rperez89 with just some minor changes that were requested at the end.


eliobricenov:
C.C: @nivida

apps/voting/app/src/math-utils.js
```diff
@@ -38,10 +39,88 @@ export function formatNumber(num, decimals = 2, { truncate = true } = {}) {
...
+    ) + suffix
+  )
+}
+
+// Converts a percentage expressed as a value + base into a number between 0 and 1.
+export function bnPercentageToNumber(value, base, precision = 10 ** 9) {
+  value = new BN(value)
+  base = new BN(base)
+  return (
+    parseInt(divideRoundBigInt(value.mul(new BN(precision)), base), 10) /
```
nivida, line 74:
```suggestion
    parseInt(divideRoundBigInt(new BN(value).mul(new BN(precision)), new BN(base)), 10) /
```

re-assignment is not required I think :-)

apps/voting/app/src/math-utils.js
```diff
@@ -38,10 +39,88 @@ export function formatNumber(num, decimals = 2, { truncate = true } = {}) {
     : formattedNumber
 }
 
-export function percentageList(values, digits = 0) {
+export function percentageList(values) {
   return scaleBNValuesSet(values).map(value => value.toNumber())
 }
 
+// Format a percentage from BN values.
```
nivida, line 46: Can you add to those functions standard JSDoc annotations? It will increase the readability.

This is how I normally do them: 
https://github.com/aragon/govern/blob/feature/govern-tx/packages/govern-tx/src/wallet/Wallet.ts#L36-L48

apps/voting/app/src/math-utils.js
```diff
@@ -38,10 +39,88 @@ export function formatNumber(num, decimals = 2, { truncate = true } = {}) {
...
+export function bnPercentageToNumber(value, base, precision = 10 ** 9) {
+  value = new BN(value)
+  base = new BN(base)
+  return (
+    parseInt(divideRoundBigInt(value.mul(new BN(precision)), base), 10) /
+    precision
+  )
+}
+
+export class Percentage {
```
nivida, line 79: Because you anyways use here ValueObjects to represent those values accordingly would I also only use those objects instead of exporting the formating functions and the classes. This means for example ``formatBnPercentage`` and ``bnPercentageToNumber`` can be accessed as static methods and non-static methods of those objects. 

e.g.:
```javascript
new Percentage(value, base).toNumber();
// OR
Percentage.toNumber(value, base, precision);
```

nivida (changes requested):
I have only added a few comments I would be happy if you could have a look at.

apps/voting/app/src/math-utils.js
eliobricenov: Can you please check again?

apps/voting/app/src/math-utils.js
nivida: I think the FuncDoc only got added in one place. :-)

apps/voting/app/src/math-utils.js
nivida: Looks already better I think we can even move the ``formatBnPercentage`` and ``bnPercentageToNumber`` logic directly within the class methods. WDYT?

apps/voting/app/src/math-utils.js
eliobricenov: You're right, my bad :)

apps/voting/app/src/math-utils.js
```diff
@@ -32,13 +33,103 @@ export function formatNumber(num, decimals = 2, { truncate = true } = {}) {
...
+  toNumber(precision = 10 ** 9) {
+    return (
+      parseInt(
+        divideRoundBigInt(this._value.mul(new BN(precision)), this._base),
+        10
+      ) / precision
+    )
+  }
+}
+
```
c0rv0s, line 103: should there be func docs here for TokenAmount?

eliobricenov: Yes it should, I will be adding one

eliobricenov: Done, can you check?

PR #1315, opened 2021-09-16
DAO-276:  fix agent issue

yuetloo:
Removed the role to fix the error `role not used in contract`


yuetloo:
@nivida , I'm closing this ticket because I'm not comfortable removing role definition and have found the workaround to use --skip-validation flag works.  I created DAO-336 to request for the deployment of Aragon Agent using the master branch.

PR #1435, opened 2022-06-02
fix assignVested notice

loga4:
Pay attention, the @notice parameter of the assignVested method in the TokenManagercontract breaks theradspec` parser: there is a infinity loop.

Context: https://github.com/aragon/radspec/pull/127


Voronchuk (approved):
Looks good, thanks!

